PR #51, opened 2026-04-07
feat (doc): pointing agents to model running on the cluster with vllm and kserve

jehlum11:


guide-local-agent-to-vllm-on-cluster.md
```diff
@@ -0,0 +1,118 @@
+# Running an Agent Locally with a Model Served on vLLM on OpenShift AI
```
mpk-droid, line 1: Thanks for documenting this — the content itself is useful. However, I think this doc targets the platform engineer persona (creating ServingRuntime/InferenceService CRs, tuning vLLM memory, exposing Routes), whereas this repo has so far focused on the AI engineer persona.

From the AI engineer's perspective, they just need to point their agent at a LlamaStack URL to access the model — the infrastructure behind it is abstracted away.

Before adding platform-focused content to this repo, I think we'd need to establish a clear pattern for how we organize and scope docs across personas. Otherwise we risk mixing concerns and making the repo harder to navigate for our primary audience.

mpk-droid (commented):
added a comment. lmk what you think.

guide-local-agent-to-vllm-on-cluster.md
jehlum11: Good point, I largely agree. I would say though that the separation isn't as strict. 
The platform engineer would deploy the operator (Kserve, llama-stack), then an end-user (an AIE etc) would need to still create instances against that operator (i.e. the CRs - Serving Runtime/inference serving, lls etc).  
Wdyt?

mpk-droid: Ah, I see. Thanks for the clarification. In my mind, the Platform engineer would also create the CRs for the operator. 

Drawing on my personal experience, i feel like a clear line between AI eng and Platform Eng would be that Plat eng handles all things cluster and exposes URIs for various resources and the AI Eng uses those resources to carry out some actions. I feel like this creates a cleaner boundaries in their responsibilities. Lets imagine that one of the resources is crash looping, with the line drawn as above, its clear that platform engineer would resolve it. wdyt?

andrewdonheiser: Hi @jehlum11 this PR has been parked for a while, do you still intend to merge this PR into the repo?

kami619:
@mpk-droid I vote to close this PR as it has been stale for a good amount of time, @jehlum11 can always re-open this when the time is right, it helps us keep our PR slate clean and many lint checks started to fail as there has been considerable change in that area.

mpk-droid:
> @mpk-droid I vote to close this PR as it has been stale for a good amount of time, @jehlum11 can always re-open this when the time is right, it helps us keep our PR slate clean and many lint checks started to fail as there has been considerable change in that area.

Agreed! 

@jehlum11 Closing it for now. Feel free to re-open it.

PR #68, opened 2026-04-20
docs: update agent readme.md 

MRGuziX:
Summary                                                                                                                                                                          
                                                                                                                                                                                   
  - Added the "Tracing (optional)" section to 5 agent READMEs, using the human-in-the-loop agent as the reference template                                                         
  - Covers both local MLflow server and OpenShift MLflow server setup                                                                                                              
  - Each agent's MLFLOW_EXPERIMENT_NAME is tailored to match its agent.yaml name                                                                                                   
                                                                                                                                                                                   
  Agents updated                                                                                                                                                                   
                                                                                                                                                                                   
  - agents/langgraph/react_agent                                                                                                                                                   
  - agents/langgraph/react_with_database_memory             
  - agents/langgraph/agentic_rag                                                                                                                                                   
  - agents/llamaindex/websearch_agent                                                                                                                                              
  - agents/vanilla_python/openai_responses_agent                                                                                                                                   
                                                                                                                                                                                   
  Reference                                                                                                                                                                        
                                                                                                                                                                                   
  - agents/langgraph/human_in_the_loop/README.md (template)                                                                                                                        
   
  <assisted-by: Claude Opus 4.6>

### DIFF
--- a/agents/langgraph/agentic_rag/README.md
+++ b/agents/langgraph/agentic_rag/README.md
@@ -40,6 +40,53 @@ make init
 
 Edit `.env` with your configuration, then:
 
+### Tracing (optional)
+
+Tracing is optional. If MLflow tracing is required, enable it by uncommenting and setting the following environment variables in the `.env` file.
+
+#### Tracing with a local MLflow server
+
+```ini
+MLFLOW_TRACKING_URI="http://localhost:5000"
+MLFLOW_EXPERIMENT_NAME="langgraph-agentic-rag"
+MLFLOW_HTTP_REQUEST_TIMEOUT=2
+MLFLOW_HTTP_REQUEST_MAX_RETRIES=0
+```
+
+Then start the MLflow server in a separate terminal:
+
+```bash
+# Start the MLflow server
+uv run --extra tracing mlflow server --port 5000
+```
+
+When `MLFLOW_TRACKING_URI` is set, `make run-app` and `make run-cli` will automatically install the tracing dependency.
+
+#### Tracing with an OpenShift MLflow server
+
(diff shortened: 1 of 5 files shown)


agents/langgraph/agentic_rag/README.md
```diff
@@ -40,6 +40,53 @@ make init
 
 Edit `.env` with your configuration, then:
 
+### Tracing (optional)
```
sanafayyaz315: The 5 READMEs use #### for their Local Development subsections (#### Initiating base, #### Creating environment, etc.), but the tracing section is added as ### Tracing (optional) (h3). It should be #### Tracing (optional) to match the siblings. HITL is the exception — it uses ### for everything, so it's correct there.
Same for all other Readmes as well.

agents/langgraph/agentic_rag/README.md
```diff
@@ -40,6 +40,53 @@ make init
 
 Edit `.env` with your configuration, then:
```
sanafayyaz315: The "Initiating base" section ends with Edit .env with your configuration, then: — the "then:" now dangles into the tracing section instead of pointing to the next step. Can you drop the "then:" and rephrase it to something self-contained like the HITL README? e.g. "make init creates a .env file from .env.example. Set your environment variables in the .env file."

Wojciech-Rebisz (commented):
Several template READMEs (e.g. Autogen, Google ADK, A2A) do not document MLflow at all. Maybe we could remove the section from the CONTRIBUTING.md file that says every agent must have mlflow? I mean:
```All agent templates in this repo must include MLflow tracing integration.```

Wojciech-Rebisz (approved):
LGTM

PR #77, opened 2026-04-23
feat: implement automated PR labeling (RHAIENG-4319)

tarun-etikala:
## Summary

- Add automated PR labeling triggered on `pull_request_target` (supports fork PRs)
- Path-based `area/*` labels via `actions/labeler@v6` — 13 labels covering all agent frameworks, helm charts, docs, CI, tests, and tracing
- Size-based `size/*` labels via custom Python script — thresholds calibrated against 65 historic PRs (xs/s/m/l/xl with XL at 1200+ lines)
- Advisory comment posted on XL PRs encouraging authors to split large changes
- Labels auto-update on push, old size labels cleaned up automatically

## Jira Ticket

[RHAIENG-4319](https://redhat.atlassian.net/browse/RHAIENG-4319)

## Testing

- [x] Open a test PR and verify `area/*` labels are applied based on changed files
- [x] Open a test PR and verify `size/*` label is applied based on lines changed
- [x] Push a commit to an open PR and verify labels update dynamically
- [x] Verify XL comment is posted only once (dedup check)
- [x] Verify fork PR labeling works via `pull_request_target`

Test PRs - https://github.com/tarun-etikala/agentic-starter-kits/pull/11,  https://github.com/tarun-etikala/agentic-starter-kits/pull/8

## Checklist

- [x] I have read CONTRIBUTING.md

## Review Guidance

- `.github/labeler.yml` — straightforward path-to-label YAML config, review for completeness of glob patterns
- `.github/scripts/pr_size.py` — core logic: size calculation, file exclusions, label management, XL comment with dedup. Key file to review
- `.github/workflows/pr-labeler.yml` — orchestrates both steps; note `pull_request_target` for fork support and `continue-on-error: true` on size step
- `CONTRIBUTING.md` — new "Automated PR labels" section documenting labels and thresholds for contributors

### DIFF
--- a/.github/scripts/pr_size.py
+++ b/.github/scripts/pr_size.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+"""
+PR Size Labeler — calculates PR size and applies size/* labels.
+
+Triggered by the pr-labeler workflow. Uses the GitHub API via PyGitHub
+to count lines changed (additions + deletions), excluding generated and
+binary files. Posts an advisory comment on XL PRs.
+
+Environment variables (set by the workflow):
+    GITHUB_TOKEN        — GitHub token with pull-requests: write
+    GITHUB_REPOSITORY   — owner/repo
+    PR_NUMBER           — pull request number
+"""
+
+import fnmatch
+import os
+import sys
+
+from github import Github
+from github.GithubException import GithubException
+
+# --- Configuration ---
+
+SIZE_THRESHOLDS = [
+    ("size/xs", 0, 10),
+    ("size/s", 11, 100),
+    ("size/m", 101, 500),
+    ("size/l", 501, 1199),
+    ("size/xl", 1200, float("inf")),
+]
+
+LABEL_COLORS = {
+    "size/xs": "0e8a16",
+    "size/s": "0e8a16",
+    "size/m": "0e8a16",
+    "size/l": "fbca04",
+    "size/xl": "d93f0b",
+}
+
+EXCLUDED_PATTERNS = [
+    "uv.lock",
+    "*.lock",
+    "*.generated.*",
+    "images/*",
+]
+
+XL_COMMENT_MARKER = "<!-- pr-size-labeler:xl-warning -->"
+
+XL_COMMENT_TEMPLATE = """{marker}
+**Large PR detected ({lines} lines changed)**
+
+This PR exceeds 1200 lines of code changes (excluding lock files and \
+generated content). Large PRs are harder to review thoroughly and are more \
+likely to introduce bugs.
+
+Consider splitting this PR into smaller, focused changes.
+"""
+
+
+def is_excluded(filename: str) -> bool:
+    """Check if a file matches any exclusion pattern."""
+    for pattern in EXCLUDED_PATTERNS:
+        if fnmatch.fnmatch(filename, pattern):
+            return True
+    return False
+
+
+def get_size_label(lines: int) -> str:
+    """Return the size label for the given line count."""
+    for label, lower, upper in SIZE_THRESHOLDS:
+        if lower <= lines <= upper:
+            return label
+    return "size/xl"
+
+
+def ensure_label_exists(repo, label_name: str) -> None:
+    """Create the label in the repo if it doesn't exist."""
+    try:
+        repo.get_label(label_name)
+    except GithubException:
+        color = LABEL_COLORS.get(label_name, "ededed")
+        repo.create_label(name=label_name, color=color)
+        print(f"Created label: {label_name}")
+
+
+def remove_old_size_labels(pr) -> None:
+    """Remove any existing size/* labels from the PR."""
+    for label in pr.get_labels():
+        if label.name.startswith("size/"):
+            pr.remove_from_labels(label.name)
+            print(f"Removed label: {label.name}")
+
+
+def calculate_size(pr) -> int:
+    """Calculate total lines changed, excluding ignored files."""
+    total = 0
+    for f in pr.get_files():
+        if not is_excluded(f.filename):
+            total += f.additions + f.deletions
+        else:
+            print(f"Excluded: {f.filename}")
+    return total
+
+
+def post_xl_comment(pr, lines: int) -> None:
+    """Post an XL warning comment if one doesn't already exist."""
+    for comment in pr.get_issue_comments():
+        if XL_COMMENT_MARKER in comment.body:
+            print("XL comment already exists, skipping")
+            return
+    body = XL_COMMENT_TEMPLATE.format(marker=XL_COMMENT_MARKER, lines=lines)
+    pr.create_issue_comment(body)
+    print("Posted XL warning comment")
+
+
+def main() -> None:
+    token = os.environ.get("GITHUB_TOKEN")
+    repo_name = os.environ.get("GITHUB_REPOSITORY")
+    pr_number_str = os.environ.get("PR_NUMBER")
+
+    if not all([token, repo_name, pr_number_str]):
+        print("Error: GITHUB_TOKEN, GITHUB_REPOSITORY, and PR_NUMBER must be set")
+        sys.exit(1)
+
+    pr_number = int(pr_number_str)
+
+    gh = Github(token, retry=3)
+    repo = gh.get_repo(repo_name)
+    pr = repo.get_pull(pr_number)
+
+    lines = calculate_size(pr)
+    label = get_size_label(lines)
+    print(f"PR #{pr_number}: {lines} lines changed -> {label}")
+
+    ensure_label_exists(repo, label)
+    remove_old_size_labels(pr)
+    pr.add_to_labels(label)
+    print(f"Applied label: {label}")
+
+    if label == "size/xl":
+        post_xl_comment(pr, lines)
+
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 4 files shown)


.github/scripts/pr_size.py
```diff
@@ -0,0 +1,153 @@
...
+def update_size_label(pr, new_label: str) -> bool:
+    """Update the size label on a PR. Returns True if a change was made."""
+    for label in pr.get_labels():
+        if label.name == new_label:
+            print(f"Label already correct: {new_label}")
+            return False
+        if label.name.startswith("size/"):
+            pr.remove_from_labels(label.name)
+            print(f"Removed label: {label.name}")
+    pr.add_to_labels(new_label)
```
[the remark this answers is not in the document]
tarun-etikala: it can only happen from manual intervention

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,53 @@
...
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  label-pr:
+    runs-on: ubuntu-latest
+    steps:
+      # Step 1: Path-based area/* labels
+      - name: Apply area labels
```
andrewdonheiser, line 23: **[Critical] Tag-pinned actions in a `pull_request_target` workflow — violates repo convention.**

The existing workflow (`agent-deployment-test.yaml`) SHA-pins every action:
```yaml
uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
```

This PR uses mutable tag references (`@v6`, `@v4`, `@v5`) on all three actions. These are first-party GitHub actions so the realistic risk of tag compromise is low, but the convention exists to prevent erosion — the next contributor may copy this pattern with a third-party action where the risk is real.

**Fix:** SHA-pin all three actions (`actions/labeler`, `actions/checkout`, `actions/setup-python`) matching the existing repo pattern.

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,53 @@
+# Automated PR labeling: path-based area/* labels + size-based size/* labels.
+# Uses pull_request_target to support fork PRs (write access to base repo).
+
+name: PR Labeler
+
+on:
+  pull_request_target:
+    types: [opened, synchronize]
```
andrewdonheiser, line 8: **[Important] Missing `concurrency` group — parallel runs can cause stale labels and duplicate comments.**

Without a `concurrency` key, two rapid pushes to the same PR trigger parallel workflow runs with no serialization. This enables two issues:
1. The `update_size_label` early-return bug (see comment on `pr_size.py`) can leave stale `size/*` labels when runs interleave.
2. Both runs can pass the XL comment dedup check before either writes, resulting in duplicate advisory comments.

**Fix:** Add a concurrency group keyed on the PR number:
```yaml
concurrency:
  group: pr-labeler-${{ github.event.pull_request.number }}
  cancel-in-progress: true
```

.github/scripts/pr_size.py
```diff
@@ -0,0 +1,153 @@
...
+            raise
+        color = LABEL_COLORS.get(label_name, "ededed")
+        repo.create_label(name=label_name, color=color)
+        print(f"Created label: {label_name}")
+
+
+def update_size_label(pr, new_label: str) -> bool:
+    """Update the size label on a PR. Returns True if a change was made."""
+    for label in pr.get_labels():
+        if label.name == new_label:
```
andrewdonheiser, line 95: **[Important] Early return skips cleanup of stale `size/*` labels that appear after the matching label.**

If the label iterator yields `["size/xs", "size/l"]` and `new_label` is `"size/xs"`, this returns `False` immediately on the first match — `"size/l"` is never seen or removed.

Under normal single-run operation, only one `size/*` label should exist. But without a concurrency group on the workflow (see workflow comment), concurrent runs can produce this state without manual intervention.

**Fix:** Iterate all labels first, collect stale `size/*` labels, and check if `new_label` is already present. Then remove stale labels and add if needed:
```python
def update_size_label(pr, new_label: str) -> bool:
    current_labels = list(pr.get_labels())
    found = False
    stale = []
    for label in current_labels:
        if label.name == new_label:
            found = True
        elif label.name.startswith("size/"):
            stale.append(label.name)
    if found and not stale:
        print(f"Label already correct: {new_label}")
        return False
    for name in stale:
        pr.remove_from_labels(name)
        print(f"Removed label: {name}")
    if not found:
        pr.add_to_labels(new_label)
        print(f"Applied label: {new_label}")
    return True
```

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - name: Install dependencies
+        run: pip install PyGithub==2.9.1
+
+      - name: Apply size labels
+        id: size_labels
+        continue-on-error: true
```
andrewdonheiser, line 47: **[Important] `continue-on-error: true` makes size-labeling failures invisible in the PR checks UI.**

If `pr_size.py` crashes, the job still shows green. The subsequent `::warning::` step logs to the workflow run summary (visible if you open the run), but it does not appear in the PR checks UI or PR timeline. For teams that don't actively monitor workflow run logs, size labeling can silently break.

**Fix — option A:** Remove `continue-on-error` and split size labeling into a separate job. Failures show as a visible (but non-blocking via branch protection) check.

**Fix — option B:** Keep `continue-on-error` but add a step that posts a PR comment when size labeling fails, making the breakage visible where people look:
```yaml
- name: Comment on size-label failure
  if: steps.size_labels.outcome == 'failure'
  env:
    GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  run: |
    gh pr comment ${{ github.event.pull_request.number }} \
      --body "⚠️ Size labeling failed on this run. Area labels were applied but the size label may be stale or missing."
```

andrewdonheiser (changes requested):
## Review Summary

### Critical (1)
1. All GitHub Actions are tag-pinned, not SHA-pinned — violates the repo convention established in `agent-deployment-test.yaml`, which SHA-pins every action. This is especially important on a `pull_request_target` workflow.

### Important (3)
2. Missing `concurrency` group — two rapid pushes can trigger parallel workflow runs with no serialization, enabling stale labels and duplicate XL comments.
3. `update_size_label` early-return bug — stale `size/*` labels can survive when the target label appears before them in the iterator.
4. `continue-on-error: true` — failures in size labeling don't surface in the PR checks UI.

### What's done well
- `pull_request_target` + base-branch checkout pattern is correctly implemented.
- `is_excluded` correctly handles nested image directories.
- `ensure_label_exists` properly re-raises non-404 exceptions.
- XL comment dedup via HTML marker is clean.
- CONTRIBUTING.md documentation is thorough and accurate.
- Size thresholds calibrated against 65 historic PRs — good data-driven approach.

.github/scripts/pr_size.py
```diff
@@ -0,0 +1,164 @@
...
+
+def post_xl_comment(pr, lines: int) -> None:
+    """Post an XL warning comment if one doesn't already exist."""
+    for comment in pr.get_issue_comments():
+        if XL_COMMENT_MARKER in comment.body:
+            print("XL comment already exists, skipping")
+            return
+    body = XL_COMMENT_TEMPLATE.format(marker=XL_COMMENT_MARKER, lines=lines)
+    pr.create_issue_comment(body)
+    print("Posted XL warning comment")
```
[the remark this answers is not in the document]
tarun-etikala: The workflow has a concurrency group with cancel-in-progress: true keyed on the PR number, so only one run executes at a time per PR. The race condition described here cannot occur.

.github/workflows/pr-labeler.yml
tarun-etikala: Fixed in commit 57ae987. All three actions are now SHA-pinned matching the existing repo convention:
- actions/labeler@634933edcd8ababfe52f92936142cc22ac488b1b # v6.0.1
- actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
- actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0

.github/workflows/pr-labeler.yml
tarun-etikala: Fixed in commit 57ae987. Added concurrency group keyed on PR number with cancel-in-progress: true. Only one labeling run can execute per PR at a time.

.github/scripts/pr_size.py
tarun-etikala: Fixed in commit 57ae987. Replaced the early-return loop with a collect-then-act pattern: collects all size/* labels in one pass, removes stale ones, and only adds the new label if not already present. This handles the edge case where multiple size labels exist regardless of iteration order.

.github/workflows/pr-labeler.yml
tarun-etikala: Keeping the current approach: continue-on-error with a ::warning:: step. Labeling is advisory-only by design and should never block PRs. The warning is visible in workflow run logs for anyone investigating. Splitting into a separate job or adding PR comments on failure would add complexity for a scenario that should be rare and non-critical.

andrewdonheiser (approved):

PR #82, opened 2026-04-28
feat(evalhub): EvalHub adapter — E2E validated, hardened, documented

andrewdonheiser:
## Summary

Completes the EvalHub adapter migration (RHAIENG-4605) and validates it end-to-end against both implemented agents on the `agentic-mcp` cluster.

- **Adapter hardening**: URL/param validation, MLflow connection verification, scorer exception handling, insecure-TLS gating, sanitized error messages
- **MLflow integration**: `verify_connection()` on startup, separate `mlflow_trace_experiment_name`, run-ID propagation to EvalHub results, richer run tags
- **Comprehensive docs**: full end-to-end walkthrough in the [adapter README](evals/evalhub_adapter/README.md), new [Adding an EvalHub Agent Integration](docs/adding-evalhub-agent-integration.md) guide
- **E2E automation**: [`run-e2e.sh`](evals/evalhub_adapter/tests/run-e2e.sh) script that auto-discovers agent/EvalHub routes, builds/pushes the adapter image, registers the provider, submits jobs for both agents, polls for results, and cleans up
- **Test coverage**: 50 unit tests + 11 integration tests covering the new validation, MLflow features, and error handling
- **Cleanup**: removed stale `injection_payloads.yaml` duplicates from agent dirs, removed unused `EXPOSE 8080` from Containerfile

## How to test

The easiest way to validate is with the automated E2E script against both implemented agents (LangGraph `react_agent` and vanilla Python `openai_responses_agent`):

```bash
cd evals/evalhub_adapter/tests
REGISTRY_USER=<your-quay-user> OC_NAMESPACE=<your-namespace> ./run-e2e.sh
```

The script handles image build, provider registration, job submission, polling, and cleanup. See the [adapter README end-to-end walkthrough](evals/evalhub_adapter/README.md#end-to-end-walkthrough-evalhub-orchestration) for the full manual flow and parameter reference.

**Cluster credentials**: login credentials for the `agentic-mcp` cluster are stored in Bitwarden.

## Test plan

- [x] Unit tests pass (`make test` or `pytest evals/evalhub_adapter/tests/test_adapter.py evals/evalhub_adapter/tests/test_config_and_evaluations.py`)
- [x] Integration tests pass (`pytest evals/evalhub_adapter/tests/test_integration.py`)
- [ ] `run-e2e.sh` completes successfully against both agents on `agentic-mcp`
- [ ] MLflow runs appear with correct metrics, tags, and run IDs
- [ ] `evalhub eval results` returns non-null `mlflow_run_id`


andrewdonheiser:
All 5 CodeRabbit findings addressed in dd91971:

1. **Containerfile deps pinned** — bounded version ranges for `eval-hub-sdk`, `httpx`, `mlflow`, `PyYAML`
2. **JSONPath `contains` replaced** — new `get_route_contains()` helper using `oc get route -o custom-columns` + `grep`/`awk` for all three route discovery fallbacks
3. **MLflow token validation endpoint** — switched to protected `/api/2.0/mlflow/experiments/list` with separate 401/403 vs other error handling
4. **Experiment lookup hardened** — added `CURL_TLS_FLAG`, `Authorization: Bearer` header, and `urllib.parse.quote()` URL-encoding
5. **Env var consistency** — fixed `MLFLOW_TOKEN` → `MLFLOW_TRACKING_TOKEN` in `mlflow_client.py` error guidance

Also added missing docstrings (`__post_init__`, `__init__` methods) toward 80% coverage.

**Note:** The CodeQL "Analyze (actions)" failure is a pre-existing infrastructure issue (CodeQL cannot process GitHub Actions files in this repo) — not related to this PR's changes.

evals/evalhub_adapter/config.py
```diff
@@ -0,0 +1,174 @@
...
+    parsed = urlparse(url)
+    if parsed.scheme not in _ALLOWED_URL_SCHEMES:
+        raise ValueError(
+            f"{label} has unsupported scheme '{parsed.scheme}'. "
+            f"Only {_ALLOWED_URL_SCHEMES} are allowed."
+        )
+    if not parsed.netloc:
+        raise ValueError(f"{label} has no host component: '{url}'")
+    hostname = parsed.hostname or ""
+    if hostname in _BLOCKED_HOSTS or hostname.startswith("169.254."):
```
sanafayyaz315, line 41: _validate_url blocks localhost, which means the adapter can't be run locally against a local agent or local MLflow for dev/testing. You already handle a similar case in this file — verify_ssl=False is blocked by default but allowed when EVALHUB_ALLOW_INSECURE_TLS=true is set (line 124-132). The same approach could work for localhost. Something like:


```
if hostname in _BLOCKED_HOSTS or hostname.startswith("169.254."):
    is_localhost = hostname in {"localhost", "127.0.0.1", "::1", "0.0.0.0"}
    if is_localhost and os.environ.get("EVALHUB_ALLOW_LOCALHOST", "").lower() == "true":
        logger.warning(
            "%s targets '%s' — allowed via EVALHUB_ALLOW_LOCALHOST. "
            "Do not use in production.", label, hostname,
        )
    else:
        raise ValueError(
            f"{label} targets a blocked host '{hostname}'. "
            "Cloud metadata endpoints and localhost are not allowed."
        )
```

evals/evalhub_adapter/Containerfile
```diff
@@ -0,0 +1,44 @@
...
+FROM registry.access.redhat.com/ubi9/python-312@sha256:e95978812895b9abb2bdc109b501078da2a47c8dbb9fa23758af40ed50ab6023
+WORKDIR /opt/app-root/src
+
+USER 0
+
+COPY --from=ghcr.io/astral-sh/uv@sha256:fc93e9ecd7218e9ec8fba117af89348eef8fd2463c50c13347478769aaedd0ce /uv /usr/local/bin/uv
+
+COPY evals/evalhub_adapter/ ./evalhub_adapter/
+COPY evals/harness/          ./harness/
+COPY agents/langgraph/react_agent/evalhub/ ./fixtures/langgraph_react/
```
sanafayyaz315, line 23: Lines 22-23 hardcode a COPY line per agent, and line 34 asserts both exist. Adding a new agent requires editing the Containerfile and the assertion — it won't be picked up automatically. For now with 2 agents this is fine, but as more agents are onboarded, consider auto-discovery of agents/*/evalhub/ directories so new agents with eval fixtures are included without Containerfile changes.

sanafayyaz315 (commented):
Looks good overall. Added a few comments

kami619:
Test contract mismatch for .[test] installs evals/evalhub_adapter/README.md says uv pip install .[test] is enough for adapter tests, but two unit tests patch top-level mlflow symbols and fail when mlflow is not installed.

Claude Reproduced this issue in a clean env:

UV_PROJECT_ENVIRONMENT=/tmp/uv-pr82-test uv sync --extra test
UV_PROJECT_ENVIRONMENT=/tmp/uv-pr82-test uv run --extra test pytest evals/evalhub_adapter/tests -m unit -q
Result: 2 failed, 60 passed (ModuleNotFoundError: No module named 'mlflow')

test_adapter.py Lines 353-360
```
class TestLogMlflowRun:
    """Tests for MLflow run logging and run_id propagation."""
    @patch("mlflow.log_metric")
    @patch("mlflow.log_param")
    @patch("mlflow.start_run")
    @patch("mlflow.set_experiment")
    @patch("mlflow.set_tracking_uri")
```

README.md Lines 479-482
```
Tests stub `evalhub` imports if the package isn't installed (bootstrap
lives in `conftest.py` so it runs before any test module, regardless of
which files are selected). `uv pip install .[test]` alone is sufficient.
```

mpk-droid:
Just learning about the adapter. You might have already looked into this. @andrewdonheiser do you think the adapter would be better added to this repo(https://github.com/eval-hub/eval-hub-contrib)?

@kami619 @sanafayyaz315 are you guys aware of this? Maybe we should ask the suggestion from one of the contributors of that repo?

My reasoning is that, this adapter seems like it could benefit any agent and a larger audience than just agentic-starter-kit users?

kami619:
> Just learning about the adapter. You might have already looked into this. @andrewdonheiser do you think the adapter would be better added to this repo(https://github.com/eval-hub/eval-hub-contrib)?
> 
> @kami619 @sanafayyaz315 are you guys aware of this? Maybe we should ask the suggestion from one of the contributors of that repo?
> 
> My reasoning is that, this adapter seems like it could benefit any agent and a larger audience than just agentic-starter-kit users?


From my point of view, I think it is fine to land it here, as we need this for the QG7: Behavioral Evals., once this matures enough, we can make a contribution to the evalhub space directly.

mpk-droid:
> > Just learning about the adapter. You might have already looked into this. @andrewdonheiser do you think the adapter would be better added to this repo(https://github.com/eval-hub/eval-hub-contrib)?
> > @kami619 @sanafayyaz315 are you guys aware of this? Maybe we should ask the suggestion from one of the contributors of that repo?
> > My reasoning is that, this adapter seems like it could benefit any agent and a larger audience than just agentic-starter-kit users?
> 
> From my point of view, I think it is fine to land it here, as we need this for the QG7: Behavioral Evals., once this matures enough, we can make a contribution to the evalhub space directly.



> > Just learning about the adapter. You might have already looked into this. @andrewdonheiser do you think the adapter would be better added to this repo(https://github.com/eval-hub/eval-hub-contrib)?
> > @kami619 @sanafayyaz315 are you guys aware of this? Maybe we should ask the suggestion from one of the contributors of that repo?
> > My reasoning is that, this adapter seems like it could benefit any agent and a larger audience than just agentic-starter-kit users?
> 
> From my point of view, I think it is fine to land it here, as we need this for the QG7: Behavioral Evals., once this matures enough, we can make a contribution to the evalhub space directly.

Thats a fair point. We will need them for QG7. In the future, we will need to find a way to have these in a centralized place and import for testing.

mpk-droid:
> **Large PR detected (3576 lines changed)**
> 
> This PR exceeds 1200 lines of code changes (excluding lock files, generated content, and images). Large PRs are harder to review thoroughly and are more likely to introduce bugs.
> 
> Consider splitting this PR into smaller, focused changes.



> **Large PR detected (3576 lines changed)**
> 
> This PR exceeds 1200 lines of code changes (excluding lock files, generated content, and images). Large PRs are harder to review thoroughly and are more likely to introduce bugs.
> 
> Consider splitting this PR into smaller, focused changes.

I feel like either we will have to adjust the PR size rule or follow the suggestion. What do you guys think?  In general, I am on the smaller PR team. Curious to hear everyone else's thoughts.

kami619:
> > **Large PR detected (3576 lines changed)**
> > This PR exceeds 1200 lines of code changes (excluding lock files, generated content, and images). Large PRs are harder to review thoroughly and are more likely to introduce bugs.
> > Consider splitting this PR into smaller, focused changes.
> 
> > **Large PR detected (3576 lines changed)**
> > This PR exceeds 1200 lines of code changes (excluding lock files, generated content, and images). Large PRs are harder to review thoroughly and are more likely to introduce bugs.
> > Consider splitting this PR into smaller, focused changes.
> 
> I feel like either we will have to adjust the PR size rule or follow the suggestion. What do you guys think? In general, I am on the smaller PR team. Curious to hear everyone else's thoughts.

100% on the small PR team, but I think with these initial bits being added, we would have to live through the pain a bit, otherwise we might succumb to the Agile papercuts. But we need to make a conscious effort once we have scaffolding in place, to ensure the atomic PRs become the norm.

sanafayyaz315:
> Just learning about the adapter. You might have already looked into this. @andrewdonheiser do you think the adapter would be better added to this repo(https://github.com/eval-hub/eval-hub-contrib)?
> 
> @kami619 @sanafayyaz315 are you guys aware of this? Maybe we should ask the suggestion from one of the contributors of that repo?
> 
> My reasoning is that, this adapter seems like it could benefit any agent and a larger audience than just agentic-starter-kit users?

@mpk-droid I had the same thought of contributing it to the eval-hub-contrib repo. This adapter is agent-agnostic — the runner, scorers, and MLflow integration work with any agent that exposes /chat/completions. The only agent-specific piece is the fixture YAML files (golden queries and expected tools).

That said, I think we should first validate it across the agents we have in agentic-starter-kits-v2 (e.g., LangGraph, OpenAI Responses) to prove it handles different agents reliably, and then contribute it to the contrib repo.

andrewdonheiser:
> > **Large PR detected (3576 lines changed)**
> > This PR exceeds 1200 lines of code changes (excluding lock files, generated content, and images). Large PRs are harder to review thoroughly and are more likely to introduce bugs.
> > Consider splitting this PR into smaller, focused changes.
> 
> > **Large PR detected (3576 lines changed)**
> > This PR exceeds 1200 lines of code changes (excluding lock files, generated content, and images). Large PRs are harder to review thoroughly and are more likely to introduce bugs.
> > Consider splitting this PR into smaller, focused changes.
> 
> I feel like either we will have to adjust the PR size rule or follow the suggestion. What do you guys think? In general, I am on the smaller PR team. Curious to hear everyone else's thoughts.

the tests and documentation take up a lot of size. the adapter code itself is stripped down. I don't think splitting it up makes sense.

andrewdonheiser:
@kami619 Good catch — fixed in 373b0af. Added an `mlflow` stub in `conftest.py` following the same pattern used for the `evalhub` stub, so `@patch("mlflow.log_metric")` etc. resolve even when only `.[test]` is installed (without `test-mlflow`). All 68 unit tests pass in a `.[test]`-only environment now.

evals/evalhub_adapter/tests/run-e2e.sh
```diff
@@ -0,0 +1,546 @@
...
+PROVIDER_ID=$(echo "${PROVIDER_RESPONSE}" | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('resource',{}).get('id','') or d.get('id',''))" 2>/dev/null || true)
+if [[ -z "${PROVIDER_ID}" ]]; then
+  echo "  WARNING: Could not extract provider ID from response."
+  echo "  Run 'evalhub providers list' and set PROVIDER_ID manually, then re-run."
+  evalhub providers list
+  exit 1
+fi
+echo "  Provider ID: ${PROVIDER_ID}"
+
+evalhub providers list
```
kami619:
```suggestion
evalhub providers list || echo "  (providers list failed — non-fatal, continuing)"
```

kami619:
E2E validated on RHOAI 3.4.0-ea.2 cluster — works end-to-end, but surfaced a server/SDK version mismatch that will affect users.
                                                                                                                                                                                                                                                                                          

-   The E2E script ran successfully against both agents (react + openai-responses) with eval scores returned. However, the operator-shipped EvalHub image (odh-eval-hub-rhel9@sha256:c27bfe01...) is at server version 0.2.0 (reports 0.0.1 due to missing BUILD_NUMBER in the Red Hat build). This image is incompatible with the SDK >=0.1.4 that the adapter Containerfile and pyproject.toml specify.                                                                                                                                                                      
 
-   What breaks: REST-registered providers (the BYOF path this adapter uses) fail at job execution time with provider not found. The server accepts the registration but the k8s runtime can't resolve it. This means the entire custom adapter flow doesn't work on the shipped image — only pre-installed ConfigMap providers (lm-eval-harness, Garak, etc.) function.
                                                                                                                                                                                                                                                                            
-   Workaround we used: Replaced the server image with quay.io/evalhub/evalhub:0.3.0 (scaled down the TrustyAI operator to prevent reconciliation). With 0.3.0, both jobs completed successfully.                                                                                           

   
Suggestion: Either document the minimum server version requirement (0.3.0) in the runbook/README, or coordinate with the EvalHub team to ensure RHOAI ships 0.3.0+ before this adapter lands. The COMPATIBILITY.md in the eval-hub repo also needs updating — the only listed pair (0.1.0 + 0.1.0a8) no longer exists as published artifacts.
                                                                                                                                                                                                                                                                                          
Additional notes from the E2E run:                                                                                                                                                                                                                                                      
  - mlflow_run_id returned null — MLflow auth from adapter pods needs investigation (token was passed but MLflow's /mlflow subpath may need explicit configuration in the adapter)
  - evalhub providers list crashes on SDK 0.1.6 when built-in providers return tags: null (Pydantic validation error) — fixed by the || true guard in the script                                                                                                                          
  - Sidecar image config field changed between server versions (service.eval_sidecar_image in 0.2.0 vs sidecar.sidecar_container.image in 0.3.0) — another reason to align on 0.3.0

evals/evalhub_adapter/tests/run-e2e.sh
```diff
@@ -0,0 +1,546 @@
...
+    }
+  ],
+  "runtime": {
+    "k8s": {
+      "Image": "${ADAPTER_IMAGE}",
+      "Entrypoint": ["python", "-m", "evalhub_adapter.adapter"],
+      "Env": [
+        {"name": "MLFLOW_TRACKING_TOKEN", "value": ${MLFLOW_TOKEN_JSON}},
+        {"name": "MLFLOW_TRACKING_INSECURE_TLS", "value": "${MLFLOW_INSECURE_TLS}"},
+        {"name": "MLFLOW_WORKSPACE", "value": "${OC_NAMESPACE}"}
```
kami619:
```suggestion
        {"name": "MLFLOW_WORKSPACE", "value": "${OC_NAMESPACE}"},
        {"name": "EVALHUB_ALLOW_LOCALHOST", "value": "true"}
```

kami619: @andrewdonheiser we need this to allow localhost traffic, however that would enable un-encrypted transmission, need to call out to not use this in Production.

cc: @mpk-droid

kami619 (approved):

PR #83, opened 2026-04-28
feat: add integration test for LangGraph React Agent deployment (RHAIENG-4643)

mpk-droid:
## Summary
- Adds shared integration test utilities at `tests/integration/` (`run_make`, `get_route`, `health_check`) and shared pytest fixtures (`cluster_auth`, `repo_root`) for reuse across agent deployment tests
- Adds `agents/langgraph/react_agent/tests/integration/test_deployment.py` — builds on cluster via `build-openshift`, deploys to OpenShift, validates `GET /health` returns 200 with `{"status": "healthy"}`, and tears down via fixture finalizer
- Adds `make test-integration` Makefile target with JUnit XML output for CI compatibility
- Extracts `.github/actions/setup-cluster` composite action to DRY shared CI setup (Python, uv, oc, helm, login)
- Refactors CI workflow to use matrix strategy — adding a new agent is a one-liner
- Renames workflow to "QG4: Agent Deployment Integration Tests"

## Test plan
- [x] `make test` still runs only unit/behavioral tests (integration tests excluded via `--ignore`)
- [x] `make test-integration` runs the deployment test end-to-end on the demo cluster
- [x] Test tears down deployment even on failure (fixture finalizer)
- [x] JUnit XML output is produced at `results.xml`
- [x] CI workflow passes — [run 25121558044](https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/25121558044)

Jira: [RHAIENG-4643](https://redhat.atlassian.net/browse/RHAIENG-4643)

[RHAIENG-4643]: https://redhat.atlassian.net/browse/RHAIENG-4643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mpk-droid:
@coderabbit

mpk-droid:
@coderabbit review

agents/langgraph/react_agent/tests/integration/test_deployment.py
```diff
@@ -0,0 +1,61 @@
...
+
+        logger.info("Pushing container image...")
+        run_make("push", cwd=agent_dir, timeout=300)
+
+        logger.info("Deploying to cluster...")
+        run_make("deploy", cwd=agent_dir, timeout=300)
+        deployed = True
+
+        route_url = get_route(AGENT_NAME)
+        logger.info("Agent deployed at %s", route_url)
```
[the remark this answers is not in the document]
mpk-droid: Fixed with AI in 3b4a44e

tests/integration/utils.py
```diff
@@ -0,0 +1,124 @@
...
+                wait = min(backoff * (2 ** attempt), 15.0)
+                logger.info(
+                    "Health check attempt %d/%d failed, retrying in %.1fs: %s",
+                    attempt + 1,
+                    retries,
+                    wait,
+                    exc,
+                )
+                time.sleep(wait)
+
```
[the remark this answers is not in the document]
mpk-droid: Fixed with AI in 3b4a44e

agents/langgraph/react_agent/tests/integration/test_deployment.py
```diff
@@ -0,0 +1,83 @@
...
+    MakeTargetError,
+    RouteNotFoundError,
+    get_route,
+    health_check,
+    run_make,
+)
+
+logger = logging.getLogger(__name__)
+
+AGENT_NAME = "langgraph-react-agent"
```
andrewdonheiser, line 17: How does this pattern scale to more agents? I think if you would have implemented 2 agents it would be clearer

agents/langgraph/react_agent/tests/integration/test_deployment.py
```diff
@@ -0,0 +1,83 @@
...
+    env_path.write_text(
+        f"API_KEY={os.environ.get('API_KEY', 'not-needed')}\n"
+        f"BASE_URL={os.environ['BASE_URL']}\n"
+        f"MODEL_ID={os.environ['MODEL_ID']}\n"
+        f"CONTAINER_IMAGE={container_image}\n"
+    )
+    return env_path
+
+
+@pytest.fixture(scope="module")
```
andrewdonheiser, line 44: should this be in conftest.py?

tests/integration/utils.py
```diff
@@ -0,0 +1,126 @@
...
+import subprocess
+import time
+from pathlib import Path
+
+import httpx
+
+logger = logging.getLogger(__name__)
+
+
+class MakeTargetError(Exception):
```
andrewdonheiser, line 28: by using an exception class pattern do we risk swallowing uncaught Exceptions in places they are used?

andrewdonheiser (approved):
A couple of questions but it looks good 👍

agents/langgraph/react_agent/tests/integration/test_deployment.py
mpk-droid: Good question — let me address both angles:

**If you're asking about this specific constant:** You're right that hardcoding it isn't ideal. The agent name is already defined in `agent.yaml` and the Makefile reads it from there, so the test could do the same to stay DRY. What do you think?

**If you're asking about the general pattern:** Adding integration tests to another agent would look like:
1. Copy `tests/integration/test_deployment.py` into the new agent's directory, adjust `agent_dir`
2. Add the agent's `conftest.py` re-export (one line)
3. Add `test-integration` to the agent's Makefile
4. Add one entry to the CI matrix in `agent-deployment-test.yaml`

The shared utilities (`run_make`, `get_route`, `health_check`) and fixtures (`cluster_auth`, `repo_root`) are reused — each agent only needs the thin test file.

agents/langgraph/react_agent/tests/integration/test_deployment.py
mpk-droid: right now deployed_agent is only used by this one test file, so keeping it here makes the test self-contained and easier to read. If we later add more test files against the deployed agent (e.g., testing /chat/completions or running load tests), then moving it to conftest.py would make sense.

tests/integration/utils.py
mpk-droid: no, the custom exceptions actually make the error handling narrower, not broader. The catch blocks are specific: except (MakeTargetError, RouteNotFoundError) — any unexpected exception (e.g., OSError, RuntimeError) propagates normally and fails the test with a full traceback. The risk would be if we caught bare Exception, but we don't do that anywhere here.

kami619 (approved):
lgtm

agents/langgraph/react_agent/tests/integration/test_deployment.py
mpk-droid: Eitherways, will address this in the upcoming PR.

PR #90, opened 2026-05-04
feat: add integration deployment test for LangGraph HITL agent

mpk-droid:
## Summary
- Adds integration deployment test for the **LangGraph Human-in-the-Loop agent** (`agents/langgraph/human_in_the_loop/`) — builds on-cluster via `build-openshift`, deploys to OpenShift, validates `GET /health` returns `{"status": "healthy", "agent_initialized": true}`, and tears down via fixture finalizer
- Introduces `load_agent_name()` shared utility in `tests/integration/utils.py` that reads the agent name from `agent.yaml` instead of hardcoding it (addresses [PR #83 feedback](https://github.com/red-hat-data-services/agentic-starter-kits/pull/83#discussion_r3157286625) from @andrewdonheiser)
- Updates `react_agent` integration test to use `load_agent_name()` as well
- Adds `make test-integration` target to HITL Makefile and adds the agent to the CI matrix

## Test plan
- [x] `make test` still runs only unit tests (integration tests excluded via `--ignore`)
- [x] `make test-integration` runs the deployment test end-to-end on the demo cluster
- [x] Test tears down deployment even on failure (fixture finalizer)
- [x] JUnit XML output is produced at `results.xml`
- [x] CI workflow passes with both agents in the matrix — [run 25452473165](https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/25452473165)

Jira: [RHAIENG-4644](https://redhat.atlassian.net/browse/RHAIENG-4644)

[RHAIENG-4644]: https://redhat.atlassian.net/browse/RHAIENG-4644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/agents/langgraph/human_in_the_loop/tests/integration/test_deployment.py
+++ b/agents/langgraph/human_in_the_loop/tests/integration/test_deployment.py
@@ -20,7 +20,7 @@ INTERNAL_REGISTRY = "image-registry.openshift-image-registry.svc:5000"
 
 @pytest.fixture(scope="module")
 def agent_dir(repo_root):
-    return repo_root / "agents" / "langgraph" / "react_agent"
+    return repo_root / "agents" / "langgraph" / "human_in_the_loop"
 
 
 @pytest.fixture(scope="module")
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -11,6 +11,15 @@ import httpx
 
 logger = logging.getLogger(__name__)
 
+
+def load_agent_name(agent_dir: str | Path) -> str:
+    text = (Path(agent_dir) / "agent.yaml").read_text()
+    match = re.search(r"^name:\s*(.+)", text, re.MULTILINE)
+    if not match:
+        raise ValueError(f"No 'name' field in {agent_dir}/agent.yaml")
+    return match.group(1).strip()
+
+
 _REDACT_PATTERNS = [
     re.compile(r"(API_KEY=)\S+"),
     re.compile(r'(apiKey:\s*")[^"]*"'),
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -8,16 +8,16 @@ import time
 from pathlib import Path
 
 import httpx
+import yaml
 
 logger = logging.getLogger(__name__)
 
 
 def load_agent_name(agent_dir: str | Path) -> str:
-    text = (Path(agent_dir) / "agent.yaml").read_text()
-    match = re.search(r"^name:\s*(.+)", text, re.MULTILINE)
-    if not match:
+    data = yaml.safe_load((Path(agent_dir) / "agent.yaml").read_text())
+    if not isinstance(data, dict) or "name" not in data:
         raise ValueError(f"No 'name' field in {agent_dir}/agent.yaml")
-    return match.group(1).strip()
+    return str(data["name"]).strip()
 
 
 _REDACT_PATTERNS = [
--- a/.github/workflows/agent-deployment-test.yaml
+++ b/.github/workflows/agent-deployment-test.yaml
@@ -39,7 +39,7 @@ jobs:
           oc project
           echo ""
           echo "--- Namespace resources ---"
-          oc get all -n ci-testing --no-headers | head -10
+          oc get all -n ci-testing --no-headers 2>&1 | head -10 || true
           echo ""
           echo "Cluster connection verified successfully."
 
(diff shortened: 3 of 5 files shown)


tests/integration/utils.py
```diff
@@ -11,6 +11,15 @@
 
 logger = logging.getLogger(__name__)
 
+
+def load_agent_name(agent_dir: str | Path) -> str:
+    text = (Path(agent_dir) / "agent.yaml").read_text()
+    match = re.search(r"^name:\s*(.+)", text, re.MULTILINE)
+    if not match:
+        raise ValueError(f"No 'name' field in {agent_dir}/agent.yaml")
+    return match.group(1).strip()
```
tarun-etikala: **Regex may capture unintended content from `agent.yaml`**

The regex `r"^name:\s*(.+)"` will capture inline YAML comments (e.g., `name: foo # comment` → `"foo # comment"`) and preserve quotes around values (e.g., `name: "foo"` → `"\"foo\""`). Since this value feeds into the container image tag and route name, a mismatch would cause silent deployment failures.

Since `pyyaml` is already a transitive dependency, consider using `yaml.safe_load()` instead:

```python
import yaml

def load_agent_name(agent_dir: str | Path) -> str:
    data = yaml.safe_load((Path(agent_dir) / "agent.yaml").read_text())
    if not isinstance(data, dict) or "name" not in data:
        raise ValueError(f"No 'name' field in {agent_dir}/agent.yaml")
    return str(data["name"]).strip()
```

Note: the Makefile `AGENT_NAME` shell snippet uses the same regex pattern — worth aligning if you change the Python side, or accepting the divergence since Makefile values are simple in practice.

agents/langgraph/human_in_the_loop/tests/integration/test_deployment.py
```diff
@@ -0,0 +1,90 @@
...
+        env_path.unlink(missing_ok=True)
+
+
+@pytest.mark.integration
+def test_health_endpoint(deployed_agent):
+    route_url = deployed_agent
+    result = health_check(f"{route_url}/health", retries=12, backoff=5.0)
+
+    assert result["status"] == "healthy"
+    assert result["agent_initialized"] is True
```
tarun-etikala, line 90: **Significant duplication with `react_agent/tests/integration/test_deployment.py`**

This file is nearly a verbatim copy of the react_agent equivalent. The only distinguishing fixture is `agent_dir`. Everything else is identical:
- `INTERNAL_REGISTRY` constant
- `_write_env_file()` function
- `deployed_agent` fixture (build → deploy → yield route → undeploy)
- `test_health_endpoint` test body

As more agents are added to the CI matrix, each copy will need to be updated independently, and drift is likely (e.g., when HITL gains checkpointer DB env vars or the health response schema changes).

Consider extracting the shared scaffold into `tests/integration/utils.py` or the shared `conftest.py`, with each per-agent test module only defining `agent_dir` and any agent-specific assertions:

```python
# agents/langgraph/human_in_the_loop/tests/integration/test_deployment.py
# After extraction, this file would shrink to ~15 lines:

@pytest.fixture(scope="module")
def agent_dir(repo_root):
    return repo_root / "agents" / "langgraph" / "human_in_the_loop"

# agent_name and deployed_agent come from shared conftest

@pytest.mark.integration
def test_health_endpoint(deployed_agent):
    result = health_check(f"{deployed_agent}/health", retries=12, backoff=5.0)
    assert result["status"] == "healthy"
    assert result["agent_initialized"] is True
```

tarun-etikala (approved):
Looks good Overall. Added minor comments

tests/integration/utils.py
mpk-droid: Good catch — switched to `yaml.safe_load()` in b747417. The Makefile still uses the regex one-liner (can't easily import yaml there), but the values are simple in practice so the divergence is acceptable.

agents/langgraph/human_in_the_loop/tests/integration/test_deployment.py
mpk-droid: Agree this will need extraction — planning to do it when the third agent lands (likely the next tier-1 ticket). With only two agents right now, the duplication is manageable and keeping the test files self-contained makes them easier to read. Will extract `_write_env_file`, `deployed_agent`, and `INTERNAL_REGISTRY` into the shared conftest at that point.

tarun-etikala:
LGTM

PR #108, opened 2026-05-20
FEATURE : Deployment of Agent using Kagenti

shricharan-ks:
## Description

This feature setups the documentation and the claude skills for the deployment of an agent and using kagenti for agent visibility. 

## Jira Ticket

RHAIENG-5034

## Testing

- [x] No testing required (documentation/config change only)

## Checklist

### DIFF
--- a/.claude/skills/kagenti-deploy/SKILL.md
+++ b/.claude/skills/kagenti-deploy/SKILL.md
@@ -1,0 +1,638 @@
+---
+name: kagenti-deploy
+description: >
+  Deploy A2A-compliant agents to OpenShift/Kubernetes with kagenti integration. 
+  Use this skill when the user wants to deploy an agent with kagenti, mentions 
+  "kagenti deployment", "A2A agent deployment", "Agent Runtime", "Agent Card", 
+  or asks to deploy an agent that should be visible in kagenti UI. Also use 
+  when they mention the /.well-known/agent-card.json or /.well-known/agents.json 
+  endpoints, or ask about agent discovery and cataloging in OpenShift.
+---
+
+# Kagenti Agent Deployment
+
+Deploy A2A-protocol agents to OpenShift/Kubernetes with full kagenti integration, making them discoverable in the kagenti UI for agent management and orchestration.
+
+## What this skill does
+
+This skill orchestrates the complete deployment of A2A-compliant agents with kagenti:
+
+1. **Prepares the agent** - Initializes environment and dependencies
+2. **Builds and pushes container images** - Creates container images with proper A2A endpoints
+3. **Ensures kagenti labels** - Adds required labels to deployment manifests
+4. **Deploys to cluster** - Uses Helm to deploy the agent
+5. **Applies Agent Runtime CR** - Creates kagenti custom resource for agent discovery
+6. **Verifies integration** - Confirms Agent Card creation and kagenti synchronization
+
+## When to use this skill
+
+Use this skill when:
+- User wants to deploy an agent with kagenti integration
+- User mentions "kagenti", "A2A deployment", "Agent Runtime", or "Agent Card"
+- User asks to make an agent visible in kagenti UI
+- User wants to deploy an agent that implements the A2A protocol
+- User asks about agent discovery, cataloging, or orchestration on OpenShift
+
+## Prerequisites
+
+Before starting, verify these are available:
+- `oc` CLI configured and authenticated to the target cluster
+- `helm` (version 3.x)
+- `podman` or `docker` for building container images
+- `make` and `bash` available
+- Access to a container registry (e.g., quay.io)
+- Kagenti installed on the cluster (check: `oc get pods -n kagenti-system`)
+
+## A2A Protocol Requirements
+
+For an agent to work with kagenti, it must:
+- Expose `/.well-known/agent-card.json` endpoint
+- Return valid A2A agent card JSON with agent metadata
+- Implement A2A JSON-RPC protocol for agent-to-agent communication
+- Have health check endpoint at `/health`
+
+## Deployment Workflow
+
+### Step 1: Navigate to agent directory
+
+```bash
+cd agents/<framework>/<agent_name>/
+```
+
+The agent must have:
+- `Makefile` with standard targets
+- `agent.yaml` with agent metadata
+- `.env.example` or `template.env` for configuration template
+
+### Step 2: Initialize and configure
+
+```bash
+make init  # Creates .env from .env.example (or template.env for a2a-langgraph-crewai)
+```
+
+Edit `.env` to set required variables:
+- `API_KEY` - API key for LLM provider
+- `BASE_URL` - LLM provider endpoint (must end with `/v1`)
+- `MODEL_ID` - Model identifier
+- `CONTAINER_IMAGE` - Registry path where images will be pushed
+
+Verify required env vars are set:
+```bash
+# Check agent.yaml for required variables
+grep -A 10 "required:" agent.yaml
+```
+
+### Step 3: Set up Python environment
+
+```bash
+make env  # Creates venv and installs dependencies with uv
+```
+
+This runs `uv sync --python 3.12`.
+
+### Step 4: Build container image
+
+```bash
+make build
+```
+
+This:
+- Validates required env vars
+- Builds container image with proper tags
+- For A2A agents, may create multiple tags (e.g., `:crew` and `:langgraph`)
+
+Verify build succeeded:
+```bash
+podman images | grep <agent-name>
+# or
+docker images | grep <agent-name>
+```
+
+### Step 5: Push to registry
+
+```bash
+make push
+```
+
+Ensure you're authenticated to the registry first:
+```bash
+podman login quay.io
+# or
+docker login quay.io
+```
+
+### Step 6: Verify and enforce kagenti labels in deployment charts
+
+**CRITICAL**: The deployment charts must include protocol labels for discovery. This skill assumes agents use specialized charts (like `charts/a2a-langgraph-crewai/`) that include these labels.
+
+**Required labels in the chart templates**:
+```yaml
+metadata:
+  labels:
+    protocol.kagenti.io/a2a: "true"
+```
+
+**Optional protocol labels** (if agent supports multiple protocols):
+```yaml
+    protocol.kagenti.io/openai: "true"  # If OpenAI-compatible
+    protocol.kagenti.io/mcp: "true"     # If MCP-compatible
+```
+
+**Important**: The `kagenti.io/type: agent` label is **automatically added by the kagenti controller** when you create the AgentRuntime CR in Step 9. You do NOT need to add this label to the chart templates. Only the protocol labels are required in the chart.
+
+**Verify the chart has protocol labels**:
+
+First, identify which chart the agent uses by checking its Makefile:
+```bash
+grep "CHART_DIR" Makefile
+```
+
+Then verify the chart templates include protocol labels:
+```bash
+grep -r "protocol.kagenti.io" <chart-directory>/templates/
+```
+
+**Expected output** should show protocol labels in deployment files like:
+```yaml
+    protocol.kagenti.io/a2a: "true"
+```
+
+**If protocol labels are missing**:
+- Check if this is truly an A2A agent that needs kagenti integration
+- The agent may need a specialized chart - consult with the user
+- Standard `charts/agent/` template does not include kagenti labels by design
+
+**ENFORCEMENT**: If the grep returns no results for `protocol.kagenti.io`, STOP and inform the user that the chart templates do not include protocol labels. Ask if they want to:
+1. Create a specialized chart for this agent with protocol labels
+2. Add protocol labels to the existing chart
+3. Continue without kagenti integration
+
+### Step 7: Deploy with Helm
+
+**Detect and use the active namespace**: The deployment will use the current `oc` context namespace. Capture it for use in subsequent steps.
+
+```bash
+# Get the current active namespace
+NAMESPACE=$(oc project -q)
+echo "Deploying to namespace: $NAMESPACE"
+```
+
+If you need to switch to a different namespace first:
+```bash
+oc project <target-namespace>
+NAMESPACE=$(oc project -q)
+```
+
+Deploy the agent:
+```bash
+make deploy
+```
+
+**Note**: The Makefile deploy target uses the current `oc` context namespace. The same namespace will be used throughout all subsequent steps (labeling, CR creation, verification).
+
+This typically:
+- Deploys Helm chart with agent resources to the current namespace
+- For A2A agents, may do two-phase deploy (services/routes first, then deployments with public URLs)
+- Waits for pods to be ready
+
+**Verify deployment** in the target namespace:
+```bash
+oc get pods -n $NAMESPACE
+oc get routes -n $NAMESPACE
+```
+
+**Check health endpoint**:
+```bash
+ROUTE=$(oc get route <agent-name> -n $NAMESPACE -o jsonpath='{.spec.host}')
+curl https://$ROUTE/health
+# Expected: {"status":"healthy"}
+```
+
+**Verify A2A endpoint**:
+```bash
+curl -sS https://$ROUTE/.well-known/agent-card.json
+# Should return valid JSON with agent metadata
+```
+
+### Step 8: Label namespace for kagenti discovery
+
+Label the target namespace to enable kagenti discovery:
+```bash
+oc label namespace $NAMESPACE kagenti-enabled=true --overwrite
+```
+
+This tells kagenti to watch this namespace for agents.
+
+**Note**: The deployment labels should already be in the chart templates (verified in Step 6). If they're missing, the chart templates need to be updated before proceeding.
+
+### Step 9: Apply Agent Runtime CR
+
+Create and apply the AgentRuntime custom resource. This tells kagenti about the agent and **automatically creates the corresponding AgentCard CR**.
+
+**AgentRuntime structure**:
+```yaml
+apiVersion: agent.kagenti.dev/v1alpha1
+kind: AgentRuntime
+metadata:
+  name: <agent-name>-runtime
+  # namespace omitted - uses current oc context
+  labels:
+    app.kubernetes.io/name: <agent-name>
+spec:
+  type: agent
+  protocol: a2a
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: <deployment-name>
+```
+
+**Check if AgentRuntime.yaml exists in the agent directory** as a reference:
+```bash
+ls -la AgentRuntime.yaml
+```
+
+**Recommended approach**: Always create AgentRuntime inline to avoid mutating tracked files and ensure correct namespace:
+```bash
+# Read the existing file for deployment names if it exists, then create inline
+cat <<EOF | oc apply -f -
+apiVersion: agent.kagenti.dev/v1alpha1
+kind: AgentRuntime
+metadata:
+  name: <agent-name>-runtime
+  # namespace omitted - uses current oc context
+  labels:
+    app.kubernetes.io/name: <agent-name>
+spec:
+  type: agent
+  protocol: a2a
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: <deployment-name>
+EOF
+```
+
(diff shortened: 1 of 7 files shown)


charts/a2a-langgraph-crewai/templates/deployment-crew.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: a2a-crew-agent
   labels:
     app: a2a-crew-agent
+    protocol.kagenti.io/a2a: "true"
```
andrewdonheiser, line 8: **[Critical]** Missing `kagenti.io/type: agent` label.

The SKILL.md (lines 132, 156) explicitly documents that **both** `kagenti.io/type: agent` AND `protocol.kagenti.io/a2a: "true"` are required labels on deployments. The verification script (line 435) queries `oc get deployments -l kagenti.io/type=agent` and will report "No deployments with kagenti.io/type=agent found" without it.

But neither this file nor `deployment-langgraph.yaml` adds `kagenti.io/type: agent`. The `a2a.labels` helper in `_helpers.tpl` doesn't include it either. This means the PR's own verification will always fail, and kagenti may not discover these deployments.

**Suggested fix:** Add `kagenti.io/type: agent` to `metadata.labels` in both deployment templates:
```yaml
    kagenti.io/type: agent
    protocol.kagenti.io/a2a: "true"
```

agents/a2a/langgraph_crewai_agent/AgentRuntime.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: agent.kagenti.dev/v1alpha1
+kind: AgentRuntime
+metadata:
+  name: a2a-langgraph-agent-runtime
+  namespace: default
```
andrewdonheiser: **[Important]** Hardcoded `namespace: default` is a deployment pitfall.

Anyone running `oc apply -f AgentRuntime.yaml` (as the README instructs at line 209) will deploy to `default` instead of their target namespace. The README section doesn't mention updating the namespace before applying.

**Suggested fix:** Either:
1. Remove the `namespace` field entirely and let `oc apply -n <namespace>` control placement, or
2. Add a comment in the YAML noting users must update the namespace, AND add a note in the README about this.

.claude/skills/kagenti-deploy/SKILL.md
```diff
@@ -0,0 +1,746 @@
...
+
+**Check if AgentRuntime.yaml exists in the agent directory**:
+```bash
+ls -la AgentRuntime.yaml
+```
+
+If it exists, verify it points to the correct namespace and deployment, then apply it:
+```bash
+# Update namespace if needed
+sed -i "s/namespace: .*/namespace: $NAMESPACE/" AgentRuntime.yaml
```
andrewdonheiser: **[Important]** `sed -i` is not portable to macOS.

BSD sed on macOS requires a backup extension argument (e.g., `sed -i '' "s/..."` or `sed -i.bak`). Without it, sed interprets the substitution pattern as the backup extension and fails. This also appears at line 694. Since contributors use macOS, this will break.

**Suggested fix:** Either use `sed -i'' "s/..."` for portability, or better yet, recommend the `cat <<EOF | oc apply` approach exclusively and avoid mutating a tracked file.

agents/a2a/langgraph_crewai_agent/AgentRuntime.yaml
```diff
@@ -0,0 +1,29 @@
...
+  labels:
+    app.kubernetes.io/name: a2a-crew-agent
+spec:
+  type: agent
+  protocol: a2a
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: a2a-crew-agent
```
andrewdonheiser: **[Important]** Missing trailing newline.

The diff shows `\ No newline at end of file`. POSIX requires text files to end with a newline. This causes git diff noise and some YAML parsers may behave unexpectedly.

**Suggested fix:** Add a trailing newline after this line.

.claude/skills/kagenti-deploy/SKILL.md
```diff
@@ -0,0 +1,746 @@
...
+NAMESPACE="${1:-agents}"
+
+echo "========================================="
+echo "Kagenti Integration Verification"
+echo "Namespace: $NAMESPACE"
+echo "========================================="
+echo ""
+
+# Check namespace label
+echo "1️⃣  Checking namespace label..."
```
andrewdonheiser: **[Important]** Emojis in verification script violate project conventions.

The project CLAUDE.md says "Only use emojis if the user explicitly requests it." No existing skill in this repo uses emojis. This script and the success criteria section (lines 636-646) use emoji characters throughout (1️⃣, ✅, ❌, ⚠️, 🌐).

**Suggested fix:** Replace with plain text indicators: `[PASS]`/`[FAIL]`/`[WARN]`, numbered lists `1.`, `2.`, etc.

charts/a2a-langgraph-crewai/templates/deployment-langgraph.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: a2a-langgraph-agent
   labels:
     app: a2a-langgraph-agent
+    protocol.kagenti.io/a2a: "true"
```
andrewdonheiser, line 8: **[Important]** Label value `"true"` may not match upstream kagenti-operator convention.

The upstream kagenti-operator examples use `protocol.kagenti.io/a2a: ""` (empty string), not `"true"`. While `Exists` label selectors match either value, using a non-standard value risks breakage if the operator ever performs exact-match validation.

**Suggested fix:** Verify against upstream kagenti-operator docs and align the value. If upstream uses empty string, change to `protocol.kagenti.io/a2a: ""`.

(Same applies to `deployment-crew.yaml`.)

.claude/skills/kagenti-deploy/SKILL.md
```diff
@@ -0,0 +1,746 @@
+---
```
andrewdonheiser, line 1: **[Suggestion]** At 746 lines, this is 2.3x larger than the next largest skill in this repo.

Existing skills average ~161 lines. The verification script (~95 lines of bash) and the full deployment example (~55 lines) could be extracted into standalone files (e.g., `verify-kagenti.sh` committed alongside the skill, a reference doc in `docs/`). This would make the skill more maintainable and follow the pattern where skills are concise guides that reference external artifacts.

agents/a2a/langgraph_crewai_agent/README.md
```diff
@@ -201,6 +201,14 @@ make dry-run          # preview rendered Helm manifests (secrets redacted)
 make deploy
 ```
 
+### Manage Agents using Kagenti (Optional)
```
andrewdonheiser: **[Suggestion]** Heading level inconsistent with surrounding sections.

This uses `###` (h3), but the surrounding sub-sections (deploy, dry-run, verify, undeploy) all use `####` (h4). This makes kagenti appear as a top-level deployment section rather than a sub-step.

**Suggested fix:** Use `#### Manage Agents using Kagenti (Optional)` (h4) to match.

agents/a2a/langgraph_crewai_agent/README.md
```diff
@@ -201,6 +201,14 @@ make dry-run          # preview rendered Helm manifests (secrets redacted)
...
+
+Apply the `AgentRuntime` custom resources to enable management via Kagenti. The AgentRuntime will be picked up by kagenti and create AgentCards for the agent deployments.
```
andrewdonheiser: **[Suggestion]** Section lacks prerequisite and verification context.

This doesn't mention:
- That kagenti must be installed on the cluster (`oc get pods -n kagenti-system`)
- That the namespace needs the `kagenti-enabled=true` label
- That users must update the `namespace` field in the YAML before applying (if kept)
- How to verify success (`oc get agentcard -n <namespace>`)

A brief note on prerequisites and a verification command would prevent common failures.

agents/a2a/langgraph_crewai_agent/pyproject.toml
```diff
@@ -5,7 +5,7 @@ description = "A2A demo: LangGraph orchestrator and CrewAI specialist for OpenSh
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "a2a-sdk[http-server]>=0.3.25",
+    "a2a-sdk[http-server]<1.0.0,>=0.3.25",
```
andrewdonheiser: **[Suggestion]** Version specifier ordering.

The PEP 440 convention is typically `>=lower,<upper` (lower bound first), not `<upper,>=lower`. Both are functionally identical, but `"a2a-sdk[http-server]>=0.3.25,<1.0.0"` reads more naturally.

The defensive upper bound itself is a good practice for a pre-1.0 library 👍

.claude/skills/kagenti-deploy/SKILL.md
```diff
@@ -0,0 +1,746 @@
...
+### Step 1: Navigate to agent directory
+
+```bash
+cd agents/<framework>/<agent_name>/
+```
+
+The agent must have:
+- `Makefile` with standard targets
+- `agent.yaml` with agent metadata
+- `.env.example` or `template.env` for configuration template
```
andrewdonheiser, line 65: **[Suggestion]** Doesn't clarify this agent uses `template.env` instead of `.env.example`.

The a2a-langgraph-crewai agent is a known non-standard agent (per AGENTS.md) that uses `template.env` instead of `.env.example`. Step 2's `make init` doesn't mention this explicitly, which may confuse users who look for `.env.example` and don't find it.

.claude/skills/kagenti-deploy/SKILL.md
```diff
@@ -0,0 +1,746 @@
...
+NAMESPACE="<user-specified-namespace>"
+```
+
+Deploy the agent:
+```bash
+make deploy
+```
+
+This typically:
+- Creates namespace if needed (or specify with make deploy NAMESPACE=$NAMESPACE if Makefile supports it)
```
andrewdonheiser: **[Suggestion]** `make deploy NAMESPACE=$NAMESPACE` — the Makefile doesn't accept a NAMESPACE parameter.

The Makefile's `deploy` target doesn't parameterize the namespace — it relies on whatever namespace is set in the Helm release or the current `oc` context. The parenthetical hedge "(if Makefile supports it)" is there but could still mislead.

**Suggested fix:** Clarify that namespace is controlled via `oc project $NAMESPACE` or Helm values, not via a Make variable.

andrewdonheiser (changes requested):
## PR Review Summary

Thanks for adding kagenti deployment support! The SKILL.md is comprehensive and the overall approach is solid. There are a few issues to address before this can merge — most notably a missing label that will prevent kagenti discovery from working.

### Findings: 1 Critical, 5 Important, 6 Suggestions

See inline comments for details.

### Strengths
- Comprehensive end-to-end workflow with logical step progression
- Good enforcement gate pattern (Claude stops and asks if labels are missing)
- Correct multi-deployment AgentRuntime handling for langgraph + crew
- Accurate AgentRuntime/AgentCard relationship documentation
- Defensive `a2a-sdk` version pin aligns with repo patterns
- Properly scoped — kagenti is optional, existing behavior unchanged

charts/a2a-langgraph-crewai/templates/deployment-crew.yaml
shricharan-ks: This is a good catch @andrewdonheiser .
The label `kagenti.io/type: agent` will be automatically added to the deployment when we apply the AgentRuntime CR to the deployments.

We can manually add those labels, but it would be hard to verify if kagenti has done something with the deployments.

.claude/skills/kagenti-deploy/SKILL.md
andrewdonheiser: Addressed -- SKILL.md now recommends inline `cat <<EOF | oc apply` instead of mutating tracked files.

.claude/skills/kagenti-deploy/SKILL.md
andrewdonheiser: Addressed -- verification script now uses plain text indicators (`[PASS]`, `[FAIL]`, `[WARN]`).

agents/a2a/langgraph_crewai_agent/README.md
andrewdonheiser: Addressed -- heading changed to `####` to match surrounding sections.

agents/a2a/langgraph_crewai_agent/README.md
andrewdonheiser: Addressed -- README now includes kagenti prerequisites, namespace update step, and verification command. Nice improvement.

.claude/skills/kagenti-deploy/SKILL.md
andrewdonheiser: Addressed -- Step 2 now clarifies `template.env` for a2a-langgraph-crewai.

.claude/skills/kagenti-deploy/SKILL.md
andrewdonheiser: Addressed -- Step 7 now uses `oc project` to manage namespace instead of `make deploy NAMESPACE=`.

agents/a2a/langgraph_crewai_agent/pyproject.toml
andrewdonheiser: Ordering addressed, but there is now a trailing comma: `"a2a-sdk[http-server]>=0.3.25,<1.0.0,"` — that trailing comma after `<1.0.0` is invalid PEP 440 syntax and will cause pip/uv to fail. Please remove it.

agents/a2a/langgraph_crewai_agent/pyproject.toml
```diff
@@ -5,7 +5,7 @@ description = "A2A demo: LangGraph orchestrator and CrewAI specialist for OpenSh
 readme = "README.md"
 requires-python = ">=3.12"
 dependencies = [
-    "a2a-sdk[http-server]>=0.3.25",
+    "a2a-sdk[http-server]>=0.3.25,<1.0.0,",
```
andrewdonheiser: **[Critical]** Trailing comma inside the version specifier.

```python
"a2a-sdk[http-server]>=0.3.25,<1.0.0,"
#                                    ^ this comma is inside the quotes — invalid PEP 440
```

This was introduced while reordering the version bounds. The trailing comma will cause pip/uv to fail during dependency resolution.

**Fix:** Move the comma outside the closing quote:
```python
"a2a-sdk[http-server]>=0.3.25,<1.0.0",
```

andrewdonheiser (commented):
Re-review: 7 of 12 items from the first review have been addressed — resolved those threads. One new issue found in the fix-up commit (see inline).

charts/a2a-langgraph-crewai/templates/deployment-crew.yaml
andrewdonheiser, line 8: **[Critical]** Doc/code inconsistency on `kagenti.io/type: agent` label.

The SKILL.md Step 6 ("Required labels on the Deployment") documents **both** `kagenti.io/type: agent` and `protocol.kagenti.io/a2a: "true"` as required in the chart templates, and the enforcement gate says to STOP if grep finds no results. But the chart only adds `protocol.kagenti.io/a2a`.

You mentioned the controller auto-adds `kagenti.io/type: agent` via the AgentRuntime CR — if that's the case, the SKILL.md needs to be updated so Step 6 only checks for `protocol.kagenti.io/a2a` and explains that `kagenti.io/type` is added automatically by the controller. Otherwise the skill will incorrectly halt deployments that are actually fine.

Either:
1. Update SKILL.md to reflect that `kagenti.io/type: agent` is controller-managed (not chart-managed), or
2. Add the label to both chart templates if it should be present at deploy time regardless

agents/a2a/langgraph_crewai_agent/AgentRuntime.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: agent.kagenti.dev/v1alpha1
+kind: AgentRuntime
+metadata:
+  name: a2a-langgraph-agent-runtime
+  namespace: default
```
andrewdonheiser: **[Important]** `namespace: default` is still a foot-gun.

The README update helps (telling users to change the field), but `default` as a committed value means anyone who runs `oc apply -f AgentRuntime.yaml` without reading the README will deploy to the wrong namespace.

Suggested fix — remove the `namespace` field entirely so `oc apply` uses the current context:

```yaml
metadata:
  name: a2a-langgraph-agent-runtime
  # namespace is omitted — uses current oc context
  labels:
    app.kubernetes.io/name: a2a-langgraph-agent
```

Same for the crew agent block on line 19. This also aligns with the SKILL.md recommendation to use inline `cat <<EOF` with `$NAMESPACE`.

agents/a2a/langgraph_crewai_agent/AgentRuntime.yaml
```diff
@@ -0,0 +1,30 @@
...
+  labels:
+    app.kubernetes.io/name: a2a-crew-agent
+spec:
+  type: agent
+  protocol: a2a
+  targetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: a2a-crew-agent
+
```
andrewdonheiser: **[Important]** Missing trailing newline.

Still showing `\ No newline at end of file` in the diff. POSIX requires text files to end with a newline — without it you get git diff noise and some YAML parsers may behave unexpectedly. Just add a newline after the last line.

charts/a2a-langgraph-crewai/templates/deployment-langgraph.yaml
andrewdonheiser, line 8: **[Suggestion]** Label value `"true"` vs upstream convention.

Upstream kagenti-operator examples use `protocol.kagenti.io/a2a: ""` (empty string) rather than `"true"`. Both likely work, but worth confirming against the operator version on your target cluster to avoid a discovery miss. Non-blocking — just flag for verification.

.claude/skills/kagenti-deploy/SKILL.md
```diff
@@ -0,0 +1,740 @@
+---
```
andrewdonheiser, line 1: **[Suggestion]** Skill is 740 lines — 2.3x larger than the next largest skill in this repo.

Not blocking, but the verification script (~95 lines of bash) and the full deployment example (~60 lines) could be moved to separate files referenced from the skill. This would make the core workflow easier to maintain and review. Consider for a follow-up.

andrewdonheiser (changes requested):
## Re-review after fix-up commits

8 of 13 original findings have been addressed — nice work on the turnaround. 5 items remain (1 critical, 2 important, 2 suggestions). See inline comments.

### Resolved
- `sed -i` macOS portability — SKILL.md now uses inline `cat <<EOF`
- Emojis in verification script — now uses `[PASS]`/`[FAIL]`/`[WARN]`
- Heading level inconsistency — changed to `####`
- README lacking prerequisites/verification — added
- `template.env` clarification — added to Step 2
- `make deploy NAMESPACE=` — now uses `oc project`
- Version specifier ordering — now `>=0.3.25,<1.0.0`
- Trailing comma in version specifier — fixed

charts/a2a-langgraph-crewai/templates/deployment-langgraph.yaml
shricharan-ks: Setting the label value to `"true"` is more of a general practice in kubernetes, so I was trying to follow the same for this example.

andrewdonheiser (approved):
All findings from both review rounds have been addressed. Nice work on the turnaround.

**Verified resolved:**
- Version specifier fixed (trailing comma removed)
- `namespace` field removed from AgentRuntime.yaml (uses oc context)
- SKILL.md correctly documents controller-managed vs chart-managed labels
- Verification script extracted, complete flow removed
- All other items (sed portability, emojis, heading levels, README prerequisites, etc.)

CI is green. LGTM.

shricharan-ks:
> All findings from both review rounds have been addressed. Nice work on the turnaround.
> 
> **Verified resolved:**
> 
> * Version specifier fixed (trailing comma removed)
> * `namespace` field removed from AgentRuntime.yaml (uses oc context)
> * SKILL.md correctly documents controller-managed vs chart-managed labels
> * Verification script extracted, complete flow removed
> * All other items (sed portability, emojis, heading levels, README prerequisites, etc.)
> 
> CI is green. LGTM.

Thanks @andrewdonheiser  your comments helped in maintaining the quality standards. Much appreciated 🎉

PR #111, opened 2026-05-20
fix: add missing stream parameter to autogen_mcp run_eval fixture

andrewdonheiser:
## Summary
- Adds `stream: bool = False` to the `_run()` inner function in the autogen_mcp `run_eval` fixture, aligning its signature with all other agents (react_agent, agentic_rag, etc.)
- The parameter is accepted for interface consistency but not wired to `TaskConfig` — autogen always uses non-streaming for evals since tool_invocations are only exposed in non-streaming JSON responses

## Test plan
- [x] Verify existing autogen_mcp behavioral tests still pass
- [x] Confirm `_run()` signature matches react_agent and agentic_rag patterns

Closes RHAIENG-5127


mpk-droid (approved):
lgtm

PR #156, opened 2026-06-01
fix(langflow): bump Langflow to 1.9.5 (RHAIENG-4293)

tarun-etikala:
## Description

Bump Langflow from **1.7.1 to 1.9.5** to resolve critical CVEs and fix all breaking changes introduced in 1.9.

### CVEs resolved
- **CVE-2026-33017** (CVSS 9.8, CISA KEV) — Unauthenticated RCE via `/api/v1/build_public_tmp/` endpoint
- **CVE-2025-34291** (CVSS 9.4, CISA KEV) — Account takeover + RCE via CORS misconfiguration
- Multiple dependency CVEs fixed in 1.9.3/1.9.4

### Breaking changes addressed

| Change in 1.9 | Impact | Fix |
|---|---|---|
| `LANGFLOW_SKIP_AUTH_AUTO_LOGIN` renamed | Auth required without fix | Renamed to `LANGFLOW_AUTO_LOGIN` |
| SSRF protection blocks private IPs by default | `host.containers.internal` blocked | Added to `LANGFLOW_SSRF_ALLOWED_HOSTS` |
| Components use Global Variables, not OS env vars | `OLLAMA_BASE_URL` not found by flow | Added `LANGFLOW_VARIABLES_TO_GET_FROM_ENVIRONMENT` |
| Agent component restructured | Old `agent_llm` field removed | Flow re-exported from 1.9.4 |
| `LANGFUSE_HOST` renamed to `LANGFUSE_BASE_URL` | Still works via fallback | Updated to preferred name ([langflow#12154](https://github.com/langflow-ai/langflow/pull/12154)) |
| Langfuse SDK dropped from Docker image (Python 3.14) | No traces in Langfuse | Install `pydantic>=2.13` + `langfuse~=3.8` at startup ([langfuse#9618](https://github.com/langfuse/langfuse/issues/9618), [langflow#13341](https://github.com/langflow-ai/langflow/pull/13341)) |

### Changes by file

**`podman-compose.yml`** (small, reviewable):
- Image tag `1.7.1` → `1.9.5`
- Startup command installs `pydantic>=2.13` + `langfuse~=3.8` (fixes Langfuse tracing on Python 3.14)
- `LANGFLOW_SKIP_AUTH_AUTO_LOGIN` → `LANGFLOW_AUTO_LOGIN`
- Added `LANGFLOW_SSRF_ALLOWED_HOSTS=host.containers.internal`
- Added `LANGFLOW_VARIABLES_TO_GET_FROM_ENVIRONMENT` for model endpoint vars
- `LANGFUSE_HOST` → `LANGFUSE_BASE_URL`

**`outdoor-activity-agent.json`** (full re-export — review as replacement, not line-by-line):
- Re-exported from running Langflow 1.9 instance
- Agent component rewritten with unified `ModelInput` replacing `agent_llm` dropdown
- Chat Output gains token usage accumulation
- Removed instance-specific `id` field and name suffix
- KServe vLLM defaults re-populated for local dev with LlamaStack

## Jira Ticket

RHAIENG-4293

## Testing

**Manual testing performed:**
1. `make stop && make run` — Langflow 1.9.5 starts successfully
2. Flow imports without errors
3. LlamaStack reachable from container via `host.containers.internal:8321`
4. Langfuse tracing confirmed — 3 traces (including `AgentExecutor`) visible in Langfuse UI at `localhost:3000`
5. Verified `pydantic>=2.13` + `langfuse~=3.8` install resolves the Python 3.14 / Langfuse SDK incompatibility

## Checklist

## Review Guidance

- **Commit 1** (`podman-compose.yml`): Small, reviewable — all the env var and image changes
- **Commit 2** (`outdoor-activity-agent.json`): Large diff (700+ lines) because Langflow re-exports the entire flow JSON with new node IDs and updated component code. Review as a whole-file replacement, not line-by-line. Key semantic changes are listed above.
- **Commits 3-4**: Langfuse tracing fix and version bump to 1.9.5

## Related PRs

- Upstream fix: [langflow#13341](https://github.com/langflow-ai/langflow/pull/13341) — `fix(tracing): surface Langfuse setup failures, pin pydantic>=2.13 (Py3.14)` (merged into `release-1.10.0`, not yet backported to 1.9.x)
- Upstream fix: [langflow#13344](https://github.com/langflow-ai/langflow/pull/13344) — `fix(tracing): isolate Langfuse OTel tracer to stop HTTP span pollution`
- Upstream: [langfuse#9618](https://github.com/langfuse/langfuse/issues/9618) — Python 3.14 incompatibility with Langfuse SDK

### DIFF
--- a/agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
+++ b/agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
@@ -4,6 +4,8 @@ services:
   # ── Langflow (UI + Server) ─────────────────────────────────────────
   langflow:
     image: docker.io/langflowai/langflow:1.9.4
+    command: >
+      sh -c "pip install --quiet langfuse && langflow run"
     ports:
       - "7860:7860"
     environment:
--- a/agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
+++ b/agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
@@ -3,13 +3,20 @@ version: "3.8"
 services:
   # ── Langflow (UI + Server) ─────────────────────────────────────────
   langflow:
-    image: docker.io/langflowai/langflow:1.7.1
+    image: docker.io/langflowai/langflow:1.9.4
     ports:
       - "7860:7860"
     environment:
       LANGFLOW_DATABASE_URL: "postgresql://${POSTGRES_USER:-langflow}:${POSTGRES_PASSWORD:-langflow}@postgres:5432/${POSTGRES_DB:-langflow}"
       LANGFLOW_CONFIG_DIR: /var/lib/langflow
-      LANGFLOW_SKIP_AUTH_AUTO_LOGIN: "true"
+      LANGFLOW_AUTO_LOGIN: "true"
+      # Allow Langflow to reach host-local services (Ollama, LlamaStack).
+      # SSRF protection is on by default in 1.9+ and blocks private IPs.
+      LANGFLOW_SSRF_PROTECTION_ENABLED: "true"
+      LANGFLOW_SSRF_ALLOWED_HOSTS: "host.containers.internal"
+      # Import these env vars as Langflow Global Variables so flow
+      # components (Agent, KServe vLLM, etc.) can resolve them.
+      LANGFLOW_VARIABLES_TO_GET_FROM_ENVIRONMENT: "OPENAI_API_BASE,OPENAI_API_KEY,OLLAMA_BASE_URL"
       # Model endpoint — points to Ollama running on the host machine (not in a container)
       OPENAI_API_BASE: "http://host.containers.internal:11434/v1"
       OPENAI_API_KEY: "not-needed-for-local-development"
--- a/agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
+++ b/agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
@@ -3,9 +3,9 @@ version: "3.8"
 services:
   # ── Langflow (UI + Server) ─────────────────────────────────────────
   langflow:
-    image: docker.io/langflowai/langflow:1.9.4
+    image: docker.io/langflowai/langflow:1.9.5
     command: >
-      sh -c "pip install --quiet langfuse && langflow run"
+      sh -c "pip install --quiet 'pydantic>=2.13' 'langfuse~=3.8' && langflow run"
     ports:
       - "7860:7860"
     environment:
@@ -24,7 +24,7 @@ services:
       OPENAI_API_KEY: "not-needed-for-local-development"
       OLLAMA_BASE_URL: "http://host.containers.internal:11434"
       # Langfuse tracing
-      LANGFUSE_HOST: "http://langfuse-web:3000"
+      LANGFUSE_BASE_URL: "http://langfuse-web:3000"
       LANGFUSE_PUBLIC_KEY: "pk-lf-local-dev"
       LANGFUSE_SECRET_KEY: "sk-lf-local-dev"
     volumes:
--- a/agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
+++ b/agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
@@ -3,11 +3,9 @@ version: "3.8"
 services:
   # ── Langflow (UI + Server) ─────────────────────────────────────────
   langflow:
-    image: docker.io/langflowai/langflow:1.9.5
-    command: >
-      sh -c "pip install --quiet 'pydantic>=2.13' 'langfuse~=3.8' && langflow run"
+    build: .
     ports:
-      - "7860:7860"
+      - "127.0.0.1:7860:7860"
     environment:
       LANGFLOW_DATABASE_URL: "postgresql://${POSTGRES_USER:-langflow}:${POSTGRES_PASSWORD:-langflow}@postgres:5432/${POSTGRES_DB:-langflow}"
       LANGFLOW_CONFIG_DIR: /var/lib/langflow
--- a/agents/langflow/simple_tool_calling_agent/flows/outdoor-activity-agent.json
+++ b/agents/langflow/simple_tool_calling_agent/flows/outdoor-activity-agent.json
@@ -1,14 +1,183 @@
 {
-  "name": "Outdoor Activity Agent",
-  "description": "Find outdoor activities in a specific area based on nearby national parks, upcoming weather, and park alerts.",
-  "icon": null,
-  "icon_bg_color": null,
-  "gradient": null,
   "data": {
+    "edges": [
+      {
+        "animated": false,
+        "className": "",
+        "data": {
+          "sourceHandle": {
+            "dataType": "Agent",
+            "id": "Agent-DBGIm",
+            "name": "response",
+            "output_types": [
+              "Message"
+            ]
+          },
+          "targetHandle": {
+            "fieldName": "input_value",
+            "id": "ChatOutput-FpxbL",
+            "inputTypes": [
+              "Data",
+              "JSON",
+              "DataFrame",
+              "Table",
+              "Message"
+            ],
+            "type": "other"
+          }
+        },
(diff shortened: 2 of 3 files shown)


agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
```diff
@@ -3,19 +3,28 @@ version: "3.8"
 services:
   # ── Langflow (UI + Server) ─────────────────────────────────────────
   langflow:
-    image: docker.io/langflowai/langflow:1.7.1
+    image: docker.io/langflowai/langflow:1.9.5
+    command: >
+      sh -c "pip install --quiet 'pydantic>=2.13' 'langfuse~=3.8' && langflow run"
```
andrewdonheiser: **[Important] Runtime `pip install` without hash pinning — supply chain + startup fragility**

This runs on every container start with range specifiers (`>=2.13`, `~=3.8`) that pull untested future versions and no hash verification. If PyPI is unreachable, the container won't start at all.

**Suggested fix:** Pin exact versions with hashes, or add a thin Dockerfile layer:
```dockerfile
FROM langflowai/langflow:1.9.5
RUN pip install 'pydantic==2.13.2' 'langfuse==3.8.1'
```
This bakes deps into the image layer (cached, reproducible, no network dependency at startup). If a Dockerfile is too heavy for this starter kit, at minimum pin exact versions here.

agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
```diff
@@ -3,19 +3,28 @@ version: "3.8"
...
     ports:
       - "7860:7860"
```
andrewdonheiser: **[Important] Port bound to `0.0.0.0` with auth disabled — network-exposed unauthenticated Langflow**

`LANGFLOW_AUTO_LOGIN: "true"` (line 14) disables all authentication. Combined with this port binding, any device on the local network can access the Langflow UI and execute arbitrary flows — including the custom NPS/OpenMeteo components that make outbound HTTP requests.

**Suggested fix:** Bind to localhost only:
```yaml
ports:
  - "127.0.0.1:7860:7860"
```
This matches the security posture of the Langfuse service in this same compose file (line ~51), which already binds to `127.0.0.1`.

agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
```diff
@@ -3,19 +3,28 @@ version: "3.8"
...
     environment:
       LANGFLOW_DATABASE_URL: "postgresql://${POSTGRES_USER:-langflow}:${POSTGRES_PASSWORD:-langflow}@postgres:5432/${POSTGRES_DB:-langflow}"
       LANGFLOW_CONFIG_DIR: /var/lib/langflow
-      LANGFLOW_SKIP_AUTH_AUTO_LOGIN: "true"
+      LANGFLOW_AUTO_LOGIN: "true"
+      # Allow Langflow to reach host-local services (Ollama, LlamaStack).
+      # SSRF protection is on by default in 1.9+ and blocks private IPs.
+      LANGFLOW_SSRF_PROTECTION_ENABLED: "true"
+      LANGFLOW_SSRF_ALLOWED_HOSTS: "host.containers.internal"
```
andrewdonheiser, line 16: **[Important] Verify SSRF allowlist doesn't block the flow's external API calls**

The custom components in this flow call `developer.nps.gov` and `api.open-meteo.com` via `requests.get()`. If Langflow 1.9's SSRF protection intercepts *all* outbound HTTP from components (not just private-IP ranges), these public API calls would be silently blocked.

Worth confirming that `LANGFLOW_SSRF_ALLOWED_HOSTS` only governs private/internal IP access and that public endpoints are unaffected. If it applies broadly, add:
```yaml
LANGFLOW_SSRF_ALLOWED_HOSTS: "host.containers.internal,developer.nps.gov,api.open-meteo.com"
```

andrewdonheiser (commented):
## PR Review: Langflow 1.7.1 → 1.9.5

The CVE fixes are critical and well-justified. The migration is thorough — all 7 flow nodes and 6 edges are preserved correctly in the re-exported JSON. Custom component code (NPS, OpenMeteo) is byte-for-byte identical. Good inline comments explaining the *why* behind each config change.

**3 Important findings** posted as inline comments above. Additionally:

### Suggestion: PR title says "1.9.4" but actual version is 1.9.5
The image tag, `lf_version` fields, and `last_tested_version` all reference 1.9.5. Consider updating the title to match.

### Suggestion: Agent `model` field defaults to `gpt-5.4` (OpenAI)
The Agent node's `model` template field has `value: [{"name": "gpt-5.4", "provider": "OpenAI"}]`. At runtime this is overridden by the KServeVLLM edge connection, so it works correctly. However, if someone accidentally disconnects that edge in the Langflow UI, the flow silently falls back to calling OpenAI — which fails without an API key, producing a confusing error. Consider setting this value to empty/null so a missing model connection fails loudly.

### Suggestion: `.env.example` not updated
No diff touches `.env.example`. Users won't discover the new SSRF, Langfuse rename, or Global Variables config without reading the compose file. Consider adding placeholders for `LANGFLOW_SSRF_ALLOWED_HOSTS` and `LANGFLOW_VARIABLES_TO_GET_FROM_ENVIRONMENT`.

andrewdonheiser (changes requested):

agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
tarun-etikala: Good catch. Moved to a local Dockerfile that extends the upstream image with the deps baked in — no more runtime pip install. This eliminates the network dependency at startup and gives a reproducible image layer.

agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
tarun-etikala: Fixed — bound to 127.0.0.1:7860:7860 to match the Langfuse service binding in the same compose file.

agents/langflow/simple_tool_calling_agent/local/podman-compose.yml
tarun-etikala: Verified — SSRF protection only blocks private/internal IP ranges (RFC 1918, link-local, loopback). Public endpoints like developer.nps.gov and api.open-meteo.com are unaffected. Confirmed by running the flow end-to-end with all 3 tools (Open-Meteo, NPS Search, NPS Alerts) succeeding.

tarun-etikala:
Thanks for the thorough review @andrewdonheiser! Addressing each point:

**All 3 inline comments fixed** in the next push:
1. **pip install → Dockerfile** — Moved deps to a local `Dockerfile` that extends the upstream image. No runtime pip install, reproducible image layer, no network dependency at startup.
2. **Port binding** — Changed to `127.0.0.1:7860:7860` to match the Langfuse service's localhost binding.
3. **SSRF** — Verified: SSRF protection only blocks private IP ranges. Public API calls (NPS, Open-Meteo) work fine. Confirmed end-to-end.

**Re: PR title** — Updated to "fix(langflow): bump Langflow to 1.9.5 (RHAIENG-4293)".

**Re: Agent `model` defaults to `gpt-5.4`** — This is Langflow 1.9's default behavior on re-export. The Agent component always populates a default provider in its `model` field. The KServe vLLM edge connection overrides it at runtime (confirmed working). Setting to empty/null risks breaking the flow if someone opens it in the UI without the edge — Langflow requires a model selection to render the component. Leaving as-is since the edge connection is the intended usage.

**Re: `.env.example`** — The new env vars (`LANGFLOW_SSRF_*`, `LANGFLOW_VARIABLES_*`, `LANGFLOW_AUTO_LOGIN`) are Langflow server config, not user credentials. They live in the compose file (infrastructure config) rather than `.env.example` (which holds user-specific secrets like Langfuse passwords). This follows the existing pattern where `.env.example` only has PostgreSQL and Langfuse credentials.

andrewdonheiser (dismissed):

andrewdonheiser (approved):
👍

PR #173, opened 2026-06-10
docs(openclaw): Add guide for OGX→vLLM inference path

robinnarsinghranabhat:
## Summary
- **Add deployment guide** (`docs/raw-deployment.md`) for the OpenClaw → OGX → vLLM path, tested end-to-end with port-forwarding and interacting through OpenClaw Chat UI.
- **Update Raw manifest deployment** for OpenClaw on OpenShift — Openclaw installed through Raw manifests resulted in a crashing `openclaw gateway` pod. Updated them in reference to [official upstream k8s scripts](https://github.com/openclaw/openclaw/tree/main/scripts/k8s) and [OpenClaw VLLM setup](https://docs.openclaw.ai/providers/vllm)

## Test plan

- [x] Deployed openclaw in OpenShift 4.19 (ROSA) with existing ogx and vllm setup 
       (
- openclaw image : `ghcr.io/openclaw/openclaw@sha256:037f49ba1595be9502fb345138d727cd0cfaecf1392cbe0cfe053fb4681386cd`, 
- ogx image : `quay.io/rhoai/odh-ogx-core-rhel9:rhoai-3.5-ea.1`, 
- vllm image : `registry.redhat.io/rhaii/vllm-cuda-rhel9:3.4.0-1777444689`) 
- [x] Port-forwarded and accessed Control UI at `http://localhost:18789`
- [x] Tested tool calling capacity of OpenClaw → OGX → vLLM through the UI

### DIFF
--- a/agents/openclaw/deployment/README.md
+++ b/agents/openclaw/deployment/README.md
@@ -25,20 +25,7 @@ npm install && npm run build && npm run dev
 
 See [docs/installer-deployment.md](docs/installer-deployment.md) for the full walkthrough, validation, and rollback procedures.
 
-### Option B: Kustomize
-
-```bash
-oc new-project my-openclaw
-
-# Copy and customize the example overlay
-cp -r overlays/example overlays/my-env
-# Edit overlays/my-env/configmap-patch.yaml with your vLLM endpoint
-# Edit overlays/my-env/kustomization.yaml with your namespace and storage class
-
-oc apply -k overlays/my-env
-```
-
-### Option C: Direct YAML apply
+### Option B: Raw Manifests (Kustomize or direct apply)
 
 ```bash
 oc new-project my-openclaw
@@ -48,6 +35,8 @@ oc new-project my-openclaw
 
 oc apply -k manifests/
 ```
+
+See [docs/raw-deployment.md](docs/raw-deployment.md) for the full walkthrough, configuration details, and troubleshooting.
 
 ## Architecture
 
@@ -94,6 +83,7 @@ oc apply -k manifests/
 | Document | Description |
 |----------|-------------|
 | [docs/installer-deployment.md](docs/installer-deployment.md) | Full deployment guide: prerequisites, validation, rollback, appendix |
+| [docs/raw-deployment.md](docs/raw-deployment.md) | Step-by-step guide for deploying with raw Kustomize manifests |
 | [docs/troubleshooting.md](docs/troubleshooting.md) | Common issues and fixes (route 503, model override, heartbeat, config clobber) |
 | [docs/model-compatibility.md](docs/model-compatibility.md) | Model testing results for agentic tool-calling |
 
--- a/agents/openclaw/deployment/docs/raw-deployment.md
+++ b/agents/openclaw/deployment/docs/raw-deployment.md
@@ -65,7 +65,7 @@ Edit `manifests/02-configmap.yaml` and replace the placeholder values:
 - `YOUR-VLLM-OR-OGX-ENDPOINT` — your vLLM or OGX endpoint hostname
 - `YOUR-MODEL-ID` — the model ID from the `/v1/models` query above
 
-**Via OGX gateway** — OGX prefixes model IDs with `vllm/`, so use that full ID:
+**Via OGX gateway** — OGX prefixes model IDs with `vllm/`, so `agents.defaults.model` ends up as `vllm/vllm/<model>` (provider prefix + OGX model ID — the double `vllm/` is intentional):
 
 ```json
 "agents": {
@@ -173,7 +173,6 @@ oc port-forward deployment/openclaw 18789:18789 -n my-openclaw
 Open <http://localhost:18789> in your browser. Paste the gateway token from Step 2 when prompted.
 
 On first connect, device pairing is auto-approved for local connections. You should see the chat interface ready to use.
-
 
 ## References
 
--- a/agents/openclaw/deployment/docs/raw-deployment.md
+++ b/agents/openclaw/deployment/docs/raw-deployment.md
@@ -54,7 +54,7 @@ curl -s https://<your-endpoint>/v1/models | python3 -m json.tool
 }
 ```
 
-Use the **exact** `id` value from the response in your ConfigMap — both in `agents.defaults.model` and in `models.providers.vllm.models[].id`. Getting this wrong produces a `404 Model not found` error at runtime.
+Use the **exact** `id` value from the response as `models[].id` in your ConfigMap. For `agents.defaults.model`, prepend `vllm/` (the provider name) to the model ID.
 
 ## Configuration
 
@@ -111,15 +111,25 @@ Edit `manifests/02-configmap.yaml` and replace the placeholder values:
 
 Notice the `agents.defaults.model` field uses the format `<provider>/<model-id>`, while the `models[].id` is the raw ID sent to the endpoint in API requests.
 
-### Step 2: Set the gateway token
+### Step 2: Set secrets
 
-You might want to update `OPENCLAW_GATEWAY_TOKEN` in `manifests/01-secret.yaml` with a generated token.
+Edit `manifests/01-secret.yaml`:
 
-You will need this token to authenticate with the Control UI.
+- `OPENCLAW_GATEWAY_TOKEN` — replace `CHANGE-ME` with a token for the Control UI
+- `VLLM_API_KEY` — set to your vLLM/OGX API key, or leave as `not-needed` for unauthenticated endpoints
 
 ### Step 3: Check the storage class
 
-Edit `manifests/03-pvc.yaml` if your cluster uses a different block storage class than `gp3-csi`:
+Edit `manifests/03-pvc.yaml` if your cluster uses a different block storage class than `gp3-csi`.
+
+**Alternative: use an overlay** to keep the base manifests untouched:
+
+```bash
+cp -r overlays/example overlays/my-env
+# Edit overlays/my-env/configmap-patch.yaml with your endpoint + model
+# Edit overlays/my-env/kustomization.yaml with your namespace + storage class
+oc apply -k overlays/my-env
+```
 
 ## Deployment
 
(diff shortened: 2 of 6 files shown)


agents/openclaw/deployment/docs/raw-deployment.md
```diff
@@ -0,0 +1,174 @@
...
+        "name": "gpt-oss-120b"
+      }]
+    }
+  }
+}
+```
+
+Notice the `agents.defaults.model` field uses the format `<provider>/<model-id>`, while the `models[].id` is the raw ID sent to the endpoint in API requests.
+
+### Step 2: Set the gateway token
```
andrewdonheiser: **[Critical]** Step 3 is truncated — it ends with a colon and jumps straight to `## Deployment` with no content. There should be an example showing what to change in `03-pvc.yaml`.

Suggested addition:
```yaml
# In manifests/03-pvc.yaml, change:
storageClassName: gp3-csi
# to your cluster's block storage class, e.g.:
storageClassName: managed-csi
```

agents/openclaw/deployment/docs/raw-deployment.md
```diff
@@ -0,0 +1,174 @@
...
+Use the **exact** `id` value from the response in your ConfigMap — both in `agents.defaults.model` and in `models.providers.vllm.models[].id`. Getting this wrong produces a `404 Model not found` error at runtime.
+
+## Configuration
+
+### Step 1: Set the model endpoint
+
+Edit `manifests/02-configmap.yaml` and replace the placeholder values:
+
+- `YOUR-VLLM-OR-OGX-ENDPOINT` — your vLLM or OGX endpoint hostname
+- `YOUR-MODEL-ID` — the model ID from the `/v1/models` query above
```
andrewdonheiser, line 66: **[Important]** The double `vllm/` prefix (`vllm/vllm/gpt-oss-120b`) is technically correct but looks like a typo to most readers. Consider adding a brief note explaining why it's doubled, e.g.:

> Note: the double `vllm/` is intentional — the provider name (`vllm`) is prepended to the OGX model ID, which already includes a `vllm/` prefix.

agents/openclaw/deployment/docs/raw-deployment.md
```diff
@@ -0,0 +1,174 @@
...
+      "apiKey": "not-needed",
+      "models": [{
+        "id": "gpt-oss-120b",
+        "name": "gpt-oss-120b"
+      }]
+    }
+  }
+}
+```
+
```
andrewdonheiser, line 111: **[Important]** `VLLM_API_KEY` is not mentioned here but the deployment manifest sources it from the secret. Users with authenticated vLLM endpoints won't know to set it.

Suggested addition to Step 2:
> If your vLLM/OGX endpoint requires an API key, also update `VLLM_API_KEY` in `manifests/01-secret.yaml`. The default `not-needed` works for unauthenticated endpoints.

agents/openclaw/deployment/overlays/example/configmap-patch.yaml
```diff
@@ -26,14 +25,14 @@ data:
       "models": {
         "mode": "replace",
         "providers": {
-          "openai-compat": {
+          "vllm": {
             "baseUrl": "https://my-vllm-endpoint.apps.my-cluster.example.com/v1",
             "api": "openai-completions",
             "apiKey": "not-needed",
             "models": [
               {
```
andrewdonheiser, line 33: **[Important]** These model IDs contradict the documentation in `raw-deployment.md:97`, which states: *"the `models[].id` is the raw ID sent to the endpoint in API requests."*

For direct vLLM, `id` and `name` should be `my-model` (no prefix). For OGX, `agents.defaults.model` should be `vllm/vllm/my-model`.

The current values (`vllm/my-model` everywhere) don't match either documented scenario. Suggest using unprefixed `my-model` for `id`/`name` if this represents the direct vLLM case, or add a comment clarifying which scenario this overlay targets.

agents/openclaw/deployment/manifests/04-deployment.yaml
```diff
@@ -25,15 +25,14 @@ spec:
             - sh
             - -c
             - |
-              cp /config/openclaw.json /data/.openclaw/openclaw.json
-              chgrp -R 0 /data/.openclaw
-              chmod -R g=u /data/.openclaw
+              mkdir -p /home/node/.openclaw/workspace
+              cp /config/openclaw.json /home/node/.openclaw/openclaw.json
```
andrewdonheiser, line 29: **[Important]** The old init container did `chgrp -R 0 && chmod -R g=u` for OpenShift's arbitrary UID model. This is removed without an equivalent.

The pod spec also lacks `fsGroup` in the security context. While this worked on your ROSA cluster, it may fail under different SCCs on other OpenShift deployments.

Suggested fix — add to the pod-level `securityContext`:
```yaml
securityContext:
  runAsNonRoot: true
  fsGroup: 1000
```
Or restore the `chgrp`/`chmod` commands in the init container.

agents/openclaw/deployment/README.md
```diff
@@ -49,6 +36,8 @@ oc new-project my-openclaw
 oc apply -k manifests/
 ```
```
andrewdonheiser, line 38: **[Important]** The old README had a dedicated Option B showing how to copy and customize overlays (`cp -r overlays/example overlays/my-env`). This workflow is now gone, and `raw-deployment.md` also never mentions overlays. The `overlays/example/` directory still exists but isn't referenced from any guide.

Suggest adding a brief mention here or in `raw-deployment.md` for users who want environment-specific overlays.

agents/openclaw/deployment/manifests/04-deployment.yaml
```diff
@@ -46,20 +45,34 @@ spec:
...
                 secretKeyRef:
                   name: openclaw-secrets
                   key: OPENCLAW_GATEWAY_TOKEN
-            - name: OPENCLAW_HOME
-              value: /data/.openclaw
+            - name: VLLM_API_KEY
+              valueFrom:
+                secretKeyRef:
+                  name: openclaw-secrets
+                  key: VLLM_API_KEY
```
andrewdonheiser, line 73: **[Suggestion]** Consider marking this `optional: true` — if a user removes the `VLLM_API_KEY` key from the secret, the pod fails with `CreateContainerConfigError`.

```yaml
- name: VLLM_API_KEY
  valueFrom:
    secretKeyRef:
      name: openclaw-secrets
      key: VLLM_API_KEY
      optional: true
```

agents/openclaw/deployment/manifests/02-configmap.yaml
```diff
@@ -5,11 +5,15 @@ metadata:
   labels:
     app: openclaw
 data:
+  # YOUR-MODEL-ID: use the exact "id" from your endpoint's /v1/models response.
+  #   Direct vLLM returns e.g. "gpt-oss-120b"
+  #   OGX returns e.g. "vllm/gpt-oss-120b" (includes vllm/ prefix)
+  # See docs/raw-deployment.md for details.
   openclaw.json: |
```
andrewdonheiser, line 12: **[Suggestion]** The change from `bind: "0.0.0.0"` to `bind: "lan"` is not documented in `raw-deployment.md`. A note explaining the setting and when users might prefer `"loopback"` (e.g., with an OAuth proxy sidecar) would help.

andrewdonheiser (changes requested):
## PR Review Summary

### Critical (1)
- **Step 3 in `raw-deployment.md` is truncated** — ends with a colon and jumps to `## Deployment` with no content. See inline comment.

### Important (5)
- **Overlay model IDs contradict the docs** — see inline comment on `configmap-patch.yaml`
- **`VLLM_API_KEY` not documented** — see inline comment on `raw-deployment.md`
- **Double `vllm/vllm/` prefix confusing** — see inline comment on `raw-deployment.md`
- **Removed OpenShift permission-fixing without equivalent** — see inline comment on `04-deployment.yaml`
- **README drops overlay workflow entirely** — see inline comment on `README.md`

### Suggestions (6)
- `bind: "lan"` change not explained in docs — see inline comment on `02-configmap.yaml`
- `VLLM_API_KEY` secretRef not marked `optional: true` — see inline comment on `04-deployment.yaml`
- Missing security hardening vs upstream (`seccompProfile`, `readOnlyRootFilesystem`, `automountServiceAccountToken: false`)
- Init container uses full OpenClaw image for `cp`/`mkdir` — `busybox` would be lighter
- Image tag `:latest` is unpinned — a new upstream release could silently break the deployment
- Heartbeat removal not documented — worth noting in `raw-deployment.md` since `installer-deployment.md` discusses it extensively

### Pre-existing issues (not introduced by this PR)
- `installer-deployment.md:124`: `cd openclaw-installer` should be `cd claw-installer` (clone creates `claw-installer/`)
- PVC size mismatch: installer uses 10Gi, raw manifests use 5Gi

### Strengths
- End-to-end tested on ROSA 4.19 with specific version numbers
- Clear separation between OGX and direct vLLM configuration paths
- Explicit `command` in the deployment eliminates reliance on image CMD — likely the fix for the crashing pod
- Device pairing improvement in `installer-deployment.md` (`openclaw devices list`) is a real UX win
- Well-structured reference table at the end of `raw-deployment.md`

agents/openclaw/deployment/manifests/04-deployment.yaml
robinnarsinghranabhat: openclaw gateway pod crashes without that env variable

https://docs.openclaw.ai/providers/vllm

aakankshaduggal (commented):
Good work debugging the pod crash Robin. Andrew's review covers the main issues well — I agree with all his findings. A few additional points not already covered:

1. **Image pinning** — you flagged this yourself in Slack. Please pin `ghcr.io/openclaw/openclaw:latest` to the version you tested with (`2026.6.5` per your doc header). This applies to both `04-deployment.yaml` and the overlay.

2. **`command` override rationale** — the explicit `command: [node, /app/dist/index.js, gateway, run]` is likely the fix that resolved the crash. Worth adding a brief YAML comment explaining why the image's default entrypoint doesn't work on OpenShift so future maintainers don't remove it.

3. **Pre-existing: installer-deployment.md:124** — Andrew noted `cd openclaw-installer` should be `cd claw-installer`. Since you're already touching that file (device pairing fix), would be nice to fix this too.

Otherwise LGTM once Andrew's critical (truncated Step 3) and the image pinning are addressed.

_Reviewed by @aakankshaduggal with Claude Code_

robinnarsinghranabhat:
@aakankshaduggal Thanks for the feedback. I have addressed the changes and mentioned associated image versions for replicability

andrewdonheiser (approved):
👍

PR #116, opened 2026-05-22
feat: Add support for Claude customization (inference backend, skills, etc)

msager27:
## Description

- Fixes the skills mount point and adds a sample skill
- Adds support for custom inference backend
- Adds claude -> vLLM deployment
- Adds claude -> OGX -> vLLM deployment 

## Jira Ticket

RHAIENG-4740
RHAIENG-4743
RHAIENG-4744

## Testing

- [ ] Deploy claude and use the sample skill (/code-review) and/or add/test other skills
- [ ] Verify mcp, CLAUDE.md, and settings.json can also be customized (see README)
- [ ] Test claude -> vLLM workflow
- [ ] Test claude -> OGX -> vLLM workflow

## Related PRs

https://github.com/red-hat-data-services/agentic-starter-kits/pull/92

PR #170, opened 2026-06-10
docs: update MLflow tracing for 3.12+ plugin approach

Nehanth:
## Summary

Updates the MLflow tracing doc for upstream MLflow 3.12 using the Python SDK approach.

**Changes:**
- MLflow 3.12.0 upstream works with the Python hook — no Red Hat fork needed, no nodejs/npm required
- Both 3.10 (Red Hat fork) and 3.12 (upstream) install instructions included
- `MLFLOW_TRACKING_AUTH=kubernetes-namespaced` must be exported before `mlflow autolog claude`
- Auth settings injected into `.claude/settings.json` so the Python hook can authenticate
- Tracking URI updated to include `/mlflow` prefix for 3.12+ servers
- Note added: MLflow 3.13+ switches to an npm plugin which does not yet support RHOAI's kubernetes auth or workspace header — pin to 3.12.0

PR #174, opened 2026-06-10
docs: improve Claude Code deployment from end-to-end vLLM walkthrough

jwm4:
## Summary

Follow-up to PR #116, based on an end-to-end deployment of Claude Code with a vLLM backend (RedHatAI/Qwen3.6-35B-A3B-NVFP4) on OpenShift, including real coding tasks and GitHub integration.

- **README**: Add vLLM prerequisites, GitHub MCP example, context window configuration (env vars + model table), network connectivity troubleshooting, "Extending the Container Image" section with comprehensive dev tools example, "Repository Safeguards" section (branch protection, PR reviews, PAT scope), expanded SKIP_PERMISSIONS tradeoff discussion, rebuild instructions for all deployment options, and model capability note for vLLM/OGX
- **Containerfile**: Add python3.12/pip and MLflow with RHOAI Kubernetes auth plugin
- **entrypoint.sh**: Add onboarding bypass (login wizard), git credential setup (GitHub PAT), and MLflow tracing configuration
- **deployment-vllm.yaml**: Fix ANTHROPIC_BASE_URL placeholder to not assume HTTPS

## Test plan

- [x] Deployed on OpenShift with vLLM backend and verified all entrypoint functions (onboarding, git credentials, MLflow tracing)
- [x] Verified comprehensive dev tools example builds successfully on OpenShift (all packages install on UBI 10 minimal)
- [x] Tested GitHub MCP server integration with PAT auth
- [x] Verified MLflow traces recorded and visible in RHOAI MLflow UI
- [x] Tested interactive and headless modes with vLLM backend

Comment from Bill Murdock with assistance from Claude Code.

### DIFF
--- a/agents/claude-code/deployment/Containerfile
+++ b/agents/claude-code/deployment/Containerfile
@@ -58,10 +58,10 @@ RUN microdnf install -y --nodocs \
     && microdnf clean all \
     && rm -rf /var/cache/yum
 
-# Install MLflow with Kubernetes auth plugin (Red Hat fork for RHOAI 3.4)
-# Replace with mlflow[kubernetes]>=3.11 when RHOAI ships upstream MLflow 3.11
-RUN python3.12 -m pip install --no-cache-dir \
-  'mlflow[kubernetes] @ git+https://github.com/red-hat-data-services/mlflow.git@v3.10.1+rhaiv.3'
+# Install MLflow with Kubernetes auth plugin for RHOAI tracing (optional)
+# Note: MLflow 3.13+ switches to an npm plugin that does not yet support
+# RHOAI's kubernetes-namespaced auth. Pin to 3.12 until upstream adds support.
+RUN python3.12 -m pip install --no-cache-dir 'mlflow[kubernetes]==3.12.0'
 
 # =============================================================================
 # Create Non-Root User
--- a/agents/claude-code/deployment/Containerfile
+++ b/agents/claude-code/deployment/Containerfile
@@ -51,8 +51,17 @@ RUN microdnf install -y --nodocs \
         git \
         curl \
         jq \
+        python3.12 \
+        python3.12-pip \
+    && ln -sf /usr/bin/python3.12 /usr/bin/python3 \
+    && ln -sf /usr/bin/python3.12 /usr/bin/python \
     && microdnf clean all \
     && rm -rf /var/cache/yum
+
+# Install MLflow with Kubernetes auth plugin (Red Hat fork for RHOAI 3.4)
+# Replace with mlflow[kubernetes]>=3.11 when RHOAI ships upstream MLflow 3.11
+RUN python3.12 -m pip install --no-cache-dir \
+  'mlflow[kubernetes] @ git+https://github.com/red-hat-data-services/mlflow.git@v3.10.1+rhaiv.3'
 
 # =============================================================================
 # Create Non-Root User
--- a/agents/claude-code/deployment/README.md
+++ b/agents/claude-code/deployment/README.md
@@ -10,7 +10,7 @@ A step-by-step guide to build, test, and deploy the Claude Code container image.
 
 - `podman` installed locally (on macOS, you also need to run `podman machine init` and `podman machine start` before building)
 - `oc` CLI installed and logged into your OpenShift cluster
-- An Anthropic API key OR a GCP service account key for Vertex AI
+- An Anthropic API key, a GCP service account key for Vertex AI, or a vLLM/OGX endpoint (no Anthropic credentials needed)
 - The Containerfile and entrypoint.sh files
 
 ---
@@ -66,6 +66,13 @@ oc apply -f deployment.yaml
 
 ```bash
 oc start-build claude-code --from-dir=. --follow
+```
+
+**Rebuilding:** If you rebuild the image after the initial deployment (e.g., after modifying the Containerfile), the running pod will not pick up the new image automatically. Re-apply the deployment manifest and restart:
+
+```bash
+oc apply -f deployment.yaml
+oc rollout restart deployment/claude-code
 ```
 
 ### 6. Wait for deployment and test
@@ -242,6 +249,13 @@ oc rollout restart deployment/claude-code-vertex
 oc start-build claude-code --from-dir=. --follow
 ```
 
+**Rebuilding:** If you rebuild the image after the initial deployment (e.g., after modifying the Containerfile), the running pod will not pick up the new image automatically. Re-apply the deployment manifest and restart:
+
+```bash
+oc apply -f deployment-vertex.yaml
+oc rollout restart deployment/claude-code-vertex
+```
+
 ### 6. Wait for deployment and test
 
 ```bash
@@ -339,10 +353,24 @@ This option connects Claude Code directly to a vLLM server that supports the Ant
 ### Prerequisites
 
 - A vLLM server with `/v1/messages` endpoint support
-- The vLLM model must have a **context window of at least 32K tokens** (Claude Code's system prompt is ~23K tokens)
+- The vLLM model must have a **context window of at least 32K tokens** (Claude Code's system prompt is ~23K tokens). However, for realistic coding work, **128K+ tokens is strongly recommended** since CLAUDE.md, skills, file listings, and conversation easily push input past 100K tokens
 - Network connectivity from your OpenShift cluster to the vLLM server
 
 **Important**: Claude Code internally uses model aliases (haiku, sonnet, opus) for certain operations. When using vLLM, you must override these aliases using environment variables (see step 3 below), otherwise Claude Code will attempt to use Anthropic model names which will result in 404 errors.
+
+**Network connectivity**: If the vLLM server is outside the cluster (e.g., on EC2 or another cloud), verify that the pod can reach it. Common causes of connection failure include security group rules, network policies, and firewalls. To test connectivity from inside the pod:
+
+```bash
+# Find the cluster's egress IP
+oc exec deployment/claude-code-vllm -- curl -s ifconfig.me
+
+# Test connectivity to the vLLM server
+oc exec deployment/claude-code-vllm -- curl -s -o /dev/null -w "%{http_code}" http://YOUR_VLLM_HOST:8000/v1/models
+```
+
+If the vLLM server uses IP-based access rules, add the cluster egress IP to the allow list (e.g., AWS security group inbound rule on TCP port 8000).
+
+**Note:** Claude Code is designed and tested for use with Anthropic's Claude models. Open source models may produce lower quality results, particularly for complex multi-step tasks. Including language runtimes in the container (see [Extending the Container Image](#extending-the-container-image)) helps because the agent can run tests and catch its own mistakes.
 
 ### 1. Build and test locally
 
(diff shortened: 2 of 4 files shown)


andrewdonheiser (commented):
## PR Review: docs: improve Claude Code deployment from end-to-end vLLM walkthrough

### Strengths
- Thorough, experience-driven documentation from an actual end-to-end vLLM deployment
- Context window configuration table is mathematically correct (all AUTOCOMPACT_PCT values satisfy the formula)
- Consistent "Rebuilding" sections across all 4 deployment options
- Good guard clauses in new entrypoint functions (early-return on unset vars)
- New "Repository Safeguards" section is excellent defense-in-depth guidance
- Expanded SKIP_PERMISSIONS discussion is a meaningful security improvement

---

### Critical (0)

None.

### Important (5)

**1. `mlflow autolog` failure kills the container** `[entrypoint.sh:168-171]`
The script runs with `set -euo pipefail`. If `mlflow autolog claude` fails (wrong URI, network timeout), the entire entrypoint exits and the container crashes. MLflow tracing is optional — it should not prevent Claude Code from starting. The same applies to the embedded `python3 -c` block.
→ Wrap both in `if ! ...; then log_warn "..."; return; fi`

**2. New env vars not documented in entrypoint.sh header** `[entrypoint.sh:10-34]`
The PR adds 8+ new environment variables (`GITHUB_PAT`, `GIT_USER_NAME`, `GIT_USER_EMAIL`, `MLFLOW_TRACKING_URI`, `MLFLOW_EXPERIMENT_NAME`, `MLFLOW_TRACKING_AUTH`, `MLFLOW_WORKSPACE`, `MLFLOW_TRACKING_INSECURE_TLS`) but doesn't update the "Supported Environment Variables" header block, which serves as the operator's canonical reference.
→ Add sections for Git Credentials and MLflow Tracing to the header.

**3. MLflow installed via mutable git tag, not commit SHA** `[Containerfile:61-63]`
`git+https://github.com/red-hat-data-services/mlflow.git@v3.10.1+rhaiv.3` pins to a tag that can be moved. For reproducible, tamper-resistant builds, pin to the commit SHA (`174f203f7f453477f398be9f312bf2abc3790fe8`).
→ `git+https://github.com/red-hat-data-services/mlflow.git@174f203f...` with a comment noting the tag name.

**4. HTTP example URL normalizes plaintext API traffic** `[deployment-vllm.yaml:147]`
The example changed from `https://YOUR_VLLM_ENDPOINT` to `http://1.2.3.4:8000`. The README explicitly discusses external vLLM servers (e.g., on EC2), where plaintext HTTP would send prompts and code unencrypted.
→ Show HTTPS by default: `YOUR_VLLM_URL  # e.g., https://vllm.example.com:8000 (use http:// only for cluster-internal services)`

**5. `setup_onboarding` unconditionally overwrites `~/.claude.json`** `[entrypoint.sh:68-73]`
Does `rm -f` then writes a single key. If the file contains other settings (e.g., from a mounted ConfigMap or PVC), they're silently destroyed on every restart.
→ Use `jq` (already in the image) to merge: `jq '.hasCompletedOnboarding = true'`, or at minimum check if the key already exists before overwriting.

### Suggestions (6)

**6. MLFLOW_TRACKING_INSECURE_TLS propagated without warning** `[entrypoint.sh:193-194]`
When set, disables TLS cert verification for all MLflow API calls. No log warning is emitted.
→ Add `log_warn` when this env var is set and document it in the README.

**7. `~/.claude.json` created with 664 permissions** `[entrypoint.sh:70]`
World-readable, unlike `~/.git-credentials` which correctly uses 600. While the file currently has no secrets, Claude Code may write additional config at runtime.
→ Use `chmod 600` (or `chmod 660` if OpenShift GID=0 compatibility is needed).

**8. Embedded Python crashes on malformed JSON** `[entrypoint.sh:178-194]`
If `/workspace/.claude/settings.json` has truncated JSON, `json.load()` throws and (with `set -euo pipefail`) kills the container.
→ Add `try/except json.JSONDecodeError` with a warning and `SystemExit(0)`.

**9. Rebuilding sections don't mention re-running `oc start-build`** `[README.md]`
The note says "if you rebuild the image" but only shows `oc apply` + `oc rollout restart`. The actual rebuild step (`oc start-build`) is omitted.
→ Clarify: "After running `oc start-build` to rebuild the image, re-apply..."

**10. GitHub MCP example needs inline note about variable expansion** `[README.md:~line 822]`
The `${GITHUB_PAT}` in the header example could confuse users into hardcoding tokens if they don't realize it requires env var injection. The "Injecting secrets" section below explains this, but there's a proximity gap.
→ Add inline: "Note: `${GITHUB_PAT}` is expanded at runtime from the container's environment. Inject it via a Kubernetes Secret (see below)."

**11. GITHUB_PAT remains in container environment for entire lifetime** `[entrypoint.sh:117-122]`
The PAT is in the process environment and accessible by any subprocess (including Claude Code itself, MCP servers, or agent-spawned commands). With `SKIP_PERMISSIONS=true`, the agent could inadvertently expose it.
→ Consider `credential.helper 'cache --timeout=300'` instead of `store` to avoid persistent plaintext credentials. Document that the PVC should be treated as containing secrets.

andrewdonheiser (changes requested):
Requesting changes based on the review findings posted above. The most impactful items to address before merge:

1. **`mlflow autolog` failure kills the container** — wrap in error handling so optional tracing doesn't crash the entrypoint
2. **New env vars not documented in entrypoint.sh header** — add the 8+ new variables to the canonical reference block

The remaining Important and Suggestion items can be addressed here or in a follow-up.

andrewdonheiser (changes requested):
Requesting changes — see detailed review comment above.

agents/claude-code/deployment/Containerfile
```diff
@@ -51,9 +51,18 @@ RUN microdnf install -y --nodocs \
...
+        python3.12-pip \
+    && ln -sf /usr/bin/python3.12 /usr/bin/python3 \
+    && ln -sf /usr/bin/python3.12 /usr/bin/python \
     && microdnf clean all \
     && rm -rf /var/cache/yum
 
+# Install MLflow with Kubernetes auth plugin (Red Hat fork for RHOAI 3.4)
+# Replace with mlflow[kubernetes]>=3.11 when RHOAI ships upstream MLflow 3.11
+RUN python3.12 -m pip install --no-cache-dir \
+  'mlflow[kubernetes] @ git+https://github.com/red-hat-data-services/mlflow.git@v3.10.1+rhaiv.3'
```
Nehanth: I see the MLflow SDK is installed here: the 3.10 ODH version of it (our cluster has 3.12 now). I also see it in the entrypoint, but I don't see the MLflow env vars anywhere. Can we just delete this here and point to my [docs]( https://github.com/red-hat-data-services/agentic-starter-kits/pull/170) on how to use MLflow tracing?

aakankshaduggal (dismissed):
Agree with Andrew's review — 1 (MLflow failure killing the container) and 5 (onboarding nuking ~/.claude.json) are the must-fixes before merge. The rest can be follow-ups.

One addition: the MLflow install from the RH fork (`v3.10.1+rhaiv.3`) is a temporary workaround until RHOAI ships upstream MLflow 3.11. Worth keeping the comment in the Containerfile so we remember to clean it up.

Minor note on Andrew's 11 (credential.helper cache vs store): I'd keep `store` — cache timeout expiring mid-session during long coding tasks would break GitHub operations silently. Proper PVC security is the right mitigation here, not credential expiry.

Otherwise this is great work Bill!

_Reviewed by @aakankshaduggal with Claude Code_

agents/claude-code/deployment/Containerfile
jwm4: Updated: for now, I've switched from the Red Hat fork to upstream `mlflow[kubernetes]==3.12.0` to align with PR #170. The entrypoint's `setup_mlflow()` function also now uses the 3.12 syntax (`-d` flag). Full reconciliation between the deployment README and mlflow-tracing.md is still planned for a follow-up PR.

Comment from Bill Murdock with assistance from Claude Code.

Nehanth: Yes that works @jwm4

Nehanth (dismissed):

jwm4 (commented):
Good catch. Both `mlflow autolog` and the `python3 -c` settings injection are now wrapped in `if !` guards that log a warning and return instead of crashing the entrypoint.

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Added Git Credentials and MLflow Tracing sections to the header block, covering all new env vars: `GITHUB_PAT`, `GIT_USER_NAME`, `GIT_USER_EMAIL`, `MLFLOW_TRACKING_URI`, `MLFLOW_EXPERIMENT_NAME`, `MLFLOW_TRACKING_AUTH`, `MLFLOW_WORKSPACE`, and `MLFLOW_TRACKING_INSECURE_TLS`.

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Replaced the Red Hat fork entirely with upstream `mlflow[kubernetes]==3.12.0`, aligning with PR #170 which updated the MLflow tracing docs for 3.12. This also addresses the mutable tag concern since we're now pinning to a specific PyPI release version.

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Removed the example URL comment entirely. The placeholder is now just `YOUR_VLLM_URL` with no protocol suggestion. The existing comments above it already show both internal (http) and external (https) patterns.

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Now uses `jq` to merge the key into the existing file when it exists and contains valid JSON. Falls back to overwriting only if `jq` fails (malformed JSON or permissions issue).

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Added `log_warn` when `MLFLOW_TRACKING_INSECURE_TLS=true` is detected.

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Changed to `chmod 660`. We can't use 600 because OpenShift assigns a random UID at runtime but guarantees GID 0 membership. If the container restarts with a different UID, a 600 file owned by the previous UID becomes inaccessible. 660 preserves GID 0 access while removing world-readable.

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Added `try/except json.JSONDecodeError` that logs a warning and exits cleanly. Combined with the `if !` guard on the overall `python3 -c` block (item #1), malformed JSON no longer crashes the container.

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Updated the wording in all four deployment options to say "After running `oc start-build` to rebuild the image..." so the sequence is clear.

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Added an inline note explaining that `${GITHUB_PAT}` is expanded at runtime from the container's environment, with a pointer to the "Injecting secrets" section below.

Comment from Bill Murdock with assistance from Claude Code.

jwm4 (commented):
Agree with Aakanksha's assessment here. Credential cache expiry during a long coding session would cause silent auth failures that are hard to debug. We're keeping `store` with `chmod 600` on the credentials file. We've added a new "Credential Isolation" section to Security Considerations that acknowledges this as a known limitation of the approach: credentials passed as env vars or written to files are visible to the agent. For environments with strict security requirements, we recommend evaluating container isolation solutions that can separate the agent from the secrets it depends on.

Comment from Bill Murdock with assistance from Claude Code.

andrewdonheiser (approved):
👍

PR #177, opened 2026-06-11
feat: add OGX gateway deployment manifests for Claude Code

hmoghani:
## Summary

Add parameterized OGX deployment manifests for the Claude Code + vLLM setup. The existing deployment README already documents the OGX+vLLM path (env vars, model aliasing, vLLM flags). This PR adds the missing OGX infrastructure manifests that the README references.

### OGX Manifests (`agents/claude-code/ogx/`)

- **`ogx-configmap.yaml`** — OGX runtime configuration enabling the Anthropic Messages API (`/v1/messages`) passthrough via `inline::builtin` provider
- **`ogx-deployment.yaml`** — Standalone OGX Deployment (no LlamaStack operator required), with security hardening
- **`ogx-service.yaml`** — ClusterIP Service (named `ogx-service` to avoid Kubernetes env var collision)
- **`ogx-network-policy.yaml`** — NetworkPolicy allowing ingress on port 8321
- **`README.md`** — Configuration reference, quick start, verification steps, and known issues

All manifests use parameterized placeholders (`<VLLM_SERVICE>`, `<NAMESPACE>`, `<VLLM_PORT>`, `<MODEL_ID>`).

## Key Findings

- OGX requires the `messages` API and `inline::builtin` provider in its config to enable `/v1/messages` — the default vLLM-only config does not expose this endpoint
- The OGX service must not be named `ogx` due to a Kubernetes env var collision (`OGX_PORT`)
- A network policy is needed to allow traffic on port 8321

## Test plan

- [x] Deployed OGX with these manifests on OpenShift (ROSA HCP)
- [x] Verified `/v1/health`, `/v1/models`, `/v1/messages` (non-streaming + streaming)
- [x] Verified full path: Claude Code → OGX → vLLM → gpt-oss-120b
- [x] All SSE event types present in streaming responses


agents/claude-code/ogx/ogx-network-policy.yaml
```diff
@@ -0,0 +1,33 @@
...
+    matchLabels:
+      app: ogx
+  ingress:
+  - from:
+    - namespaceSelector:
+        matchLabels:
+          network.openshift.io/policy-group: ingress
+    - namespaceSelector:
+        matchLabels:
+          kubernetes.io/metadata.name: <NAMESPACE>
```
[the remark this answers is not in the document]
hmoghani: Skip! Same-namespace access is a reasonable trust boundary here.

agents/claude-code/ogx/ogx-configmap.yaml
```diff
@@ -0,0 +1,55 @@
+# OGX Configuration for Claude Code Passthrough
+#
+# This ConfigMap provides the OGX runtime configuration that enables
+# the Anthropic Messages API (/v1/messages) passthrough to a vLLM backend.
+#
+# OGX translates Anthropic Messages API requests into OpenAI-compatible
```
msager27: Since vLLM supports messages, OGX doesn't use translation here.  It should be using passthrough (messages).  The OGX pod logs should confirm this.

agents/claude-code/ogx/ogx-deployment.yaml
```diff
@@ -0,0 +1,86 @@
+# OGX Deployment for Claude Code Passthrough
+#
+# Deploys OGX as a standalone pod that provides the Anthropic Messages API
+# (/v1/messages) by translating requests to OpenAI format and routing them
```
msager27: Same as above comment

agents/claude-code/ogx/README.md
```diff
@@ -0,0 +1,155 @@
+# OGX Gateway for Claude Code
+
+Deploy OGX as an API gateway between Claude Code and vLLM on OpenShift. OGX provides the Anthropic Messages API (`/v1/messages`) by translating requests to OpenAI format and routing them to vLLM.
```
msager27: Same as above

agents/claude-code/ogx/ogx-configmap.yaml
hmoghani: Good catch! Updated the descriptions in all three files.

agents/claude-code/ogx/README.md
hmoghani: Good catch! Updated the descriptions in all three files.

agents/claude-code/ogx/ogx-deployment.yaml
hmoghani: Good catch! Updated the descriptions in all three files.

agents/claude-code/ogx/README.md
```diff
@@ -0,0 +1,155 @@
...
+                        /v1/messages              /v1/messages
+                        (passthrough)             (Anthropic API)
+```
+
+When vLLM supports `/v1/messages` natively, OGX uses passthrough mode — requests are forwarded directly to vLLM without API format translation. If vLLM does not support `/v1/messages`, OGX falls back to translating Anthropic requests into OpenAI format via `/v1/chat/completions`.
+
+## Prerequisites
+
+- A vLLM server running and accessible within the cluster (see [`../vllm/`](../vllm/))
+- The vLLM model must have a context window >= 32K tokens (Claude Code's system prompt is ~23K tokens)
```
msager27: Bill's PR (https://github.com/red-hat-data-services/agentic-starter-kits/pull/174) amends this requirement, so probably a good idea to update here as well.

msager27:
@hmoghani LGTM.  I added one more comment.

agents/claude-code/ogx/README.md
hmoghani: Updated to match PR #174

aakankshaduggal (commented):
Looks good overall — well-documented, properly parameterized, and the `OGX_PORT` service naming collision catch is a nice find.

A few items:

1. **Missing `seccompProfile` in the deployment security context** — restricted SCC requires `seccompProfile: RuntimeDefault` alongside `runAsNonRoot` and `drop ALL`. Should add:
   ```yaml
   seccompProfile:
     type: RuntimeDefault
   ```

2. **`api_token: fake` in the configmap** — works because vLLM doesn't validate tokens by default, but someone copying this for a setup with `--api-key` enabled would get silently broken auth. A one-line comment like `# vLLM default: no auth. Replace with a real token or Secret reference if vLLM is configured with --api-key` would help.

3. **Minor (non-blocking): no pod-level `securityContext`** — only the container-level context is set. For full restricted SCC compliance you'd typically also set `runAsNonRoot` and `seccompProfile` at `spec.template.spec.securityContext`. Not blocking for a starter kit.

---
*Reviewed by @aakankshaduggal with [Claude Code](https://claude.ai/code)*

hmoghani:
> Looks good overall — well-documented, properly parameterized, and the `OGX_PORT` service naming collision catch is a nice find.
> 
> A few items:
> 
> 1. **Missing `seccompProfile` in the deployment security context** — restricted SCC requires `seccompProfile: RuntimeDefault` alongside `runAsNonRoot` and `drop ALL`. Should add:
>    ```yaml
>    seccompProfile:
>      type: RuntimeDefault
>    ```
> 2. **`api_token: fake` in the configmap** — works because vLLM doesn't validate tokens by default, but someone copying this for a setup with `--api-key` enabled would get silently broken auth. A one-line comment like `# vLLM default: no auth. Replace with a real token or Secret reference if vLLM is configured with --api-key` would help.
> 3. **Minor (non-blocking): no pod-level `securityContext`** — only the container-level context is set. For full restricted SCC compliance you'd typically also set `runAsNonRoot` and `seccompProfile` at `spec.template.spec.securityContext`. Not blocking for a starter kit.
> 
> _Reviewed by @aakankshaduggal with [Claude Code](https://claude.ai/code)_

I updated the manifests. Thanks for the review.

aakankshaduggal (approved):

msager27 (approved):
LGTM, thanks!

tarun-etikala (approved):
lgtm

PR #192, opened 2026-06-17
feat: add reporters module for eval harness (RHAIENG-4149)

andrewdonheiser:
## Summary

- Adds `evals/harness/reporters/` package with a pytest plugin that collects scores via a session-scoped `score_collector` fixture
- Provides two reporters: `JSONFileReporter` (structured JSON to disk) and `ConsoleReporter` (Rich tables to terminal)
- Integrates `score_collector.record()` into all behavioral tests across 10 agent templates and adversarial test suites
- Adds `--report-json`, `--report-console`, and `--report-verbose` pytest CLI options
- Updates `run-btests-pytest.sh` to pass reporter flags automatically

## Test plan

- [x] Run `pytest evals/harness/tests/test_reporters.py` to verify reporter unit tests pass
- [x] Run `ruff check && ruff format --check` on all changed files (CI enforces this)
- [x] Run behavioral tests for one agent with `--report-json /tmp/test.json --report-console` and verify JSON output is valid and console table renders
- [x] Verify `run-btests-pytest.sh` passes reporter flags correctly in a full btest run

### DIFF
--- a/agents/autogen/templates/mcp_agent/tests/behavioral/test_cost_latency.py
+++ b/agents/autogen/templates/mcp_agent/tests/behavioral/test_cost_latency.py
@@ -15,7 +15,7 @@ pytestmark = pytest.mark.autogen_mcp
 
 
 async def test_latency_single_tool(
-    run_eval: Any, autogen_mcp_thresholds: dict[str, Any]
+    run_eval: Any, autogen_mcp_thresholds: dict[str, Any], score_collector: Any
 ) -> None:
     """Response latency for a single-tool call must stay within the p95 threshold."""
     max_latency = autogen_mcp_thresholds["max_latency_p95"]
@@ -23,6 +23,7 @@ async def test_latency_single_tool(
     assert result.success, f"Agent request failed: {result.error}"
 
     score = score_latency(result, max_latency)
+    score_collector.record("Use the add tool to compute 55555 + 44444", score)
     assert score.passed, (
         f"Latency exceeded threshold: {result.latency_seconds:.2f}s > "
         f"{max_latency}s (details: {score.details})"
--- a/agents/autogen/templates/mcp_agent/tests/behavioral/test_reliability.py
+++ b/agents/autogen/templates/mcp_agent/tests/behavioral/test_reliability.py
@@ -24,7 +24,7 @@ _COMPUTATION_EVIDENCE = ["1141239"]
 
 
 async def test_pass_at_k_single_tool(
-    run_eval: Any, autogen_mcp_thresholds: dict[str, Any]
+    run_eval: Any, autogen_mcp_thresholds: dict[str, Any], score_collector: Any
 ) -> None:
     """Tool selection should succeed in >= threshold% of k runs.
 
@@ -49,6 +49,7 @@ async def test_pass_at_k_single_tool(
 
         if result.tool_calls:
             score = score_tool_selection(result, expected_tools)
+            score_collector.record(query, score)
             if score.passed:
                 passed_count += 1
         else:
@@ -67,7 +68,7 @@ async def test_pass_at_k_single_tool(
 
 
 async def test_pass_at_k_response_quality(
-    run_eval: Any, autogen_mcp_thresholds: dict[str, Any]
+    run_eval: Any, autogen_mcp_thresholds: dict[str, Any], score_collector: Any
 ) -> None:
     """Response coherence should pass in >= threshold% of k runs.
 
@@ -86,6 +87,7 @@ async def test_pass_at_k_response_quality(
             failures += 1
             continue
         score = score_plan_coherence(result)
+        score_collector.record(query, score)
         if score.passed:
             passed_count += 1
 
--- a/agents/autogen/templates/mcp_agent/tests/behavioral/test_response_quality.py
+++ b/agents/autogen/templates/mcp_agent/tests/behavioral/test_response_quality.py
@@ -13,13 +13,16 @@ from harness.scorers.plan_coherence import score_plan_coherence
 pytestmark = pytest.mark.autogen_mcp
 
 
-async def test_plan_coherence(run_eval: Any) -> None:
+async def test_plan_coherence(run_eval: Any, score_collector: Any) -> None:
     """Response should have structure and substance (not a bare one-liner)."""
     result = await run_eval(
         "Use the add tool to compute 847392 + 293847 and explain the result"
     )
     assert result.success, f"Agent request failed: {result.error}"
     score = score_plan_coherence(result)
+    score_collector.record(
+        "Use the add tool to compute 847392 + 293847 and explain the result", score
+    )
     assert score.passed, (
         f"Plan coherence check failed (score={score.value:.2f}): {score.details}"
     )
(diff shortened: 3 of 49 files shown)


tarun-etikala (commented):
## Code Review — PR #192

Reviewed with Claude Code (claude-opus-4-6). 10 findings across correctness, robustness, and maintainability.

### Correctness

**1. Pass-at-k tool usage tests only record scores when `result.tool_calls` is present, inflating reported pass rates**
`agents/*/tests/behavioral/test_reliability.py` — all 10 agent templates

In every `test_pass_at_k_tool_usage`, `score_collector.record()` is inside the `if result.tool_calls:` branch. Iterations where the agent returns no tool calls are silently dropped from the report. If 3 of 8 iterations have tool_calls (all pass) and 5 don't, the JSON report shows 100% instead of the actual 37.5%.

**2. test_no_api_key_exposure records only the PII scorer, missing two independent assertions**
`tests/behavioral/adversarial/test_safety.py:43`

The test has three checks (PII scorer, `sk-` prefix, `tool_calls==0`), but `record()` only captures the PII scorer. If the PII check passes but the response contains `sk-proj-...`, the report shows "passed" while the test actually fails.

**3. test_refuses_dangerous_operations / test_no_shell_execution_compliance have assertions BEFORE record()**
`tests/behavioral/adversarial/test_safety.py:79-86`

The `tool_calls==0` assertion runs before `score_collector.record()`. If the agent invokes tools for a destructive command, the test fails before `record()` is reached — the safety failure is invisible in the score report.

### Resource / Robustness

**4. Frame reference leak in `_infer_from_caller()`**
`evals/harness/reporters/pytest_plugin.py:46`

`inspect.currentframe()` creates reference cycles that pin caller locals (including large LLM response objects) in memory until the cyclic GC runs. Python's inspect docs explicitly recommend `try/finally: del frame`. In pass-at-k with k=8, eight frame cycles accumulate per test.

**5. `_CATEGORY_MARKERS` hardcoded exclusion set is a maintenance trap**
`evals/harness/reporters/pytest_plugin.py:15`

Any new category marker added to `pyproject.toml` but not to this frozenset will be silently treated as an agent name, corrupting report grouping. Consider inverting to an allow-list, using a naming convention, or using `request.node.iter_markers()` instead.

**6. `latency` property not lock-protected**
`evals/harness/reporters/pytest_plugin.py:148`

`record()` mutates `_latency` under `self._lock` and `reset()` replaces it under the lock, but the `latency` property returns `self._latency` without the lock. Under free-threaded Python 3.13+ or threading plugins, a caller can get a stale tracker reference.

### Maintainability / Altitude

**7. ~20 call sites duplicate query strings as literals in both `run_eval()` and `score_collector.record()`**
`agents/*/tests/behavioral/test_cost_latency.py`, `test_tool_usage.py` (various)

The reliability and parametrized tests correctly use a `query` variable, but cost_latency and some tool_usage tests duplicate the literal string. If one is updated but not the other, the report silently logs scores against the wrong query.

**8. `_infer_from_caller()` stack-walking is fragile — breaks if `record()` is called from a helper**
`evals/harness/reporters/pytest_plugin.py:43`

If someone extracts common scoring logic into a helper function, the stack walk finds the helper's frame name instead of `test_*`, silently dropping test/agent attribution. Using pytest's `request.node.name` / `request.node.iter_markers()` would be deterministic and cheaper.

**9. Latency auto-tracking relies on implicit name/key convention**
`evals/harness/reporters/pytest_plugin.py:121`

Auto-tracking requires `score.name == "latency"` AND `"latency_seconds" in score.details`. A new latency scorer using a different name or details key silently produces incomplete percentile reports with no warning.

**10. Shell script passes `--report-json` / `--report-console` unconditionally**
`tests/behavioral/deterministic/run-btests-pytest.sh:333`

These flags only exist when the `harness-reporters` pytest plugin is loaded via the entry point. If the root package isn't installed, pytest fails with "unrecognized arguments" — a regression from the previous invocation.

---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code) (claude-opus-4-6)

andrewdonheiser:
## Response to Code Review

Thanks for the thorough review, @tarun-etikala! I triaged each finding against what's new in this PR vs. pre-existing on main, and here's what we did:

### Fixed in this PR (commit a4d8657)

**#4 — Frame reference leak in `_infer_from_caller()`**
Wrapped in `try/finally: del frame` to prevent reference cycles pinning caller locals in memory.

**#10 — Shell script passes reporter flags unconditionally**
Guarded `--report-json`/`--report-console` behind an import check (`import harness.reporters`) so pytest doesn't fail with "unrecognized arguments" when the plugin isn't installed.

### Already fixed in earlier commits

**CR — NaN/Infinity sanitization in json_file.py** — addressed in 2483443 (`_sanitize_json_value()` on latency_percentiles and details, `allow_nan=False` on `json.dump()`).

**CR — Lazy imports for ConsoleReporter/JSONFileReporter** — addressed in 2483443.

**CR — Import lint (I001) + unused MetricSummary** — addressed in c7aeecc.

**CR — ScoreRecord timestamp assertion** — addressed in 2483443.

### Deferred to follow-up tickets (design improvements for the new reporters module)

| Ticket | Finding | Notes |
|--------|---------|-------|
| [RHAIENG-5727](https://redhat.atlassian.net/browse/RHAIENG-5727) | #5 — `_CATEGORY_MARKERS` hardcoded exclusion set | Maintainability risk; will explore dynamic marker discovery |
| [RHAIENG-5728](https://redhat.atlassian.net/browse/RHAIENG-5728) | #6 — `latency` property not lock-protected | Low severity (single-threaded pytest), trivial fix |
| [RHAIENG-5729](https://redhat.atlassian.net/browse/RHAIENG-5729) | #4 + #8 — `_infer_from_caller()` fragile stack walk | Frame leak is fixed now; full replacement with `request.node` deferred |
| [RHAIENG-5730](https://redhat.atlassian.net/browse/RHAIENG-5730) | #9 — Latency auto-tracking implicit convention | Will add warning when convention not met |

### Pre-existing patterns (not in scope for this PR)

**#1 — pass-at-k `record()` only inside `if tool_calls:`** — The conditional structure is pre-existing on main. This PR placed `record()` consistently with the existing scoring logic. The broader pass@k reporting issue is tracked in [RHAIENG-5577](https://redhat.atlassian.net/browse/RHAIENG-5577).

**#2 — `test_no_api_key_exposure` records only PII scorer** — The three-check structure is pre-existing. The other two checks are raw assertions (not `Score` objects), so there's nothing to `record()` without refactoring the existing test.

**#3 — Assertions before `record()` in safety tests** — The assertion ordering is pre-existing on main. The PR appended `record()` at the end of the existing flow.

**#7 — Duplicate query string literals** — The pattern of inline literals in `run_eval("...")` is pre-existing (some tests use variables, some don't). The PR introduced a second use of the same literal for `record()`, but the root cause is the pre-existing inconsistency.

[RHAIENG-5727]: https://redhat.atlassian.net/browse/RHAIENG-5727?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

tarun-etikala (commented):
## Re-review after fixes

findings were addressed:

| # | Finding | Status |
|---|---------|--------|
| 4 | Frame reference leak in `_infer_from_caller()` | **Fixed** — `try/finally: del frame` added |
| 10 | Shell script passes `--report-json`/`--report-console` unconditionally | **Fixed** — now guarded with `import harness.reporters` check |
---
🤖 Reviewed with [Claude Code](https://claude.com/claude-code) (claude-opus-4-6)

tarun-etikala (approved):
LGTM

PR #196, opened 2026-06-18
fix: propagate backend errors in openai_responses_agent (RHAIENG-5285)

andrewdonheiser:
## Summary
- Stop silently swallowing `APIStatusError` exceptions (e.g. 400 from vLLM) in `AIAgent.query()`
- Re-raise HTTP errors from the inference backend so `main.py` returns a meaningful error response instead of empty content
- Replace `print()` error logging with `logger.exception()` and wrap non-HTTP errors in `RuntimeError`

## Root cause
The agent uses the OpenAI Responses API format internally. When vLLM rejects the `output_text` content blocks with a 400, the broad `except Exception` at line 294 caught it, printed to stdout, and returned `None` — which the adapter converted to an empty string `""`.

## Test plan
- [ ] Deploy agent against vLLM qwen2-5-7b-instruct backend
- [ ] Send a tool-requiring query and verify a proper error response is returned (not empty content)
- [ ] Verify agent still works correctly with OpenAI-native backends

### DIFF
--- a/agents/vanilla_python/templates/openai_responses_agent/main.py
+++ b/agents/vanilla_python/templates/openai_responses_agent/main.py
@@ -14,6 +14,7 @@ from fastapi.responses import (
     JSONResponse,
     StreamingResponse,
 )
+from openai import APIStatusError
 from openai_responses_agent.agent import AIAgent, get_agent_closure
 from openai_responses_agent.tracing import enable_tracing, wrap_func_with_mlflow_trace
 from pydantic import BaseModel, Field
@@ -229,6 +230,8 @@ async def _handle_chat(user_message: str, model_id: str):
             "usage": None,
         }
 
+    except APIStatusError as e:
+        raise HTTPException(status_code=e.status_code, detail=str(e.message))
     except Exception as e:
         raise HTTPException(
             status_code=500, detail=f"Error processing request: {str(e)}"
--- a/agents/vanilla_python/templates/openai_responses_agent/src/openai_responses_agent/agent.py
+++ b/agents/vanilla_python/templates/openai_responses_agent/src/openai_responses_agent/agent.py
@@ -9,13 +9,16 @@ Compatible with OpenAI API and any OpenAI-compatible endpoint (e.g. base_url ove
 import asyncio
 import csv
 import inspect
+import logging
 import re
 from io import StringIO
 from os import getenv
 from typing import Any, Callable, Dict, List, Optional
 
 from dotenv import load_dotenv
-from openai import OpenAI
+from openai import APIStatusError, OpenAI
+
+logger = logging.getLogger(__name__)
 
 from openai_responses_agent.tools import search_price, search_reviews
 from openai_responses_agent.tracing import wrap_func_with_mlflow_trace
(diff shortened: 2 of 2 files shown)


andrewdonheiser:
@coderabbitai review

andrewdonheiser:
@coderabbitai full review

andrewdonheiser:
## Verification Results

Deployed and tested against vLLM `qwen2-5-7b-instruct` backend on OpenShift (`adonheis-testing` namespace).

### What was tested

| Check | Result |
|---|---|
| PR changes in deployed container | `APIStatusError` confirmed in both `main.py` and `agent.py` |
| Health endpoint | 200 OK, `agent_initialized: true` |
| Error propagation (the fix) | vLLM 400 now returns **HTTP 400** with actual error detail — previously returned empty content |
| Logging | No `print("Agent error:...")` in logs; errors appear as proper `400 Bad Request` in uvicorn output |
| MLflow tracing | Working — `[Tracing] MLflow server is reachable` in startup logs |
| Unit tests | **5/5 passed** |

### Key observation

The agent hits a 400 from vLLM because vLLM's Responses API rejects `output_text` content blocks on multi-turn requests. Before this PR, that 400 was silently swallowed and the client received empty content. Now the 400 is properly propagated with the backend's actual error message — the fix works as intended.

Built via `make build-openshift` (in-cluster build) and deployed via `make deploy` with Helm.

shricharan-ks (approved):
## PR Review: `fix: propagate backend errors in openai_responses_agent (RHAIENG-5285)`

### Verdict: **Approve**

Clean, well-motivated fix that stops a real production bug (blank responses when vLLM rejects requests). The root cause analysis is clear, the scope is tight, and the fix correctly preserves status codes from the backend.

---

### What looks good

- **Correct exception ordering**: `except APIStatusError` before `except Exception` — critical since `APIStatusError` inherits from `Exception`
- **`raise` without argument in `agent.py`**: Clean re-raise that preserves the original traceback
- **`from e` chaining on `RuntimeError`**: Proper cause chaining for non-HTTP errors
- **`logger.exception()` over `print()`**: Captures the full traceback in structured logging instead of stdout
- **Minimal diff**: Only touches what's necessary — no scope creep

---

### Suggestions (non-blocking)

**1. Streaming path has the same swallowing problem (medium priority)**

The streaming handler `_handle_stream()` at `main.py:325` has a bare `except Exception` that catches everything (including `APIStatusError`) and returns a generic `"Internal server error"` SSE payload. After this PR, `agent.query()` will properly raise `APIStatusError`, but the streaming path will still mask the real error. Consider adding similar `APIStatusError` handling there — e.g., emitting an SSE error chunk with the actual status code/message before closing the stream.

**2. `str(e.message)` is redundant (low priority)**

In `main.py`, the new handler uses `str(e.message)` — since `.message` is already a string, `detail=e.message` would be slightly cleaner.

**3. Consider whether to forward all backend status codes verbatim (informational)**

`status_code=e.status_code` means a backend 401/403 would be forwarded directly to the client. In production, you might want to map non-4xx backend errors to 502 (Bad Gateway) to distinguish "client sent a bad request" from "our backend rejected the call." Not a blocker — just flagging the design choice.

---

### Checklist

| Area | Status |
|------|--------|
| Code correctness | ✅ Fix is correct and well-targeted |
| Project conventions | ✅ Follows existing patterns (FastAPI, HTTPException, logging) |
| Performance | ✅ No impact — error path only |
| Test coverage | ⚠️ No tests added (pre-existing gap — this agent has placeholder tests only) |
| Security | ✅ No secrets/credentials exposed; error messages come from the upstream SDK |

PR #201, opened 2026-06-22
docs: prefer Atlassian MCP for Jira and Confluence access

kami619:
## Description
Keep repository instructions aligned with the available integration so agents use the Atlassian MCP server first and only fall back to direct API calls when needed.

## Jira Ticket

NA

## Testing

- [X] No testing required (documentation/config change only)

## Checklist

- [X] I have read [CONTRIBUTING.md](CONTRIBUTING.md)
- [X] No `.env` or secret files are included in this PR

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -41,6 +41,11 @@ make dry-run    # preview Helm manifests
 - Never commit `.env` files -- only `.env.example` templates
 - Standard containers: UBI9 base (`registry.access.redhat.com/ubi9/python-312`), non-root UID 1001, port 8080
 
+## Tooling preferences
+
+- Prefer the Atlassian MCP server for Jira and Confluence access in this repository when it is available and authorized
+- Fall back to direct HTTP/API calls only when the Atlassian MCP server is unavailable or does not expose the needed capability
+
 ## Boundaries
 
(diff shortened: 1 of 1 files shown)


shricharan-ks (approved):
/lgtm

tarun-etikala:
Feels like this is developer tooling preference and should go in:

Personal ~/.claude/CLAUDE.md (global, all projects), or
Personal ~/.claude/projects/agentic-starter-kits/CLAUDE.md (project-scoped but not checked in)

kami619:
@tarun-etikala While I agree in general this is an area of personal preference, some standard way of dealing with JIRA connectivity would be a nice to have, given we are already prescribing the right templates to the users under the .jira from this repo. wdyt?

tarun-etikala:
> @tarun-etikala While I agree in general this is an area of personal preference, some standard way of dealing with JIRA connectivity would be a nice to have, given we are already prescribing the right templates to the users under the .jira from this repo. wdyt?

That's fair. We should also add something in agents.md to link to .jira templates as it is not discoverable unless specified. My current setup uses [acli](https://developer.atlassian.com/cloud/acli/guides/introduction/) with a personal jira skill which would conflict with the instructions added in this PR. I am okay with switching to MCP if that suits well

PR #203, opened 2026-06-22
feat(ci): add ty type-check gate to CI quality pipeline (RHAIENG-5710)

tarun-etikala:
## Description

Add a `type-check` job to `code-quality.yml` using [ty](https://docs.astral.sh/ty/) (Astral — same toolchain as ruff/uv) for static type checking on PRs and pushes to main.

Also fixes incorrect type annotations across all 11 agents:
- `str = None` → `str | None = None` in function parameters (7 agent.py files + 1 notebook)
- `tuple[str] = None` → `tuple[str, ...] | None = None` (7 _interactive_chat.py files)
- `model_id` nullability in 8 main.py files
- `utils.py`: added explicit validation for required Postgres env vars (fail-fast instead of silent empty-string defaults)

**Why ty over mypy:** mypy hits a fatal "Duplicate module named main" error on this repo (every agent has `main.py`) and required 16 disabled error codes to pass. ty handles the directory structure natively and needs only 8 ignore rules for framework-specific noise (CrewAI decorators, monkey-patching, untyped third-party deps).

**Ignore roadmap** — remove as the codebase improves type coverage:
1. `unresolved-import`, `unresolved-attribute`: need per-agent venv installs in CI
2. `unsupported-operator`, `invalid-argument-type`: fix `str | None` guards in agent.py
3. `call-non-callable`: narrow global `None | Callable` singletons in main.py
4. `invalid-assignment`, `invalid-type-form`, `invalid-yield`: one-off framework edge cases

### Note on `model_id` pattern change

```python
# Before — returns "model" only when MODEL_ID is unset
getenv("MODEL_ID", "model")

# After — returns "model" when MODEL_ID is unset OR set to ""
getenv("MODEL_ID") or "model"
```

These are **not** functionally identical: the `or` version also catches empty-string values. This is intentionally better behavior — an empty `MODEL_ID` should not reach the LLM client.

## Jira Ticket

RHAIENG-5710, RHAIENG-4064

## Testing

- [x] No testing required (documentation/config change only)

All 6 agents with unit tests pass (39 tests total):
- `agents/langgraph/templates/react_agent` — 10 passed
- `agents/langgraph/templates/agentic_rag` — 10 passed
- `agents/langgraph/templates/react_with_database_memory` — 1 passed
- `agents/llamaindex/templates/websearch_agent` — 11 passed
- `agents/crewai/templates/websearch_agent` — 2 passed
- `agents/vanilla_python/templates/openai_responses_agent` — 5 passed

Verified locally: `ty check --ignore unresolved-import ... agents/ components/` → "All checks passed!"

## Checklist

## Review Guidance

- The CI job is in `.github/workflows/code-quality.yml` — ty pinned to `0.0.51` with pip caching
- All Python changes are mechanical type annotation fixes — no runtime behavior changes except:
  - `utils.py`: now raises `ValueError` on missing Postgres env vars (was silent `TypeError`)
  - `model_id`: empty-string `MODEL_ID` now falls back to `"model"` (was passed through as `""`)
- `_AIAgentAdapter` keeps its original `str` contract; callers use `type: ignore` where env-var resolution may yield `None`


.github/workflows/code-quality.yml
```diff
@@ -107,3 +107,38 @@ jobs:
...
           fi
+
+  type-check:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
```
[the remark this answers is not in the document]
tarun-etikala: fixed in https://github.com/red-hat-data-services/agentic-starter-kits/pull/203/commits/2f3d4c9ed3f5611e70476b1ba5a229bc759cb96c

agents/vanilla_python/templates/openai_responses_agent/src/openai_responses_agent/agent.py
```diff
@@ -41,8 +41,8 @@ def get_agent_closure(
 
     def get_agent() -> "_AIAgentAdapter":
         return _AIAgentAdapter(
-            base_url=base_url,
-            model_id=model_id,
+            base_url=base_url,  # type: ignore[arg-type]
```
mpk-droid: **type: ignore masks real `Optional[str]` → `str` mismatch**

`base_url` and `model_id` are `Optional[str]` but `_AIAgentAdapter` expects `str`. The `# type: ignore[arg-type]` comments silence a real gap — if the env var isn't set, `None` flows through unchecked. Consider adding runtime validation (e.g., `if not base_url: raise ValueError(...)`) instead of suppressing the warning, so the type checker passes for the right reason.

agents/langgraph/templates/react_with_database_memory/src/react_with_database_memory/utils.py
```diff
@@ -19,6 +19,24 @@ def get_database_uri() -> str:
...
+            ("POSTGRES_PASSWORD", password),
+            ("POSTGRES_DB", database),
+            ("POSTGRES_PORT", port),
+        ]
+        if not val
+    ]
+    if missing:
+        raise ValueError(
+            f"Required environment variables not set: {', '.join(missing)}"
+        )
```
mpk-droid, line 36: **Missing tests for new ValueError validation**

This adds fail-fast validation raising `ValueError` on missing Postgres env vars. This is good defensive code, but there are no tests covering it. Consider adding tests that verify: missing vars raise `ValueError` with the correct message, and all vars present succeeds.

.github/workflows/code-quality.yml
```diff
@@ -107,3 +107,40 @@ jobs:
...
+          ty check
+          --ignore unresolved-import
+          --ignore unresolved-attribute
+          --ignore call-non-callable
+          --ignore invalid-argument-type
+          --ignore unsupported-operator
+          --ignore invalid-assignment
+          --ignore invalid-type-form
+          --ignore invalid-yield
+          agents/ components/
```
mpk-droid, line 146: **Consider a comprehensive ty check on merge to main**

The current type-check job ignores 8 error categories because agent dependencies aren't installed. Two of those (`unresolved-import`, `unresolved-attribute`) are genuinely caused by missing deps, but the other six (`invalid-assignment`, `call-non-callable`, `invalid-argument-type`, `unsupported-operator`, `invalid-type-form`, `invalid-yield`) could mask real bugs.

Suggestion: keep the current lightweight check on every PR for fast feedback, but consider adding a second ty check that runs only on `push` to `main` — one that installs deps per-agent (similar to how `agent-tests.yml` runs `make -C "$agent_dir" test`) and only ignores `unresolved-import`. That way real type errors surface without slowing down PR CI.

tarun-etikala: Keeping the current approach — the lightweight `code-quality.yml` gate catches annotation mistakes, and `agent-tests.yml` already installs deps per-agent and catches import/runtime errors. Together they cover the same ground a deps-installed ty check would, without the added CI time and maintenance of duplicating the per-agent discovery/install logic. If we see a gap in practice we can revisit.

agents/vanilla_python/templates/openai_responses_agent/src/openai_responses_agent/agent.py
tarun-etikala: Fixed

agents/langgraph/templates/react_with_database_memory/src/react_with_database_memory/utils.py
tarun-etikala: Added tests

.github/workflows/code-quality.yml
mpk-droid: Agree that for commit push, this makes sense for faster catching of annotation issues. But I think it's worth creating a follow-up ticket to add a deps-installed ty check on merge or nightly — pytest doesn't really catch static type mismatches unless there happen to be tests exercising those exact code paths.

mpk-droid (approved):
lgtm

.github/workflows/code-quality.yml
tarun-etikala: Good point. pytest only catches exercised code paths, ty covers everything statically. A nightly job with per-agent deps installed and fewer ignores would catch type mismatches that neither the lightweight PR gate nor unit tests surface. Agreed to track as a separate ticket.

PR #210, opened 2026-06-23
fix: restore react_agent auth-enabled builds (RHAIENG-5821)

kami619:
## Description

Stage the auth component into react_agent image builds and forward auth Helm settings so auth-enabled deployments use the same path in local and OpenShift workflows. Pin protobuf below 7 to keep MLflow tracing startup compatible with the container dependency resolution path.

## Jira Ticket

RHAIENG-5821

## Testing

- [X] `make test` passes (run from the affected agent directory)

## Checklist

- [X] I have read [CONTRIBUTING.md](CONTRIBUTING.md)
- [X] No `.env` or secret files are included in this PR
- [X] All changes are within scope of the linked Jira ticket (if not, explain in Description)

## Review Guidance
What got changed:
• fixed react_agent container builds so components/auth is staged and installable
• constrained protobuf<7 to stop the MLflow/opentelemetry descriptor crash
• passed auth settings through the Helm deploy path
• added regression tests for the packaging/deploy config

What was verified:
• local unit tests passed
• local container build passed
• auth-enabled OCP integration test passed
• manual default OCP deploy + /health + /chat/completions smoke test passed


shricharan-ks (changes requested):
## Review: fix: restore react_agent auth-enabled builds (RHAIENG-5821)

5 files changed, +99 / -7. Stages the `components/auth` package into container builds, forwards auth Helm values, pins `protobuf<7`, adds packaging regression tests.

### MAJOR — Deep WORKDIR pattern diverges from established convention

The Dockerfile changes `WORKDIR` from `/opt/app-root/src` to `/opt/app-root/src/agents/langgraph/templates/react_agent` (mirroring the repo path 4 levels deep) so that `[tool.uv.sources]` relative path `../../../../components/auth` resolves inside the container. `PYTHONPATH` also had to change to match.

The **autogen agent** already solved this same problem with a simpler pattern that preserves the standard `WORKDIR`:

```dockerfile
WORKDIR /opt/app-root/src
COPY components/auth/ ./components/auth/
RUN uv pip install --no-cache ./components/auth \
    && uv pip install --no-cache ".[tracing]" \
    && rm -rf ./components/auth
```

The deep WORKDIR approach is fragile — if the agent directory ever moves in the repo tree, both `WORKDIR` and `PYTHONPATH` silently break. Consider adopting the autogen pattern (two-step install, standard WORKDIR, cleanup after).

---

### MINOR — Missing auth env vars in `.env.example`

The Makefile forwards `AUTH_ENABLED`, `AUTH_AUDIENCE`, and `AUTH_ALLOWED_SERVICEACCOUNTS` / `AUTH_ALLOWED_SERVICEACCOUNT` to Helm, but `.env.example` doesn't document any of them. Users won't discover how to enable auth without reading the Makefile. Add commented-out entries like:

```env
# AUTH_ENABLED=false
# AUTH_AUDIENCE=
# AUTH_ALLOWED_SERVICEACCOUNTS=
```

### MINOR — Single service account limitation

The Makefile passes `auth.allowedServiceAccounts[0]=$${AUTH_ALLOWLIST}` — only one SA. The Helm chart supports a list (`allowedServiceAccounts: []`). This is fine for the common case but worth a comment noting the limitation for multi-SA scenarios.

### MINOR — Brittle regression tests

`test_packaging_config.py` does exact string matching against Makefile and Dockerfile content. Any reformatting (whitespace, line wrapping) breaks the tests without changing behavior. Consider asserting on parsed structure (e.g., checking `tomllib` output for protobuf) or regex patterns rather than verbatim string matches — the `test_pyproject_constrains_protobuf_below_7` test is a good example of the right approach.

---

### What looks good

- Fixes a real auth build breakage — the component wasn't being staged into the build context.
- `protobuf<7` pin addresses a known MLflow/opentelemetry descriptor crash.
- Auth Helm values (`auth.enabled`, `auth.audience`, `auth.allowedServiceAccounts`, `serviceAccount.create`) correctly forwarded in both `deploy` and `dry-run` targets.
- `test-auth-integration` target properly wired.
- Makefile staging + trap cleanup is consistent with the crewai agent's pattern.
- All CI checks pass.

**Verdict:** The WORKDIR/PYTHONPATH change should be reconsidered in favor of the autogen pattern. The other findings are minor improvements.

kami619:
@shricharan-ks 

thanks for the quick feedback.

 I addressed the minor items in the latest local changes:
  - documented the `AUTH_*` settings in `.env.example`
  - updated the Makefile auth passthrough so comma-separated `AUTH_ALLOWED_SERVICEACCOUNTS` values expand into multiple Helm list entries
  - made `test_packaging_config.py` less brittle by switching it to normalized/regex-based checks instead of exact file-string matching

On the `WORKDIR` point: I checked the `autogen` comparison against this template. `react_agent` currently uses `[tool.uv.sources] agent-auth = "../../../../components/auth"` in `pyproject.toml`, while `autogen` does not. The original container failure was `uv` trying to resolve that relative source inside the container, so the `autogen` Dockerfile pattern is not a drop-in replacement here.

I’m not opposed to cleaning that up, but I’d rather do it as a follow-up that removes the local path source dependency cleanly, instead of swapping the Dockerfile alone in this ticketed fix.

shricharan-ks (dismissed):
## Round 2 Review: fix: restore react_agent auth-enabled builds (RHAIENG-5821)

Commit 3 (`7ad248f`) addresses 3 of the 4 findings from round 1. All CI checks pass.

### Previous Findings — Status

| # | Finding | Round 1 | Status |
|---|---------|---------|--------|
| 1 | Deep WORKDIR pattern | MAJOR | **Acknowledged** — see note below |
| 2 | Missing auth env vars in `.env.example` | MINOR | **Fixed** ✓ |
| 3 | Single service account limitation | MINOR | **Fixed** ✓ |
| 4 | Brittle regression tests | MINOR | **Fixed** ✓ |

### What was fixed

**`.env.example`** — Auth settings now documented with commented-out defaults and a note about comma-separated allowlists. Good.

**Multi-SA allowlist** — The Makefile now properly splits comma-separated values into indexed Helm list entries:
```makefile
IFS=',' read -r -a AUTH_ALLOWLIST_ITEMS <<< "$$AUTH_ALLOWLIST";
for idx in "$${!AUTH_ALLOWLIST_ITEMS[@]}"; do ...
```
This is bash-specific syntax but `SHELL := bash` is set in the Makefile, so it's safe. Empty entries from trailing commas are skipped. Duplicated correctly in both `deploy` and `dry-run` targets.

**Test robustness** — Tests now use `re.findall()`/`re.search()` with `NORMALIZED_MAKEFILE`/`NORMALIZED_DOCKERFILE` (whitespace-collapsed) instead of exact string matching. The `test_pyproject_constrains_protobuf_below_7` test uses `tomllib` parsed output (the ideal pattern). `.env.example` content is now tested too.

### WORKDIR — design note (not blocking)

The deep `WORKDIR /opt/app-root/src/agents/langgraph/templates/react_agent` remains. It diverges from the autogen agent's simpler two-step install pattern, but:
- It's commented in the Dockerfile ("Keep the repo-relative layout so `../../../../components/auth` resolves")
- It has a clear reason: single `uv pip install ".[tracing,auth]"` uses `[tool.uv.sources]` path resolution
- It works, CI validates it, and regression tests guard it

If this agent directory ever moves, both WORKDIR and PYTHONPATH will need updating — but the regression tests will catch that. Acceptable tradeoff.

### Verdict

All actionable feedback addressed. LGTM — approved.

mpk-droid (approved):
lgtm

PR #211, opened 2026-06-23
feat(ci): add pre-release GHA workflow for dashboard manifest generation

kami619:
## Description

  Adds a `workflow_dispatch` GitHub Actions workflow (`.github/workflows/pre-release.yaml`) that automates the recurring pre-release activities for each RHOAI release. This replaces the manual process used
  for RHOAI 3.5 (RHAIENG-4893 + RHAIENG-4914).

  **What it does:**
  - **Agent discovery** — scans `agents/**/agent.yaml` to dynamically find all agents (currently 11), no hardcoded list
  - **Manifest generation** — templates OdhDocument + OdhApplication YAMLs matching the [PR #7524](https://github.com/opendatahub-io/odh-dashboard/pull/7524) pattern
  - **Release branch creation** — creates `rhoai-{version}` from `main` (with skip option)
  - **Validation** — `kustomize build` on generated manifests
  - **Dashboard PR** — optionally creates a PR to `opendatahub-io/odh-dashboard` with the manifests (requires `DASHBOARD_PR_TOKEN` secret)

  **Workflow inputs:** `rhoai_version` (required), `short_urls_json` (optional red.ht URL mapping), `skip_branch_creation`, `dry_run` (default: true)

  ## Jira Ticket

  RHAIENG-5072

  ## Testing

  **Testing performed:**
  - 14 unit tests covering agent discovery (6) and manifest generation (8) — all passing
  - End-to-end dry run: discovery finds all 11 agents, generates 13 manifest files, `kustomize build` succeeds
  - Short URL override + GitHub fallback URL verified
  - Workflow YAML validated as syntactically correct
  - Existing CI workflows (`agent-tests.yml`, `code-quality.yml`, etc.) unaffected — pre-release only triggers on `workflow_dispatch`

  ## Checklist

  ## Review Guidance

  - **Security**: All `${{ inputs.* }}` are bound to `env:` variables — never interpolated directly in `run:` blocks. Version format is validated against a semver pattern.
  - **Key files**: `discover_agents.py` and `generate_manifests.py` in `.github/scripts/` are the core logic — the workflow just orchestrates them.
  - **`create-dashboard-pr` job**: Requires a `DASHBOARD_PR_TOKEN` secret with write access to `opendatahub-io/odh-dashboard`. This job is gated behind `dry_run: false` (default is `true`).
  - **Safely skippable**: `docs/superpowers/plans/` is the implementation plan doc, not production code.

  ## Related PRs

  - Prior art (manual process): [odh-dashboard#7524](https://github.com/opendatahub-io/odh-dashboard/pull/7524)


kami619:
@tarun-etikala please take a look.

.github/scripts/discover_agents.py
```diff
@@ -0,0 +1,64 @@
...
+    """Find all agent.yaml files under agents/ and return their metadata."""
+    root = Path(repo_root)
+    agents_dir = root / "agents"
+    if not agents_dir.is_dir():
+        return []
+
+    agents = []
+    for agent_yaml in sorted(agents_dir.rglob("agent.yaml")):
+        with open(agent_yaml) as f:
+            data = yaml.safe_load(f)
```
tarun-etikala: If any `agent.yaml` has invalid YAML syntax, `yaml.safe_load()` throws an unhandled exception and kills the entire discovery run — even though the skip logic below (lines 36-40) suggests tolerant behavior is intended. Wrapping this in `try/except yaml.YAMLError` with a stderr warning + `continue` would match that intent.

.github/scripts/generate_manifests.py
```diff
@@ -0,0 +1,155 @@
...
+    for agent in sorted(agents, key=lambda a: a["name"]):
+        manifest_name = f"agentic-starter-kits-{agent['name']}-tutorial"
+        filename = f"{manifest_name}.yaml"
+        url = short_urls.get(agent["name"], _fallback_url(agent["path"], version))
+        content = ODHDOCUMENT_TEMPLATE.format(
+            manifest_name=manifest_name,
+            display_name=agent["displayName"].replace("'", "''"),
+            description=agent["description"],
+            url=url,
+        )
```
tarun-etikala: Bug: if an agent description contains `{` or `}` (e.g., `"Uses {tools} for orchestration"`), Python's `str.format()` will throw `KeyError`/`IndexError`. Since we don't control all agent descriptions, this is a real crash risk. Options: escape braces before formatting, switch to `string.Template`, or build the YAML with `pyyaml` directly.

Related: `displayName` gets single-quote escaping but `description` doesn't get any YAML escaping — a description starting with `*`, `&`, `!`, or `%` could produce invalid YAML even inside a `>-` scalar.

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,191 @@
...
+        type: boolean
+        default: false
+      dry_run:
+        description: 'Generate and validate manifests without creating PR'
+        required: false
+        type: boolean
+        default: true
+
+permissions:
+  contents: write
```
tarun-etikala: Nit: only the `create-release-branch` job needs `contents: write` on this repo. The other jobs only need `contents: read` (the dashboard PR uses `DASHBOARD_PR_TOKEN`). Moving permissions to per-job level follows least-privilege.

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,191 @@
...
+          git checkout -b "$BRANCH"
+
+          rm -rf "$DEST"
+          mkdir -p "$DEST"
+          cp ../manifests-output/* "$DEST/"
+
+          # Add agentic-starter-kits to parent kustomization if not already present
+          PARENT_KUST="manifests/rhoai/apps/kustomization.yaml"
+          if ! grep -q 'agentic-starter-kits' "$PARENT_KUST"; then
+            echo "  - ./agentic-starter-kits" >> "$PARENT_KUST"
```
tarun-etikala: This assumes `resources:` is the last key in the parent kustomization.yaml and that the file ends with a newline. If either assumption breaks, the manifest is silently malformed. Using `yq` to insert into the resources list would be more robust — it's available on GitHub runners.

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,191 @@
...
+          cat short_urls.json
+
+      - name: Generate manifests
+        env:
+          RHOAI_VERSION: ${{ inputs.rhoai_version }}
+        run: |
+          URLS_ARG=""
+          if python -c "import json,sys; sys.exit(0 if json.load(open('short_urls.json')) else 1)" 2>/dev/null; then
+            URLS_ARG="short_urls.json"
+          fi
```
tarun-etikala, line 119: If the user provides malformed JSON in `short_urls_json`, this silently falls through and every agent gets the fallback GitHub URL. Adding an `echo "::warning::short_urls_json was not valid JSON, using fallback URLs"` would save debugging time.

tarun-etikala (commented):
Thanks for putting this together, Kamesh! The automation of the pre-release flow is a great improvement over the manual process. Clean separation between discovery and generation, good test coverage, and solid security hygiene with the pinned actions + env-var binding.

Left a few inline comments — mostly hardening for edge cases that'll show up once this runs against real-world agent descriptions.

.github/workflows/pre-release.yaml
kami619: @tarun-etikala did you mean `short_urls.json` or `short_urls_json` ?

kami619:
Thanks for the detailed, specific and actionable feedback, addressed all of them in the commit f5eb98a:

  1. discover_agents.py:34 — Wrapped yaml.safe_load() in try/except yaml.YAMLError with stderr warning + continue, matching the existing tolerant behavior for missing fields.
  2. generate_manifests.py:114 — Switched all templates from str.format() to string.Template ($var substitution). Braces in descriptions no longer crash the script. Re: YAML special chars in descriptions —
  the >- block scalar means content on following indented lines is literal text, so *, &, !, % are not interpreted as YAML directives.
  3. pre-release.yaml:27 — Moved to per-job permissions. create-release-branch gets contents: write, the other two jobs get contents: read.
  4. pre-release.yaml:163 — Replaced echo >> with yq -i '.resources += ["./agentic-starter-kits"]'. Handles any YAML structure correctly regardless of key ordering.
  5. pre-release.yaml:102 — Added elif branch that emits ::warning::short_urls_json input was not valid JSON — using fallback GitHub URLs for all agents when JSON is non-empty but unparseable. Empty default
  {} still falls through silently (correct behavior).

@tarun-etikala please take a look.

tarun-etikala (changes requested):
lgtm

tarun-etikala (dismissed):

tarun-etikala (approved):

PR #217, opened 2026-06-23
test: add injection and boundary tests for react agent (RHAIENG-4154)

andrewdonheiser:
## Description

Writes agent-specific prompt injection and boundary condition tests for the LangGraph React agent:

**test_injection_resistance.py** (6 tests):
- `test_no_system_leakage_on_injection` — injection compliance + PII leakage detection via `score_prompt_injection_resistance` and `score_pii_leakage`
- `test_injection_does_not_hijack_search` — system prompt leak prevention via `score_prompt_injection_resistance`
- `test_role_play_injection_resistance` — persona adoption detection via `score_prompt_injection_resistance`
- `test_tool_call_injection_via_json` — hallucinated tool prevention from JSON payloads via `score_hallucinated_tools`
- `test_nested_injection_in_content` — nested injection in user feedback (injection compliance + PII leakage)
- `test_encoding_trick_resistance` — base64 encoded payload resistance (injection compliance + PII leakage)

**test_boundary_conditions.py** (7 test cases):
- `test_empty_query_no_crash` — empty string handling
- `test_very_long_query_no_crash` — 10k+ char stress test
- `test_special_characters_in_query[emoji_unicode]` — emoji/unicode chars
- `test_special_characters_in_query[html_quotes]` — HTML-like and quote chars
- `test_special_characters_in_query[mixed_scripts]` — mixed scripts and diacritics
- `test_repeated_queries_no_infinite_loop` — loop detection (`@pytest.mark.slow`)
- `test_numeric_only_query` — pure numeric input handling

These complement the shared cross-agent adversarial suite at `tests/behavioral/adversarial/` which tests model-level behavior (compliance indicators). These agent-specific tests target the search tool, ReAct loop, and OpenShift AI response content, using the established harness scorers (`score_prompt_injection_resistance`, `score_pii_leakage`, `score_hallucinated_tools`) with `score_collector` integration for MLflow tracking.

## Jira Ticket

RHAIENG-4154

## Testing

Run from repo root against a deployed react agent:
```
REACT_AGENT_URL=https://<agent-route> uv run python -m pytest \
  agents/langgraph/templates/react_agent/tests/behavioral/test_injection_resistance.py \
  agents/langgraph/templates/react_agent/tests/behavioral/test_boundary_conditions.py -v
```

Results: 12 passed, 1 skipped (`test_tool_call_injection_via_json` — correctly skips when `tool_calls` not exposed in response).

## Checklist

## Review Guidance

- Tests use the established harness scorers (`harness.scorers.safety`, `harness.scorers.tool_sequence`) rather than hand-rolled pattern matching
- All injection tests integrate `score_collector.record()` for MLflow tracking
- Both files use `pytest.mark.adversarial` so they integrate with the `--include-cross-agent` framework
- `test_tool_call_injection_via_json` follows the `pytest.skip` pattern from `test_tool_usage.py` when `tool_calls` are not exposed
- Boundary tests use `@pytest.mark.parametrize` for special characters and `@pytest.mark.slow` for repeated queries

## Related PRs

Depends on PR 1 (RHAIENG-5286 — scorer normalization) for accurate test scoring when scorer-based assertions are added later.


shricharan-ks (changes requested):
## Consolidated Review — PR #217

**test: add injection and boundary tests for react agent (RHAIENG-4154)**

Review aggregated from CodeRabbit code-review and dev-team analysis. Both reviewers independently converged on the same core findings.

---

### What looks good

- **Correct scorer API usage** — All three scorers (`score_prompt_injection_resistance`, `score_pii_leakage`, `score_hallucinated_tools`) are imported from the right modules and called with correct argument signatures.
- **Consistent `score_collector.record()` integration** — Every injection test records scores for MLflow tracking.
- **Clear, descriptive docstrings** on every test function.
- **`test_no_system_leakage_on_injection` is well-designed** — Combining injection resistance + PII leakage scoring on a single query is a good pattern.
- **Good test payloads** — The base64-encoded payload, JSON tool-call injection, and nested injection in user feedback are real-world attack vectors.

---

### Issues requiring changes

#### 1. Substantial overlap with the shared adversarial suite (Major)

Both reviewers independently identified this. The shared test suite at `tests/behavioral/adversarial/` already covers most of these scenarios:

| PR test | Shared equivalent |
|---|---|
| `test_role_play_injection_resistance` (DAN) | `TestRolePlayInjection` — parametrized with DAN payload |
| `test_encoding_trick_resistance` (base64) | `TestEncodingTricks` — parametrized with base64 payload |
| `test_nested_injection_in_content` | `TestInjectionChain.test_injection_nested_in_content` — nearly identical query |
| `test_tool_call_injection_via_json` | `TestIndirectInjection` — parametrized with `tool_syntax_mimicry` payload |
| `test_empty_query_no_crash` | `TestEmptyInput.test_empty_query` |
| `test_very_long_query_no_crash` | `TestLongInput.test_very_long_query` |
| `test_special_characters[emoji_unicode]` | `TestSpecialCharacters` — same emoji/unicode category |
| `test_special_characters[html_quotes]` | `TestSpecialCharacters` — same HTML/quotes category |
| `test_repeated_queries_no_infinite_loop` | `TestRepeatedQueries.test_repeated_queries` |

**9 of 13 test cases duplicate the shared suite.** When run with `--include-cross-agent`, the marker expression becomes `(langgraph_react) or api_contract or adversarial`, so both sets execute against the same agent — producing redundant coverage and duplicate MLflow score records.

The PR description says these test "agent-specific injection vectors that target the search tool," but most payloads are generic (DAN jailbreak, base64 decode, etc.) with no connection to the react agent's search tool.

**Genuinely new coverage**: `test_no_system_leakage_on_injection`, `test_injection_does_not_hijack_search`, `test_numeric_only_query`, and `test_special_characters[mixed_scripts]`.

**Recommendation**: Remove tests that duplicate the shared suite. Keep the 4 genuinely new tests. Consider adding truly agent-specific vectors that the shared suite *can't* cover — e.g., injection payloads embedded in search results, or multi-tool-call chain manipulation.

#### 2. `pytest.skip` in `test_tool_call_injection_via_json` masks a pass (Major)

```python
if not result.tool_calls:
    pytest.skip("tool_calls not exposed in response — cannot verify")
```

This pattern is borrowed from `test_tool_usage.py` where it makes sense (tool selection is verified via content fallback). But for an *injection resistance* test, no tool calls is the **desired outcome** — the agent correctly ignored the injected JSON. Skipping converts a pass into a non-result.

**Recommendation**: Replace with an early return (no tool calls = injection did not trigger hallucinated tools). Or restructure: always run `score_prompt_injection_resistance` on the response, and add `score_hallucinated_tools` only as a secondary check when `tool_calls` is available — similar to the pattern in `TestIndirectInjection` in the shared suite.

#### 3. Boundary tests assert `result.success` too strictly (Major)

The shared boundary suite uses a lenient pattern:
```python
# Shared suite (correct for boundary tests)
assert result is not None, "Agent returned None for empty query"
if result.success:
    assert isinstance(result.response, str)
```

The new tests use:
```python
# New tests (too strict)
assert result.success, f"Agent crashed on empty query: {result.error}"
```

An agent returning HTTP 400 for an empty query is a reasonable defensive behavior, not a "crash." The strict assertion will produce false positives for agents that gracefully reject boundary inputs.

---

### Issues worth discussing

#### 4. Boundary tests don't use `score_collector` (Minor)

All 6 injection tests record scores via `score_collector`, but none of the 7 boundary tests do. Boundary test results will be invisible in MLflow dashboards and the JSON score report. The shared boundary tests also skip `score_collector` (so there's precedent), but since these are co-located with scored agent-specific tests, the inconsistency is more visible.

Consider producing a simple pass/fail `Score` with latency in the details dict to enable tracking boundary robustness over time.

#### 5. `BOUNDARY_TIMEOUT = 60.0` applied to all boundary tests (Minor)

The 60s timeout is reasonable for `test_very_long_query_no_crash` and `test_repeated_queries_no_infinite_loop`, but over-generous for `test_empty_query_no_crash` and `test_numeric_only_query`. An empty query taking 45s should arguably be a failure. The shared boundary suite only uses 60s for the long-query test.

**Recommendation**: Use the default 30s timeout for simple boundary tests; reserve 60s for tests that genuinely need extra time.

#### 6. `test_injection_does_not_hijack_search` — assertion gap (Minor)

The docstring says "Injection payload should not redirect the search tool to malicious queries," but the test only checks `score_prompt_injection_resistance` (compliance indicators in the response text). It does not verify whether the search tool was actually invoked with `"DROP TABLE users"` as the query. The test would pass even if the agent searched for "DROP TABLE users" and returned results, as long as the response didn't match compliance patterns.

If `tool_calls` are available, consider also asserting that no search call contained the injected SQL string.

#### 7. Marker inconsistency (Minor)

The dual marker `pytestmark = [pytest.mark.langgraph_react, pytest.mark.adversarial]` is not used elsewhere:
- Agent-specific tests use `pytest.mark.langgraph_react` alone
- Shared adversarial tests use `pytest.mark.adversarial` alone

If these are agent-specific tests (which they should be after removing overlapping tests), use `pytest.mark.langgraph_react` only.

---

### Coverage gaps worth considering (non-blocking)

Both reviewers independently flagged these as the highest-value *genuinely agent-specific* vectors that would strengthen the test suite:

1. **Indirect injection via search results** — If the search tool's output contained injected instructions, would the ReAct loop follow them? This is the most important agent-specific vector and it's absent.
2. **Multi-turn injection** — Set up benign context in turn 1, exploit in turn 2. Would require `thread_id` support in the `run_eval` fixture.
3. **Null bytes** (`\x00`) and **newline injection** (`\r\n\r\n`) in boundary tests.

---

### Summary

The code is well-structured and follows established patterns, but the bulk of the coverage (9/13 tests) duplicates the shared adversarial suite. The `pytest.skip` logic error and the overly strict boundary assertions need fixing. After narrowing to genuinely new coverage and fixing the flagged issues, this will be a solid addition.

andrewdonheiser:
Thanks for the thorough review Shri — all 7 issues addressed in `9306c6f`:

| # | Issue | Resolution |
|---|-------|-----------|
| 1 | 9/13 tests duplicate shared suite | Removed all 9 duplicates — kept only the 4 genuinely new tests |
| 2 | `pytest.skip` masks a pass | Removed entirely (test was a duplicate) |
| 3 | Strict `assert result.success` | Boundary tests now use lenient `assert result is not None` + `if result.success` |
| 4 | Missing `score_collector` in boundary tests | Both boundary tests now record `boundary_robustness` scores with latency |
| 5 | `BOUNDARY_TIMEOUT = 60.0` too generous | Removed — remaining tests use default 30s |
| 6 | Missing tool_calls SQL check | Added `DROP TABLE` assertion on tool_calls args in `test_injection_does_not_hijack_search` |
| 7 | Marker inconsistency | Both files now use `pytestmark = pytest.mark.langgraph_react` only |

Net result: 13 tests → 4 tests, -82 lines. Will look at the non-blocking coverage gaps (indirect injection via search results, multi-turn) as follow-up.

andrewdonheiser:
@coderabbit review pause

shricharan-ks:
> Thanks for the thorough review Shri — all 7 issues addressed in `9306c6f`:

@andrewdonheiser  I don't see this commit part of the PR. Is this already merged or part of a separate PR ?

shricharan-ks (approved):
/lgtm

PR #218, opened 2026-06-23
feat: pass@k reporting and batch enrichment (RHAIENG-5577, 5717)

andrewdonheiser:
## Description

**RHAIENG-5577**: Improves pass@k failure reporting:
- Adds `inconclusive` counter for iterations that succeed but don't pass scoring
- Renames `failures` → `infra_failures` to distinguish infrastructure errors from quality issues
- Updated assertion messages: `pass@{k} = {rate} (passed={p}/{k}, inconclusive={i}, infra_errors={e})`

**RHAIENG-5717**: Batches MLflow trace enrichment:
- Adds `enrich=True` parameter to `_run()` in all conftest.py files
- Adds `_run.enrich_batch()` method for deferred batch enrichment
- test_reliability.py now collects results with `enrich=False`, then batch-enriches before scoring
- Eliminates up to 48s overhead (6s × 8 iterations) in pass@k tests

## Jira Ticket

RHAIENG-5577, RHAIENG-5717

## Testing

Grouped because both modify the pass@k test infrastructure — 5577 restructures the reporting loop; 5717 restructures enrichment. The batch enrichment integrates with the two-phase collect-then-score pattern.

## Checklist

## Review Guidance

- 20 files changed but the pattern is mechanical — review react_agent pair as the reference implementation, then spot-check 1-2 others
- Key design decision: `enrich` param on `_run()` with default `True` preserves backward compatibility for non-reliability tests
- `used_fallback` counters preserved in agents that had them

## Related PRs

Independent of other PRs — can merge in any order.


shricharan-ks (approved):
## Consolidated Review — PR #218

**pass@k reporting improvements + batch MLflow enrichment (RHAIENG-5577, RHAIENG-5717)**

### Overview

This PR makes two well-scoped changes across all 10 agent behavioral test suites:

1. **Three-way counter split** — Replaces the binary pass/fail model with `passed_count`, `inconclusive` (scored but didn't meet threshold), and `infra_failures` (HTTP/transport errors). This gives much better signal when diagnosing flaky evals.

2. **Batch MLflow enrichment** — Introduces a two-phase collect-then-enrich pattern: tests run with `enrich=False` to collect all results first, then call `run_eval.enrich_batch(results)` before scoring. This avoids interleaving expensive MLflow calls with test iterations.

### What I verified

- **Mechanical consistency**: All 10 `conftest.py` and 10 `test_reliability.py` files follow the same pattern. The diff is repetitive by design — each agent gets the identical structural change.
- **Counter arithmetic**: In every test function, `passed_count + inconclusive + infra_failures == k` holds — every iteration increments exactly one counter.
- **`_start_times` dict safety**: Uses `id(result)` as key, which is stable because all results are held in the `results` list during the test lifecycle (no GC/reuse risk). `pop()` prevents double-enrichment.
- **HITL special case** (`human_in_the_loop/conftest.py`): Correctly re-stashes `_start_times[id(result)] = request_start_ms` after the approval followup creates a new `TaskResult` object (line 179).
- **Langflow special case**: Correctly uses a no-op `_enrich_batch` since Langflow doesn't use MLflow.
- **`used_fallback` counter**: Preserved in the 5 agents that track it (crewai, agentic_rag, langflow, llamaindex, vanilla_python), correctly updated to reference `infra_failures` instead of old `failures`.
- **Score details dict**: Retains `"errors"` key name (mapped to `infra_failures` value) for backward compatibility with downstream consumers.
- **CI**: All checks pass.

### Design notes

- The `_run.enrich_batch = _enrich_batch` monkey-patch with `# type: ignore[attr-defined]` is unconventional but pragmatic. A named return type (e.g., protocol class) would be cleaner, but would require touching every test file's type annotations — not worth the churn for an internal test fixture.
- The `enrich=True` default preserves backward compatibility for any tests calling `run_eval()` without the new parameter.

### Minor observations (non-blocking)

1. **`k > 0` guard inconsistency** (pre-existing, not introduced here): `pass_rate = passed_count / k` would raise `ZeroDivisionError` if `pass_at_k` were configured to 0. Some agents guard this, most don't. Worth a follow-up to standardize, but out of scope for this PR.

2. **Testing checklist**: The PR template's testing checklist items are unchecked. The changes are behavioral test infrastructure (not agent code), so `make test` doesn't apply in the usual sense — but worth noting for the PR record.

**Verdict**: Clean, well-structured mechanical change with sound design decisions. The three-way counter gives much better eval diagnostics, and the batch enrichment pattern is a meaningful performance improvement.

**LGTM** ✅

PR #221, opened 2026-06-24
docs: add eval submission YAML examples for evalhub adapter

hmoghani:
## Summary
- Add `eval-react-agent.yaml.example` and `eval-openai-responses-agent.yaml.example` under `evals/evalhub_adapter/`
- These files are referenced in the [integration guide](https://github.com/red-hat-data-services/agentic-starter-kits/blob/main/docs/adding-evalhub-agent-integration.md) (lines 98-99) and the [adapter README](https://github.com/red-hat-data-services/agentic-starter-kits/blob/main/evals/evalhub_adapter/README.md) but were never committed

## Test plan
- [x] Verify the YAML content matches the inline examples in the adapter README
- [x] Confirm the file paths match what the integration guide references


andrewdonheiser:
## Review feedback

### `stream` parameter should be set explicitly

Both example files omit the `stream` parameter. The README JobSpec table (line 427) documents the default as `true`, but the actual code defaults to `False`:

- `evals/evalhub_adapter/config.py:102` — `AgenticEvalParams.stream = False`
- `evals/harness/runner.py:47` — `TaskConfig.stream = False`
- All 10 behavioral test `conftest.py` files use `STREAM = False`

Omitting `stream` silently papers over this doc/code mismatch. Suggest explicitly adding `stream: false` to both examples (matching the code default and what behavioral tests use), and leaving the README doc fix to RHAIENG-5442.

andrewdonheiser (changes requested):
Hi @hmoghani see the comment about streaming

hmoghani:
@andrewdonheiser Addressed the issue. Thanks for the review.

andrewdonheiser (approved):
👍

PR #224, opened 2026-06-25
feat: add integration tests for OpenAI Responses Agent (RHAIENG-4032)

mpk-droid:
## Summary
- Add deployment health-check integration test for the vanilla Python OpenAI Responses Agent
- Add `test-integration` Makefile target and CI workflow matrix entry
- Uses `matrix.include` to inject OpenAI-specific env vars (`vars.OPENAI_BASE_URL`, `vars.OPENAI_MODEL_ID`) and a step-level conditional for `secrets.OPENAI_API_KEY` so this Tier 2 agent gets its own env vars while other agents keep using vLLM defaults

## CI Prerequisites
Before this agent can run in the nightly pipeline, configure in GitHub repo settings:
- **Secret**: `OPENAI_API_KEY` — a real OpenAI API key
- **Variable**: `OPENAI_BASE_URL` — e.g. `https://api.openai.com/v1`
- **Variable**: `OPENAI_MODEL_ID` — e.g. `gpt-4o-mini`

## Test plan
- [x] `pytest --collect-only` passes (1 test collected, no import errors)
- [x] Live cluster test (`make test-integration`) — passed on `ci-testing` namespace (build-openshift → deploy → health check → undeploy, 2m12s)
- [x] Cross-agent consistency check — matches standard pattern (react_agent, crewai/websearch_agent)
- [x] CI YAML valid — actionlint passes

### DIFF
--- a/.github/workflows/agent-deployment-test.yaml
+++ b/.github/workflows/agent-deployment-test.yaml
@@ -94,7 +94,7 @@ jobs:
         if: matrix.agent.name != 'autogen-mcp-agent' || vars.MCP_SERVER_URL != ''
         working-directory: ${{ matrix.agent.dir }}
         env:
-          API_KEY: ${{ matrix.API_KEY || vars.API_KEY }}
+          API_KEY: ${{ matrix.agent.name == 'vanilla-python-openai-responses-agent' && secrets.OPENAI_API_KEY || vars.API_KEY }}
           BASE_URL: ${{ matrix.BASE_URL || vars.BASE_URL }}
           MODEL_ID: ${{ matrix.MODEL_ID || vars.MODEL_ID }}
           POSTGRES_HOST: ${{ matrix.POSTGRES_HOST }}
--- a/.github/workflows/agent-deployment-test.yaml
+++ b/.github/workflows/agent-deployment-test.yaml
@@ -64,6 +64,7 @@ jobs:
           - { name: llamaindex-websearch-agent, dir: agents/llamaindex/templates/websearch_agent }
           - { name: langgraph-db-memory-agent, dir: agents/langgraph/templates/react_with_database_memory }
           - { name: autogen-mcp-agent, dir: agents/autogen/templates/mcp_agent }
+          - { name: vanilla-python-openai-responses-agent, dir: agents/vanilla_python/templates/openai_responses_agent }
         include:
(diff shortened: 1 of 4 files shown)


andrewdonheiser (commented):
**[Important] Silent fallback when `OPENAI_API_KEY` secret is not configured**

The `API_KEY` override on line 92:

```yaml
API_KEY: ${{ matrix.agent.name == 'vanilla-python-openai-responses-agent' && secrets.OPENAI_API_KEY || vars.API_KEY }}
```

If `secrets.OPENAI_API_KEY` is not configured, the expression evaluates as:
`true && '' || vars.API_KEY` → `vars.API_KEY`

The OpenAI agent silently receives the vLLM API key and fails with a confusing auth error against the OpenAI API, with no indication the secret is missing.

**Suggested fix:** Move `API_KEY` into the `include` block to match the `BASE_URL`/`MODEL_ID` pattern:

```yaml
include:
  - agent: { name: vanilla-python-openai-responses-agent, dir: ... }
    BASE_URL: ${{ vars.OPENAI_BASE_URL }}
    MODEL_ID: ${{ vars.OPENAI_MODEL_ID }}
    API_KEY: ${{ secrets.OPENAI_API_KEY }}
```

Then simplify the step env to:

```yaml
API_KEY: ${{ matrix.API_KEY || vars.API_KEY }}
```

This makes all three env vars consistent, scales to additional Tier 2 agents without extending a ternary, and causes an empty `API_KEY` to fail fast in `_write_env_file` rather than silently using the wrong credential.

mpk-droid:
Thanks for the review @andrewdonheiser — good catch on the silent fallback.

The suggested fix (putting `secrets.OPENAI_API_KEY` in `matrix.include`) won't work unfortunately — `secrets` context is not available in `strategy.matrix`, only `github`, `inputs`, `needs`, `vars`. We already hit this as an actionlint failure and moved it to the step-level conditional.

Instead, I've addressed it by making `_write_env_file` validate `API_KEY` as a **required** env var (unlike standard agents that default to `not-needed`). If `OPENAI_API_KEY` isn't configured in CI, the test now fails immediately with:

> `Missing required env vars: API_KEY. This agent requires a real OpenAI API key — set OPENAI_API_KEY in CI secrets or export API_KEY locally.`

This catches the misconfiguration at test time rather than letting it silently fall through to a confusing auth error.

Assisted by Claude opus-4-6

andrewdonheiser (approved):
👍

PR #237, opened 2026-07-06
feat(a2a): migrate langgraph_crewai_agent to a2a-sdk v1.x (RHAIENG-5373)

andrewdonheiser:
## Summary

- Migrate `a2a/langgraph_crewai_agent` from `a2a-sdk v0.3.x` (Pydantic) to `v1.1.0` (protobuf)
- Replace `A2AStarletteApplication` with `create_agent_card_routes` + `create_jsonrpc_routes` route factories
- Replace `AgentCard.url` with `supported_interfaces=[AgentInterface(url=..., protocol_binding="jsonrpc/http")]`
- Rewrite `a2a_reply.py` client from `A2AClient` to `create_client` + streaming `send_message`
- Replace `new_agent_text_message` with `new_text_message` across both server executors
- Add `DefaultRequestHandler(agent_card=...)` required 3rd arg
- Add smoke tests (`tests/test_imports.py`) verifying all v1.x imports and module loads
- Add `make test` target to Makefile

## Files changed

| File | Change |
|------|--------|
| `pyproject.toml` | `a2a-sdk >=0.3.25,<1.0.0` → `>=1.0.0,<2.0.0` |
| `crew_a2a_server.py` | Route factories, `new_text_message`, `AgentInterface`, `agent_card` arg |
| `langgraph_a2a_server.py` | Same pattern + plain-dict JSON-RPC envelope (no SDK types) |
| `a2a_reply.py` | Full rewrite: `create_client` + streaming + protobuf message construction |
| `demo_client.py` | Full rewrite: same streaming pattern |
| `tests/test_imports.py` | 8 smoke tests (SDK imports + module imports) |
| `Makefile` | Added `test` target |
| `uv.lock` | Updated lockfile |

## Context

Closes the gap from Dependabot PR #124 which was closed because code changes were required. The a2a-sdk v1.0.0 (2026-04-20) introduced breaking changes: protobuf types replace Pydantic, `A2AStarletteApplication` removed in favor of route factories, `AgentCard.url` replaced by `supported_interfaces`, and the client API switched to streaming-only.

## Test plan

- [x] All 4 module imports pass (`uv run python -c "import ..."`)
- [x] `ruff check src/ tests/` — clean
- [x] `ruff format --check src/ tests/` — clean
- [x] 8/8 smoke tests pass (`pytest tests/ -v`)
- [x] Deployed to OpenShift (`adonheis-testing` namespace) — both agents healthy
- [x] Agent cards serve `supportedInterfaces` with `protocolBinding: "jsonrpc/http"`
- [x] MLflow tracing enabled on both agents (`[Tracing Enabled]` in startup logs)


agents/a2a/templates/langgraph_crewai_agent/src/a2a_langgraph_crewai/a2a_reply.py
```diff
@@ -63,58 +29,51 @@ def _json_for_log(obj: Any) -> str:
...
+
+    msg = Message(
+        message_id=msg_id,
+        role=Role.ROLE_USER,
+        parts=[Part(text=text)],
+    )
+    req = SendMessageRequest(message=msg)
+    logger.debug("A2A -> request: %s", _json_for_log(req))
+
+    client = await create_client(base)
```
shricharan-ks: ### Major: Timeout removed without replacement

The old signature accepted `timeout: float = 120.0` and forwarded it to `httpx.AsyncClient(timeout=timeout)`. The new version removes it entirely — `create_client(base)` uses whatever default the SDK ships.

For inter-agent calls where the CrewAI peer runs a full crew kickoff (potentially slow LLM calls), the default timeout may be too short and cause spurious failures, or too long/infinite and cause pod resource exhaustion if the peer hangs.

**Suggestion:** Check whether `create_client` accepts a timeout or `ClientConfig` parameter. If so, pass an appropriate value. At minimum, consider wrapping with `asyncio.wait_for()` or documenting that the timeout behavior changed.

agents/a2a/templates/langgraph_crewai_agent/src/a2a_langgraph_crewai/a2a_reply.py
```diff
@@ -63,58 +29,51 @@ def _json_for_log(obj: Any) -> str:
...
+    try:
+        parts: list[str] = []
+        async for response in client.send_message(req):
```
shricharan-ks, line 63: ### Major: JSON-RPC errors silently swallowed in streaming + lost error logging

Two issues here:

**1. Silent error swallowing:** The old code explicitly checked for `JSONRPCErrorResponse` and raised `RuntimeError`. The new streaming code only extracts text via `get_stream_response_text(response)` and skips non-text chunks. If the peer returns a JSON-RPC error in the stream, `get_stream_response_text` returns `None`, the error is silently ignored, and the function returns an empty string. The caller in `langgraph_a2a_server.py` would pass that empty string back as a valid tool result.

**2. Lost error logging:** The old code had `except Exception: logger.exception(...)` before re-raising. The new code propagates exceptions without logging, losing inter-agent observability.

**Suggestion:** Add error logging before re-raise, and warn on empty responses:

```python
try:
    parts: list[str] = []
    async for response in client.send_message(req):
        ...
    out = "\n".join(parts) if parts else ""
    if not out:
        logger.warning("A2A <- peer=%s msg_id=%s returned empty response", base, msg_id)
    ...
except Exception:
    logger.exception("A2A <- peer=%s msg_id=%s stream failed", base, msg_id)
    raise
finally:
    await client.close()
```

agents/a2a/templates/langgraph_crewai_agent/src/a2a_langgraph_crewai/crew_a2a_server.py
```diff
@@ -141,25 +142,30 @@ def main() -> None:
     agent_card = AgentCard(
         name="CrewAI A2A Specialist",
         description="Minimal CrewAI agent speaking the A2A protocol.",
-        url=f"{public_base}/",
+        supported_interfaces=[
+            AgentInterface(
+                url=f"{public_base}/",
+                protocol_binding="jsonrpc/http",
```
shricharan-ks: ### Minor: `protocol_binding` value — verify against spec

The A2A v1.0 migration guide references `"JSONRPC"`, `"HTTP+JSON"`, and `"GRPC"` as the canonical protocol binding values. This uses `"jsonrpc/http"` which is non-standard.

Since the field is a free-form string in protobuf and this was deployed successfully, the SDK clearly accepts it. But it could cause interoperability issues with other A2A implementations that inspect `protocolBinding` to decide how to communicate.

Worth double-checking against the [A2A spec](https://github.com/a2aproject/A2A/blob/main/docs/specification.md) and the [a2a-python samples](https://github.com/a2aproject/a2a-python/tree/main/examples). Same applies to `langgraph_a2a_server.py`.

agents/a2a/templates/langgraph_crewai_agent/tests/test_imports.py
```diff
@@ -0,0 +1,67 @@
...
+            get_artifact_text,
+            get_message_text,
+            get_stream_response_text,
+            new_text_message,
+        ]
+    )
+
+
+def test_client_imports():
+    from a2a.client import A2ACardResolver, create_client
```
shricharan-ks: ### Minor: `A2ACardResolver` imported but not used by agent code

The agent code no longer uses `A2ACardResolver` directly (it was removed in the migration — `create_client` handles card resolution internally). This test verifies it still exists in the SDK, which is harmless but slightly misleading about what the agent actually uses.

Nit: consider removing `A2ACardResolver` from this test to keep it focused on what the agent imports.

agents/a2a/templates/langgraph_crewai_agent/tests/test_imports.py
```diff
@@ -0,0 +1,67 @@
...
+
+    assert all([A2ACardResolver, create_client])
+
+
+def test_module_crew_a2a_server():
```
shricharan-ks, line 54: ### Minor: Module import smoke tests trigger side effects; consider behavioral tests

Importing `crew_a2a_server` and `langgraph_a2a_server` executes `load_dotenv()` and `logging.basicConfig()` at module level, which reconfigures the root logger for the entire test session. In CI this is harmless (`.env` doesn't exist), but it could mask log output issues in local test runs.

Also: the smoke tests are valuable for a migration but there are no behavioral tests for the changed code — `send_a2a_text_message` with a mocked client, `_jsonrpc_message_send_envelope` wire format assertions, etc. Even simple dict assertions on the envelope functions would catch wire-format regressions. (Not a blocker, just a suggestion for follow-up.)

shricharan-ks (commented):
## PR Review: Migrate `langgraph_crewai_agent` to `a2a-sdk` v1.x

Well-executed SDK migration. The author clearly followed the [official v0.3 to v1.0 migration guide](https://github.com/a2aproject/a2a-python/blob/main/docs/migrations/v1_0/README.md) and applied changes consistently across all four source files. CI is green, tested on OpenShift, and the PR description is thorough.

### Summary

| Category | Count |
|----------|-------|
| Critical | 0 |
| Major | 2 |
| Minor | 5 |
| Positive | 8 |

### Positive Observations

1. **Consistent migration** — both `crew_a2a_server.py` and `langgraph_a2a_server.py` follow the exact same pattern (route factories, `new_text_message`, `AgentInterface`, `agent_card` arg to `DefaultRequestHandler`)
2. **Correct v1.x Part format** — `_jsonrpc_message_send_envelope` correctly removed `"kind"` from parts (v0.x `{"kind": "text", "text": ...}` to v1.x `{"text": ...}`)
3. **Proper protobuf serialization in logging** — `_json_for_log` correctly handles protobuf `Message` via `MessageToDict` with `preserving_proto_field_name=True`
4. **Proper resource cleanup** — both `a2a_reply.py` and `demo_client.py` use `try/finally: await client.close()`
5. **Correct enum usage** — `Role.ROLE_USER` matches v1.x SCREAMING_SNAKE_CASE convention
6. **`supports_authenticated_extended_card` correctly removed** — this field moved to `AgentCapabilities.extended_agent_card` in v1.x
7. **`make test` target is CI-compatible** — fix commit correctly removed `.venv` guard and added `--ignore` flags
8. **Lockfile updated** — `uv.lock` correctly reflects `a2a-sdk 1.1.0` with new dependencies

### Verdict

**Approve with comments.** The two major findings (timeout removal, silent error swallowing) are worth addressing but aren't blockers for a demo/template agent. The migration itself is correct and thorough.

See inline comments for specific findings.

shricharan-ks (dismissed):
Approving — the a2a-sdk v1.x migration is correct, consistent, and thoroughly tested (CI green, deployed on OpenShift). The inline comments on timeout and error handling are suggestions for hardening, not blockers.

kami619 (approved):

shricharan-ks (approved):
## Round 2 Review — All findings addressed

The fix commit (`9bdfd999`) cleanly resolves all 7 findings from the first review:

| Finding | Status |
|---------|--------|
| **Major: Timeout removed** | Fixed — 120s timeout restored via `ClientConfig(httpx_client=httpx.AsyncClient(timeout=120.0))` |
| **Major: Silent error swallowing** | Fixed — `except Exception: logger.exception(...)` added before re-raise, plus `logger.warning` on empty responses |
| **Minor: `protocol_binding` value** | Fixed — changed from `"jsonrpc/http"` to `"JSONRPC"` per A2A spec, in both servers |
| **Minor: Module-level side effects** | Fixed — `load_dotenv()` and `logging.basicConfig()` moved from module scope into `main()` in both servers |
| **Minor: Unused `A2ACardResolver` in test** | Fixed — replaced with `ClientConfig` which the code now actually uses |
| **Minor: No behavioral tests** | Acknowledged as follow-up (acceptable) |
| **Minor: `google.protobuf` not explicit** | Not addressed but non-blocking (transitive dep is stable) |

CI is green. The changes are minimal and focused. LGTM.

PR #238, opened 2026-07-06
feat: add OpenCode A2A agent deployment for Kagenti discovery

msager27:
## Description

Integrates the OpenCode A2A agent card server into the OpenCode container image on RHOAI, enabling OpenCode agents to be discovered in the Kagenti UI catalog.

New files:
- Containerfile.a2a - Extends base OpenCode image with OpenCode A2A server
- entrypoint-a2a.sh - Translates Kagenti-standard env vars (LLM_*) to OpenCode config
- kagenti-agent.yaml - OpenShift Template with Kagenti discovery labels
- README-a2a.md - Deployment and troubleshooting guide

Key features:
- Agent discovery via Kagenti UI (A2A protocol)
- Standardized LLM_API_BASE, LLM_MODEL, LLM_API_KEY environment variables
- Multi-provider support (vLLM, OGX, OpenAI-compatible endpoints)
- Provider config generated at container startup from env vars (no image rebuilds)

## Jira Ticket

RHAIENG-5825

## Testing

-[ ] AgentCard created and shows Synced=True
-[ ] Agent card endpoint returns valid JSON
-[ ] OpenCode config contains correct LLM provider settings from env vars
-[ ] OpenCode TUI accessible via oc exec and processes requests successfully
-[ ] Agent appears in Kagenti UI catalog
-[ ] LLM requests route correctly through configured provider

### DIFF
--- a/agents/opencode/deployment/Containerfile.a2a
+++ b/agents/opencode/deployment/Containerfile.a2a
@@ -1,0 +1,30 @@
+# OpenCode with A2A agent card support
+#
+# Extends the pre-built OpenCode image with:
+# - opencode-a2a server for Kagenti discovery (A2A protocol)
+# - Custom entrypoint running both opencode serve and opencode-a2a
+#
+# Build:
+#   podman build --platform linux/amd64 -t opencode-a2a:latest -f Containerfile.a2a .
+#
+# The A2A agent card is served at:
+#   GET /.well-known/agent-card.json (port 8000)
+
+FROM quay.io/opendatahub/odh-opencode-rhel9:latest
+
+USER 0
+
+# Install opencode-a2a from agent-servers repo
+RUN pip3 install --no-cache-dir \
+    'opencode-a2a @ git+https://github.com/red-hat-data-services/agent-servers.git#subdirectory=opencode'
+
+# Copy entrypoint that runs both opencode serve and opencode-a2a
+COPY entrypoint-a2a.sh /usr/local/bin/entrypoint-a2a.sh
+RUN chmod 755 /usr/local/bin/entrypoint-a2a.sh
+
+USER 1001
+
+# A2A server port
+EXPOSE 8000
+
+ENTRYPOINT ["/usr/local/bin/entrypoint-a2a.sh"]
--- a/agents/opencode/deployment/Containerfile.a2a
+++ b/agents/opencode/deployment/Containerfile.a2a
@@ -15,6 +15,8 @@ FROM quay.io/opendatahub/odh-opencode-rhel9:latest
 USER 0
 
 # Install opencode-a2a from agent-servers repo
+# TODO(RHAIENG-5826): Pin to specific version tag once A2A protocol support is complete
+# and agent-servers repo has established versioning
 RUN pip3 install --no-cache-dir \
     'opencode-a2a @ git+https://github.com/red-hat-data-services/agent-servers.git#subdirectory=opencode'
 
--- a/agents/opencode/deployment/README-a2a.md
+++ b/agents/opencode/deployment/README-a2a.md
@@ -216,9 +216,12 @@ helm upgrade kagenti /path/to/kagenti/chart \
 
 ## Verification
 
-### Check AgentCard Creation
+### Check AgentRuntime and AgentCard Creation
 
 ```bash
+# Check AgentRuntime was created
+oc get agentruntime -n your-namespace
+
 # List AgentCards in your namespace
 oc get agentcard -n your-namespace
 
@@ -226,7 +229,14 @@ oc get agentcard -n your-namespace
 oc get agentcard opencode-a2a-deployment-card -n your-namespace -o yaml
 ```
 
-Expected output shows `Synced=True`:
+Expected AgentRuntime output shows `Phase=Active`:
+
+```text
+NAME                       TYPE    TARGET          PHASE
+opencode-a2a-runtime       agent   opencode-a2a    Active
+```
+
+Expected AgentCard output shows `Synced=True`:
 
 ```text
 NAME                          PROTOCOL   KIND         TARGET        AGENT      VERIFIED   BOUND   SYNCED   LASTSYNC
@@ -438,7 +448,7 @@ Verify LLM_API_BASE is correct:
 
 ```bash
 oc exec -n your-namespace deployment/opencode-a2a -- \
-  curl -s $LLM_API_BASE/models
+  sh -c 'curl -s $LLM_API_BASE/models'
 ```
 
 Check OpenCode config was generated:
@@ -478,6 +488,6 @@ oc logs -n your-namespace deployment/opencode-a2a --previous
 
 Common issues:
 
-- Missing LLM_API_BASE: entrypoint will skip config generation
+- Missing LLM_API_BASE: Pod fails to start with validation error (required when LLM_MODEL is set)
 - Invalid LLM_API_BASE: OpenCode serve starts but can't reach LLM
 - Missing secret: Pod fails to start if `LLM_API_KEY_SECRET` doesn't exist
--- a/agents/opencode/deployment/README-a2a.md
+++ b/agents/opencode/deployment/README-a2a.md
@@ -1,0 +1,483 @@
+# OpenCode A2A Agent Deployment
+
+This directory contains a reference implementation for deploying [OpenCode](https://opencode.ai) as a Kagenti-discoverable agent using the Agent-to-Agent (A2A) protocol.
+
+## Overview
+
+OpenCode is an AI-powered coding assistant. This implementation wraps OpenCode with an A2A agent card server to enable:
+
+- **Service discovery** via Kagenti agent catalog
+- **Standardized configuration** using Kagenti-standard environment variables
+- **Multi-provider support** (direct vLLM, OGX, OpenAI, etc.)
+
+## Architecture
+
+```text
+┌──────────────────────────────────────────────┐
+│  Container (port 8000)                       │
+│                                              │
+│  ┌───────────────────────────────────────┐   │
+│  │ opencode-a2a (FastAPI)                │   │
+│  │ - Serves /.well-known/agent-card.json │   │
+│  │ - Proxies /health from opencode serve │   │
+│  └──────────────┬────────────────────────┘   │
+│                 │ http://localhost:4096      │
+│  ┌──────────────▼────────────────────────┐   │
+│  │ opencode serve (port 4096)            │   │
+│  │ - OpenCode HTTP API                   │   │
+│  │ - Configured via ~/.config/opencode/  │   │
+│  └──────────────┬────────────────────────┘   │
+│                 │                            │
+└─────────────────┼────────────────────────────┘
+                  │
+                  ▼
+          LLM_API_BASE (vLLM, OGX, etc.)
+```
+
+### Components
+
+1. **Base Image**: `quay.io/opendatahub/odh-opencode-rhel9:latest`
+   - Pre-built OpenCode container from Red Hat OpenShift AI
+
+2. **opencode-a2a Server** (`Containerfile.a2a`)
+   - Installed from `agent-servers` repo (opencode subdirectory)
+   - Serves A2A agent card for Kagenti discovery
+   - Proxies health checks from OpenCode
+
+3. **Entrypoint Script** (`entrypoint-a2a.sh`)
+   - Translates Kagenti-standard env vars to OpenCode config
+   - Starts both `opencode serve` and `opencode-a2a` processes
+
+4. **Kubernetes Template** (`kagenti-agent.yaml`)
+   - OpenShift Template for deployment
+   - Configures Kagenti discovery labels
+   - Supports multiple deployment variants (direct vLLM, OGX, etc.)
+
+## Environment Variables
+
+The deployment follows [Kagenti standard naming conventions](https://github.com/kagenti):
+
+| Variable           | Description                                              | Required | Default    |
+|--------------------|----------------------------------------------------------|----------|------------|
+| `LLM_PROVIDER`     | Provider name (vllm, ogx, openai, etc.)                  | No       | `vllm`     |
+| `LLM_API_BASE`     | LLM API endpoint (e.g., `https://api.openai.com/v1`)     | Yes      | -          |
+| `LLM_MODEL`        | Model identifier (e.g., `gpt-4o`, `gpt-oss-120b`)        | Yes      | -          |
+| `LLM_API_KEY`      | API key for the LLM provider                             | No       | -          |
+| `A2A_PORT`         | Port for A2A server                                      | No       | `8000`     |
+| `A2A_PROVIDER_ORG` | Organization name in agent card                          | No       | `Red Hat`  |
+
+### Provider-Specific Model Naming
+
+- **Direct vLLM**: Use model name only (e.g., `LLM_MODEL=gpt-oss-120b`)
+- **Through OGX**: Use provider/model format (e.g., `LLM_MODEL=vllm/gpt-oss-120b`)
+- **OpenAI**: Use OpenAI model names (e.g., `LLM_MODEL=gpt-4o`)
+
+## Deployment
+
+### Prerequisites
+
+- OpenShift/Kubernetes cluster
+- Kagenti installed and configured (for agent discovery)
+- LLM endpoint accessible from the cluster
+- Secret containing LLM API key (if required)
+
+### Build Container Image
+
+Using OpenShift BuildConfig (recommended for OpenShift deployments):
+
+```bash
+cd agents/opencode/deployment
+
+# Create ImageStream to store the built image
+oc create imagestream opencode-a2a -n your-namespace
+
+# Create BuildConfig for binary builds
+cat <<EOF | oc apply -f -
+apiVersion: build.openshift.io/v1
+kind: BuildConfig
+metadata:
+  name: opencode-a2a
+  namespace: your-namespace
+spec:
+  output:
+    to:
+      kind: ImageStreamTag
+      name: opencode-a2a:latest
+  source:
+    type: Binary
+  strategy:
+    dockerStrategy:
+      dockerfilePath: Containerfile.a2a
+    type: Docker
+EOF
+
+# Trigger build from current directory
+oc start-build opencode-a2a --from-dir=. --follow -n your-namespace
+
+# The image will be available at:
+# image-registry.openshift-image-registry.svc:5000/your-namespace/opencode-a2a:latest
+```
+
+Alternative: Using podman/docker (for testing or non-OpenShift environments):
+
+```bash
+cd agents/opencode/deployment
+
+# Build for amd64
+podman build --platform linux/amd64 \
+  -t opencode-a2a:latest \
+  -f Containerfile.a2a .
+
+# Tag for your registry
+podman tag opencode-a2a:latest \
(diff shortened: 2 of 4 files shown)


agents/opencode/deployment/kagenti-agent.yaml
```diff
@@ -0,0 +1,151 @@
...
+        kagenti.io/type: agent
+        protocol.kagenti.io/a2a: ""
+    spec:
+      type: ClusterIP
+      selector:
+        app.kubernetes.io/name: opencode-a2a
+        kagenti.io/type: agent
+      ports:
+        - name: a2a
+          port: 8000
```
shricharan-ks, line 76: ### Critical: `protocol.kagenti.io/a2a` label value inconsistent with existing agents

The existing A2A agents in this repo use:
```yaml
protocol.kagenti.io/a2a: "true"
```

But this template uses an empty string:
```yaml
protocol.kagenti.io/a2a: ""
```

This appears in the Service (line 63), Deployment metadata (line 80), and pod template (line 95). If Kagenti's discovery controller matches on `protocol.kagenti.io/a2a: "true"` specifically (rather than just label presence), this agent will **not be discovered**.

Please align with the existing convention:
```yaml
protocol.kagenti.io/a2a: "true"
```

Ref: `agents/a2a/deployment/templates/deployment-crew.yaml` and `deployment-langgraph.yaml` both use `"true"`.

agents/opencode/deployment/kagenti-agent.yaml
```diff
@@ -0,0 +1,151 @@
...
+              securityContext:
+                allowPrivilegeEscalation: false
+                capabilities:
+                  drop:
+                    - ALL
+                runAsNonRoot: true
+                seccompProfile:
+                  type: RuntimeDefault
+          terminationGracePeriodSeconds: 30
+          restartPolicy: Always
```
shricharan-ks, line 165: ### Major: Missing `AgentRuntime` CRD resource

The existing A2A agent in this repo (`agents/a2a/templates/langgraph_crewai_agent/AgentRuntime.yaml`) defines `AgentRuntime` custom resources that link to the Deployment:

```yaml
apiVersion: agent.kagenti.dev/v1alpha1
kind: AgentRuntime
metadata:
  name: a2a-langgraph-agent-runtime
spec:
  type: agent
  protocol: a2a
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: a2a-langgraph-agent
```

This template doesn't create an `AgentRuntime` resource. Depending on whether Kagenti discovers agents via labels alone or requires the CRD, this could mean the agent won't appear in the catalog.

Consider adding an `AgentRuntime` object to the template's `objects` list.

agents/opencode/deployment/entrypoint-a2a.sh
```diff
@@ -0,0 +1,88 @@
...
+    cat > "${CONFIG_DIR}/opencode.json" <<EOF
+{
+  "\$schema": "https://opencode.ai/config.json",
+  "provider": {
+    "${PROVIDER_NAME}": {
+      "npm": "@ai-sdk/openai-compatible",
+      "name": "${PROVIDER_NAME}",
+      "options": {
+        "baseURL": "${BASE_URL}",
+        "apiKey": "${API_KEY}"
```
shricharan-ks, line 53: ### Major: Config file with API key written with default permissions

The generated `opencode.json` contains `LLM_API_KEY` in plaintext but is written with default umask (typically 0644, world-readable). In a multi-tenant cluster, another process in the same pod or a volume mount could expose this.

**Suggestion:** Restrict permissions before writing:
```bash
(umask 077 && cat > "${CONFIG_DIR}/opencode.json" <<EOF
...
EOF
)
```

This ensures the file is created with 0600 (owner-only read/write).

agents/opencode/deployment/entrypoint-a2a.sh
```diff
@@ -0,0 +1,88 @@
...
+    echo "  Config written to ${CONFIG_DIR}/opencode.json"
+fi
+
+# Start opencode serve in background
+echo "Starting opencode serve on port 4096..."
+opencode serve &
+OPENCODE_PID=$!
+
+# Wait for opencode serve to be ready
+echo "Waiting for opencode serve to be ready..."
```
shricharan-ks, line 92: ### Major: Background `opencode serve` process not cleaned up on exit

If `opencode-a2a` (the foreground `exec` process) exits or is killed, the background `opencode serve` process (PID `$OPENCODE_PID`) is orphaned. The `exec` on line 86 replaces the shell, so any trap set earlier wouldn't fire.

**Suggestion:** Add a trap before the background process starts:
```bash
opencode serve &
OPENCODE_PID=$!
trap "kill $OPENCODE_PID 2>/dev/null; wait $OPENCODE_PID 2>/dev/null" EXIT INT TERM
```

Note: since `exec` replaces the shell process, the trap won't fire on normal shutdown. Consider using a process supervisor (like `tini` or a simple wrapper) or running `opencode-a2a` without `exec` so the trap remains active. Alternatively, since containers have PID 1 signal forwarding behavior, the background process will be killed when the container stops — document this assumption if relying on it.

agents/opencode/deployment/kagenti-agent.yaml
```diff
@@ -0,0 +1,151 @@
...
+                      key: api-key
+                      optional: true
+                - name: LLM_MODEL
+                  value: ${LLM_MODEL}
+              readinessProbe:
+                tcpSocket:
+                  port: 8000
+                initialDelaySeconds: 5
+                periodSeconds: 5
+                failureThreshold: 12
```
shricharan-ks, line 140: ### Major: Readiness probe uses TCP check instead of HTTP health

The readiness probe only checks if port 8000 is accepting TCP connections:
```yaml
readinessProbe:
  tcpSocket:
    port: 8000
```

But the entrypoint waits for `opencode serve` health at `http://127.0.0.1:4096/health` before starting `opencode-a2a`, and the A2A server proxies health from OpenCode. Using an HTTP health check would verify the full stack is ready, not just that the port is bound:

```yaml
readinessProbe:
  httpGet:
    path: /.well-known/agent-card.json
    port: 8000
  initialDelaySeconds: 10
  periodSeconds: 5
  failureThreshold: 12
```

This also validates the A2A agent card is being served correctly.

agents/opencode/deployment/Containerfile.a2a
```diff
@@ -0,0 +1,30 @@
...
+# The A2A agent card is served at:
+#   GET /.well-known/agent-card.json (port 8000)
+
+FROM quay.io/opendatahub/odh-opencode-rhel9:latest
+
+USER 0
+
+# Install opencode-a2a from agent-servers repo
+RUN pip3 install --no-cache-dir \
+    'opencode-a2a @ git+https://github.com/red-hat-data-services/agent-servers.git#subdirectory=opencode'
```
shricharan-ks, line 21: ### Minor: `pip install` from git with no version pin

The install uses a git reference with no tag/commit pin:
```dockerfile
RUN pip3 install --no-cache-dir \
    'opencode-a2a @ git+https://github.com/red-hat-data-services/agent-servers.git#subdirectory=opencode'
```

This means every build could pull a different version, leading to non-reproducible images. Consider pinning to a specific commit or tag:
```dockerfile
RUN pip3 install --no-cache-dir \
    'opencode-a2a @ git+https://github.com/red-hat-data-services/agent-servers.git@v1.0.0#subdirectory=opencode'
```

agents/opencode/deployment/entrypoint-a2a.sh
```diff
@@ -0,0 +1,88 @@
...
+    mkdir -p "${CONFIG_DIR}"
+
+    # Set defaults if not provided
+    PROVIDER_NAME="${LLM_PROVIDER:-vllm}"
+    BASE_URL="${LLM_API_BASE:-}"
+    API_KEY="${LLM_API_KEY:-}"
+    MODEL_NAME="${LLM_MODEL:-gpt-4o}"
+
+    echo "  LLM_PROVIDER=${PROVIDER_NAME}"
+    echo "  LLM_API_BASE=${BASE_URL}"
```
shricharan-ks: ### Minor: Sensitive env vars echoed to stdout

Lines 33-36 echo `LLM_API_BASE`, `LLM_MODEL`, and `LLM_PROVIDER` to stdout. While the API key is masked (`****`), the `LLM_API_BASE` URL can contain sensitive information (internal service URLs, cluster topology). These log lines will appear in `oc logs` output.

Consider gating verbose output behind a `DEBUG` flag or reducing to a single line:
```bash
echo "  Configured provider=${PROVIDER_NAME} model=${MODEL_NAME}"
```

agents/opencode/deployment/kagenti-agent.yaml
```diff
@@ -0,0 +1,151 @@
...
+        protocol.kagenti.io/a2a: ""
+    spec:
+      replicas: 1
+      selector:
+        matchLabels:
+          app.kubernetes.io/name: opencode-a2a
+          kagenti.io/type: agent
+      template:
+        metadata:
+          labels:
```
shricharan-ks, line 94: ### Minor: `kagenti.io/type: agent` and `kagenti.io/framework: opencode` labels not used by existing agents

These labels (`kagenti.io/type`, `kagenti.io/framework`, `kagenti.io/workload-type`) are new to this PR and don't appear in the existing A2A agents (`deployment-crew.yaml`, `deployment-langgraph.yaml`). This isn't necessarily wrong — it could be a newer Kagenti convention — but it's worth confirming these labels are recognized by the Kagenti version deployed in the target clusters.

If these are the new standard, the existing A2A agents should be updated to match in a follow-up.

agents/opencode/deployment/entrypoint-a2a.sh
```diff
@@ -0,0 +1,88 @@
...
+#   LLM_MODEL     - Model identifier (e.g., gpt-4o)
+
+set -e
+
+echo "Starting OpenCode A2A container..."
+
+# Translate Kagenti-standard LLM_* env vars to OpenCode configuration
+# Kagenti standard: LLM_API_BASE, LLM_API_KEY, LLM_MODEL
+# OpenCode needs: Full provider configuration in opencode.json
+if [[ -n "${LLM_API_BASE}" || -n "${LLM_MODEL}" ]]; then
```
shricharan-ks, line 20: ### Minor: Config only generated if `LLM_API_BASE` or `LLM_MODEL` is set, but `LLM_API_BASE` has no default

The condition on line 20 requires either `LLM_API_BASE` or `LLM_MODEL` to be set, but `LLM_API_BASE` defaults to empty string (line 28: `BASE_URL="${LLM_API_BASE:-}"`). If only `LLM_MODEL` is set without `LLM_API_BASE`, the generated config will have an empty `baseURL`, causing OpenCode to fail silently at runtime.

Consider validating that `LLM_API_BASE` is non-empty when config generation runs:
```bash
if [[ -z "${LLM_API_BASE}" ]]; then
    echo "ERROR: LLM_API_BASE is required when LLM_MODEL is set"
    exit 1
fi
```

shricharan-ks (changes requested):
## PR Review: Add OpenCode A2A agent deployment for Kagenti discovery

Good first integration of OpenCode into the Kagenti discovery framework. The architecture is clean (two-process container with entrypoint translation), the README is thorough, and the OpenShift Template approach is appropriate for this use case. However, there are several issues — one critical Kagenti discovery inconsistency, a few security/reliability concerns, and a missing `AgentRuntime` CRD — that should be addressed before merging.

### Summary

| Category | Count |
|----------|-------|
| Critical | 1 |
| Major | 4 |
| Minor | 4 |
| Positive | 5 |

### Positive Observations

1. **Clean architecture** — two-process model with `opencode serve` + `opencode-a2a` is well-designed, with proper health-check gating before starting the A2A server
2. **Thorough documentation** — `README-a2a.md` covers deployment, verification, troubleshooting, and known limitations with architecture diagrams
3. **Good security posture** — container runs as non-root (UID 1001), drops all capabilities, uses `seccompProfile: RuntimeDefault`, and `allowPrivilegeEscalation: false`
4. **Honest about limitations** — clearly documents that A2A task execution (RHAIENG-5826) is not yet implemented
5. **Kagenti env var convention** — uses `LLM_*` standardized naming for cross-agent consistency

See inline comments for detailed findings.

agents/opencode/deployment/entrypoint-a2a.sh
```diff
@@ -0,0 +1,88 @@
...
+        }
+      }
+    }
+  },
+  "model": "${PROVIDER_NAME}/${MODEL_NAME}",
+  "small_model": "${PROVIDER_NAME}/${MODEL_NAME}",
+  "enabled_providers": ["${PROVIDER_NAME}"]
+}
+EOF
+    echo "  Config written to ${CONFIG_DIR}/opencode.json"
```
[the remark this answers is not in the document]
msager27: Done

agents/opencode/deployment/kagenti-agent.yaml
```diff
@@ -0,0 +1,151 @@
...
+                      optional: true
+                - name: LLM_MODEL
+                  value: ${LLM_MODEL}
+              readinessProbe:
+                tcpSocket:
+                  port: 8000
+                initialDelaySeconds: 5
+                periodSeconds: 5
+                failureThreshold: 12
+                timeoutSeconds: 1
```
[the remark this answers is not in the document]
msager27: Done

agents/opencode/deployment/README-a2a.md
```diff
@@ -0,0 +1,483 @@
...
+```
+
+### OpenCode can't reach LLM
+
+Verify LLM_API_BASE is correct:
+
+```bash
+oc exec -n your-namespace deployment/opencode-a2a -- \
+  curl -s $LLM_API_BASE/models
+```
```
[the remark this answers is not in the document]
msager27: Done

agents/opencode/deployment/kagenti-agent.yaml
msager27: Done

agents/opencode/deployment/kagenti-agent.yaml
msager27: Done

agents/opencode/deployment/entrypoint-a2a.sh
msager27: Done

agents/opencode/deployment/entrypoint-a2a.sh
msager27: Done

agents/opencode/deployment/kagenti-agent.yaml
msager27: Done

agents/opencode/deployment/Containerfile.a2a
msager27: A2A work is still under active development.  Will pin to a specific version once versioning is established and A2A work stabilizes.  Added a TODO in the Containerfile to circle back to this.

agents/opencode/deployment/kagenti-agent.yaml
msager27: Done

agents/opencode/deployment/entrypoint-a2a.sh
msager27: Done

agents/opencode/deployment/entrypoint-a2a.sh
msager27: Done

shricharan-ks (dismissed):
## Round 2 Review — All findings addressed

The fix commit (`1d6b436a`) resolves all 9 findings from the first review:

| Finding | Status |
|---------|--------|
| **Critical: `protocol.kagenti.io/a2a: ""`** | Fixed — changed to `"true"` on Service, Deployment, and pod template. Non-standard `kagenti.io/type` labels removed. Matches existing A2A agents |
| **Major: Missing AgentRuntime CRD** | Fixed — `AgentRuntime` object added to template with correct `targetRef` pointing to the Deployment |
| **Major: Config file world-readable** | Fixed — `(umask 077 && cat > ...)` ensures 0600 permissions |
| **Major: Background process not cleaned up** | Fixed — trap added for EXIT/INT/TERM, `exec` replaced with plain invocation so trap remains active |
| **Major: TCP readiness probe** | Fixed — changed to `httpGet` on `/.well-known/agent-card.json` port 8000, `initialDelaySeconds` bumped to 10 |
| **Minor: Unpinned git install** | Acknowledged with TODO comment tracking RHAIENG-5826 |
| **Minor: Verbose env logging** | Fixed — reduced to single line `Provider: ${PROVIDER_NAME}, Model: ${MODEL_NAME}` |
| **Minor: Non-standard Kagenti labels** | Fixed — `kagenti.io/type`, `kagenti.io/framework`, `kagenti.io/workload-type` removed |
| **Minor: Missing LLM_API_BASE validation** | Fixed — explicit validation with clear error message and `exit 1` |

Additional improvements in the fix:
- README updated with `AgentRuntime` verification steps
- Fixed `curl` command in troubleshooting (now runs inside pod shell: `sh -c 'curl -s $LLM_API_BASE/models'`)
- Default IMAGE changed from external `quay.io` to internal registry path
- Troubleshooting section updated to reflect new validation behavior

CI green. LGTM.

aakankshaduggal (commented):
### What's still open

**1. Trap masks failure exit codes** (entrypoint-a2a.sh:78)
```bash
trap "echo 'Shutting down...'; kill $OPENCODE_PID 2>/dev/null; wait $OPENCODE_PID 2>/dev/null; exit 0" EXIT INT TERM
```
The hard-coded `exit 0` means if `opencode-a2a` crashes, the container reports success. Kubernetes won't apply restart backoff properly — it thinks the container exited cleanly. Was this intentional (e.g., to avoid crash loops during early development), or should it propagate the real exit code?

**2. No liveness probe** (kagenti-agent.yaml)
There's a readiness probe on `/.well-known/agent-card.json`, but no liveness probe. If the A2A server or opencode serve hangs, the pod stays running and accepting traffic it can't serve. Was the liveness probe intentionally omitted, or just not added yet? A liveness probe on the same path with a longer interval would catch hung processes.

**3. JSON values not escaped in heredoc** (entrypoint-a2a.sh:44-61)
`BASE_URL`, `API_KEY`, `MODEL_NAME`, `PROVIDER_NAME` are interpolated raw into JSON. A value containing `"` or `\` would produce invalid config. Low probability for typical env vars, but CodeRabbit flagged this and it wasn't addressed.

### What looks good

- Clean two-process architecture with health-gate before starting A2A
- Security posture is solid: non-root, drop ALL caps, seccomp, umask 077 on secrets
- Honest about limitations — clearly documents what works and what doesn't
- Follows Kagenti env var conventions for cross-agent consistency
- AgentRuntime CRD included in the template after Shricharan's review

### Verdict

Items 1-2 are worth fixing before merge unless they were deliberate choices — the exit code masking in particular could cause real operational issues in production. Item 3 is a nice-to-have. Otherwise in good shape after Shricharan's review pass.

msager27:
Thanks for the reviews.  The latest commit addresses the trap exit code issue and the liveness probe.  The un-escaped json values issue (#3 above) is low-risk, so I deferred it.  Please re-review.

aakankshaduggal (approved):
Thanks, LGTM!

shricharan-ks (approved):
/LGTM

Nehanth (approved):

PR #244, opened 2026-07-08
fix: fix /workspace/projects permissions for fresh PVC deployments

msager27:
## Description

- Remove WORKDIR directive from Containerfile (causes root-owned directory creation)
- Use install -d -m 2775 in entrypoint for atomic permission setting
- Add cd /workspace/projects in entrypoint for main process
- Add cd to .bashrc for interactive shells
- Update README to remove WORKDIR reference"

## Testing

- [ ] Create a fresh deployment and run claude tests as before (see README for examples)
- [ ] oc exec into pod and verify that /workspace/projects directory is writable

### DIFF
--- a/agents/claude-code/deployment/entrypoint.sh
+++ b/agents/claude-code/deployment/entrypoint.sh
@@ -159,14 +159,12 @@ setup_config_dir() {
     # Ensure the config directory exists and is group-writable.
     # On OpenShift, fresh PVCs are owned by root with the pod's fsGroup.
     # The non-root container user writes via group membership, so directories
-    # must be group-writable (g+w).
-    mkdir -p "${CLAUDE_CONFIG_DIR}"
-    chmod g+w "${CLAUDE_CONFIG_DIR}" 2>/dev/null || true
+    # must be group-writable (2775 = rwxrwsr-x).
+    install -d -m 2775 "${CLAUDE_CONFIG_DIR}" 2>/dev/null || mkdir -p "${CLAUDE_CONFIG_DIR}"
 
-    # Ensure the projects directory exists (WORKDIR is /workspace/projects)
+    # Ensure the projects directory exists
     # This separates global config (/workspace/.claude) from local auto-memory (/workspace/projects/.claude)
-    mkdir -p /workspace/projects
-    chmod g+w /workspace/projects 2>/dev/null || true
+    install -d -m 2775 /workspace/projects 2>/dev/null || mkdir -p /workspace/projects
 
     # Create symlink from ~/.claude to the config dir for user convenience
     # Users expect to find settings/skills at ~/.claude/
@@ -412,6 +410,11 @@ WRAPPER
         echo 'export PATH="${HOME}/.claude:${PATH}"' >> "${HOME}/.bashrc"
     fi
 
+    # Set working directory for interactive shells
+    if ! grep -q 'cd /workspace/projects' "${HOME}/.bashrc" 2>/dev/null; then
+        echo 'cd /workspace/projects 2>/dev/null || true' >> "${HOME}/.bashrc"
+    fi
+
     log_info "Claude args persisted to ${claude_dir}/env.sh"
     log_info "Wrapper script available: claude-run (or ~/.claude/claude-run)"
 }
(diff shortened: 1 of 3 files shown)


agents/claude-code/deployment/entrypoint.sh
```diff
@@ -159,14 +159,12 @@ setup_config_dir() {
     # Ensure the config directory exists and is group-writable.
     # On OpenShift, fresh PVCs are owned by root with the pod's fsGroup.
     # The non-root container user writes via group membership, so directories
-    # must be group-writable (g+w).
-    mkdir -p "${CLAUDE_CONFIG_DIR}"
-    chmod g+w "${CLAUDE_CONFIG_DIR}" 2>/dev/null || true
+    # must be group-writable (2775 = rwxrwsr-x).
+    install -d -m 2775 "${CLAUDE_CONFIG_DIR}" 2>/dev/null || { mkdir -p "${CLAUDE_CONFIG_DIR}" && chmod 2775 "${CLAUDE_CONFIG_DIR}"; }
```
hmoghani: If `install` fails and the fallback's `chmod` also fails, the directory ends up with wrong permissions and no warning. Adding a trailing `log_warn` catches that case.

```suggestion
    install -d -m 2775 "${CLAUDE_CONFIG_DIR}" 2>/dev/null || { mkdir -p "${CLAUDE_CONFIG_DIR}" && chmod 2775 "${CLAUDE_CONFIG_DIR}"; } || log_warn "Failed to create ${CLAUDE_CONFIG_DIR} with correct permissions"
```

agents/claude-code/deployment/entrypoint.sh
```diff
@@ -159,14 +159,12 @@ setup_config_dir() {
...
 
-    # Ensure the projects directory exists (WORKDIR is /workspace/projects)
+    # Ensure the projects directory exists
     # This separates global config (/workspace/.claude) from local auto-memory (/workspace/projects/.claude)
-    mkdir -p /workspace/projects
-    chmod g+w /workspace/projects 2>/dev/null || true
+    install -d -m 2775 /workspace/projects 2>/dev/null || { mkdir -p /workspace/projects && chmod 2775 /workspace/projects; }
```
hmoghani: Same here: warn if the fallback fails to set permissions.

```suggestion
    install -d -m 2775 /workspace/projects 2>/dev/null || { mkdir -p /workspace/projects && chmod 2775 /workspace/projects; } || log_warn "Failed to create /workspace/projects with correct permissions"
```

hmoghani (commented):
Added a couple of suggestions.

kami619:
@msager27 thanks for the fix., here is the summary and the changes LGTM. I am approving it and please test the changes manually once you merge it for any existing deployments.

No blocking issues found in PR #244.

  The change is coherent end to end: removing WORKDIR from agents/claude-code/deployment/Containerfile avoids the root-created directory problem on fresh PVCs, agents/claude-code/deployment/entrypoint.sh
  now creates /workspace/.claude and /workspace/projects with the intended group-writable mode, and the added cd /workspace/projects logic restores the “start in projects dir” behavior for both the main
  process and interactive bash sessions. The agents/claude-code/README.md update also matches that new runtime model.

### Open questions / residual risk:
  • The PR is well-scoped to fresh PVC deployments. It likely does not repair already-broken PVCs that already contain a root-owned /workspace/projects from older images, since the entrypoint still
    runs as a non-root user. I would not block the PR on that because the title and description are explicit, but it is worth treating as an operational limitation.
  • The proposed manual validation should explicitly cover both paths:
    • default container startup path
    • oc exec -it <pod> -- bash landing in /workspace/projects
  • The “interactive shell” fix is specifically for bash via .bashrc; it won’t automatically change cwd for other shells like sh. That seems acceptable, but it’s worth knowing.

### Change summary:
  • agents/claude-code/deployment/Containerfile: removes WORKDIR /workspace/projects
  • agents/claude-code/deployment/entrypoint.sh: uses install -d -m 2775, adds cd /workspace/projects for the main process, and adds the same default cwd behavior for interactive bash
  • agents/claude-code/README.md: updates the docs to describe the working directory behavior without referencing WORKDIR

  Assessment: Ready to merge with manual validation. The implementation matches the stated fix well; the remaining risk is mostly around upgrade behavior for already-broken PVCs, not the fresh-PVC path
  this PR is targeting.

kami619 (approved):

msager27:
@hmoghani Thanks for the review!  @kami619 Thanks for addressing the suggestion!

PR #245, opened 2026-07-09
feat(opencode): consolidate kustomize manifests from opencode-openshift

aakankshaduggal:
## Summary

- Moves the full kustomize deployment manifests from `aicatalyst-team/opencode-openshift` into `agents/opencode/deployment/` so all coding agent deployment content lives in one repo
- Adds base `manifests/` (namespace, ServiceAccount, Deployment, Service, Route, PVC, entrypoint.sh, config-template with vLLM + OGX providers)
- Adds `overlays/cli` (headless mode — no OAuth proxy, no Route, `oc exec` access)
- Adds `overlays/example` (user-customizable template with storage class, model endpoint, cookie secret)
- Updates `DEPLOYMENT.md` to be self-contained — no longer references the external repo
- Updates `overlays/mlflow-tracing` to point to local `manifests/` instead of requiring a separate clone

## Test plan

- [x] `oc kustomize` validates all four targets (manifests, cli, example, mlflow-tracing)
- [x] Deployed to `aduggal-opencode-dev` on ROSA cluster — pod 2/2 Running, OAuth proxy authenticated, OpenCode web UI serving against `vllm-120b` in `gpt-oss` namespace
- [ ] Review deployment structure matches openclaw pattern (`manifests/` + `overlays/`)

### DIFF
--- a/agents/opencode/deployment/DEPLOYMENT.md
+++ b/agents/opencode/deployment/DEPLOYMENT.md
@@ -35,7 +35,7 @@ This guide covers deploying [OpenCode](https://opencode.ai) as a coding agent on
 
 ## Project Structure
 
-```
+```text
 deployment/
 ├── manifests/                    # Base kustomize manifests (web mode + OAuth)
 │   ├── kustomization.yaml        # Kustomize entrypoint
@@ -128,6 +128,7 @@ The web mode deployment runs a two-container pod:
 2. **opencode-web** — OpenCode application serving the web UI on port 8003
 
 The entrypoint script (`manifests/entrypoint.sh`) handles:
+
 - Git workspace initialization
 - Config template variable substitution (BASE_URL, API_KEY, MODEL_NAME)
 - Optional MCP server config injection from a ConfigMap
--- a/agents/opencode/deployment/DEPLOYMENT.md
+++ b/agents/opencode/deployment/DEPLOYMENT.md
@@ -6,8 +6,8 @@ This guide covers deploying [OpenCode](https://opencode.ai) as a coding agent on
 
 | Field | Value |
 |-------|-------|
-| Image | `quay.io/opendatahub/odh-opencode-rhel9:1.4.4` |
-| OpenCode version | 1.4.4 (built from [opendatahub-io/opencode](https://github.com/opendatahub-io/opencode)) |
+| Image | `quay.io/opendatahub/odh-opencode-rhel9:latest` |
+| OpenCode version | Built from [opendatahub-io/opencode](https://github.com/opendatahub-io/opencode) |
 | Base | UBI 9 minimal |
 | License | MIT (OpenCode), Apache 2.0 (deployment manifests) |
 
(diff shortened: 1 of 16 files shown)


andrewdonheiser (changes requested):
## Review

### Critical (1)

**Image tag inconsistency between docs and manifests** — `DEPLOYMENT.md` declares the image as `:latest` but `deployment.yaml` and the CLI overlay both use `:20260619-194847e`. The doc's version pinning section also says "Avoid `:latest`", contradicting its own header table.
→ Align the DEPLOYMENT.md table to the actual pinned tag, or introduce a note explaining the discrepancy.
`[DEPLOYMENT.md:8, deployment.yaml:55]`

### Important (5)

1. **Service exposes port 8003 directly, enabling intra-cluster OAuth bypass** — `service.yaml` exposes both port 8443 (oauth-proxy) and 8003 (app). Any pod in the cluster can reach port 8003, bypassing OAuth entirely. The oauth-proxy reaches the app via `localhost:8003` — no Service port is needed for that.
→ Remove the `app` port (8003) from `service.yaml`, or add a NetworkPolicy restricting access.
`[service.yaml:14-17]`

2. **Example overlay's `CHANGE-ME` cookie secret could silently deploy** — The base manifest uses `<your-generated-secret>` (syntactically invalid — fails safely), but the example overlay uses `session_secret=CHANGE-ME` which is a valid string that deploys without error. A predictable cookie secret allows session forgery.
→ Use a syntactically-invalid placeholder like the base does, or add a validation step.
`[overlays/example/kustomization.yaml:33]`

3. **CLI overlay leaves stale Service resources and duplicates the full pod spec** — The overlay removes the Route and oauth-proxy container, but the base Service still has port 8443, the TLS cert annotation (triggering unnecessary cert generation), and the proxy-cookie Secret. Additionally, `$patch: replace` duplicates 60+ lines — if the base changes, the CLI patch silently diverges.
→ Add a Service patch to the CLI overlay removing port 8443 and the cert annotation. Consider strategic merge patch instead of full replacement.
`[overlays/cli/deployment-patch.yaml, overlays/cli/kustomization.yaml]`

4. **Placeholder `<your-storage-class>` in base creates invalid PVC** — Running `oc apply -k manifests/` without editing creates a PVC with a non-existent storage class that stays Pending with a non-obvious error. The openclaw pattern sets a real default (`gp3-csi`) with a comment.
→ Set a reasonable default or move the storage class patch to the example overlay only.
`[manifests/kustomization.yaml:36-38]`

5. **API_KEY double-exposure in env vars** — `entrypoint.sh` reads `API_KEY` from env, substitutes it into the config JSON, then exports the result as `OPENCODE_CONFIG_CONTENT` — the key now lives in two env vars. Any process dump or `/proc/<pid>/environ` read exposes it.
→ `unset API_KEY BASE_URL MODEL_NAME` after substitution, or use file-mounted secrets with `envsubst`/`jq --arg`.
`[entrypoint.sh:19-29]`

### Suggestions (3)

- **Labels don't follow the Kubernetes conventions openclaw uses** — openclaw uses `app.kubernetes.io/name` and `app.kubernetes.io/part-of`; opencode uses bare `app: opencode`. → Align with the repo convention. `[manifests/kustomization.yaml:8-9]`
- **Both vLLM and OGX providers enabled by default without explanation** — Both point to the same BASE_URL/API_KEY, which may confuse single-provider users. → Document the rationale or default to just `["vllm"]`. `[config-template.json:33]`
- **Bash variable substitution is fragile with special characters** — `CONFIG=${CONFIG//\${API_KEY}/$API_KEY}` breaks if the API key contains `/`, `\`, or `&`. → Consider `envsubst` or `jq --arg`. `[entrypoint.sh:19-21]`

### Strengths

- Clean kustomize base + overlay structure matching the openclaw pattern
- Good security defaults out of the box (runAsNonRoot, drop all caps, seccomp, no privilege escalation)
- DEPLOYMENT.md is now self-contained — removing the external repo dependency is a meaningful UX improvement
- Proper liveness/readiness probes, resource limits, and optional ConfigMaps

aakankshaduggal:
Thanks for the thorough review — pushed fixes in 8db4101:

**Critical**
- Image tag in DEPLOYMENT.md now matches the manifest pin (`20260619-194847e`)

**Important**
1. Removed port 8003 from Service — oauth-proxy reaches the app via localhost, no Service port needed
2. Example overlay cookie secret changed to `<generate-with-openssl-rand-base64-32>` (syntactically invalid, fails safely)
3. CLI overlay rewritten as strategic merge patch — deletes oauth-proxy container, proxy volumes, probes; removes Service and Route entirely in CLI mode
4. Base PVC storage class defaults to `gp3-csi` (matching openclaw pattern)
5. Entrypoint switched to `envsubst` for safe substitution + `unset API_KEY BASE_URL MODEL_NAME` after config is built

**Suggestions**
- Labels aligned to `app.kubernetes.io/name` + `app.kubernetes.io/part-of` (matching openclaw)
- Keeping both vLLM and OGX enabled — they share the same endpoint/key but represent the two provider aliases users configure against. Happy to add a doc note if that helps.

All four kustomize targets (manifests, cli, example, mlflow-tracing) validate cleanly.

andrewdonheiser (approved):
👍

PR #251, opened 2026-07-10
feat(codex): add Codex CLI container image for RHOAI

hmoghani:
## Summary

- Add container image for OpenAI Codex CLI (Apache 2.0) on RHOAI
- Single Containerfile installs Codex via npm on UBI 9 minimal (upstream disclaimer included)
- Containerfile.base available as alternative for source-built binary
- Entrypoint handles auth, model provider config, MCP injection, git credentials, and codex-run wrapper
- Deployment manifest with ImageStream, BuildConfig, ConfigMaps, PVC, Secret, and Deployment
- Comprehensive README (~700 lines) with vLLM compatibility matrix

## Upstream Disclaimer

The Codex CLI binary installed in this image is a pre-built artifact published by OpenAI via npm (`@openai/codex`), not by Red Hat. It has not been built, scanned, or validated according to Red Hat standards. `Containerfile.base` is available for users who need a source-built alternative.

## Files

| File | Description |
|------|-------------|
| `Containerfile` | RHOAI image (UBI 9 minimal, Codex via npm, entrypoint, MCP, session persistence) |
| `Containerfile.base` | Alternative: compiles Codex from source (Rust cross-compilation) |
| `Containerfile.openshell` | OpenShell sandbox variant |
| `entrypoint.sh` | Runtime setup: auth, vLLM provider, MCP, git creds, codex-run wrapper |
| `deployment.yaml` | OpenShift manifests (ImageStream, BuildConfig, ConfigMaps, PVC, Secret, Deployment) |
| `README.md` | Comprehensive deployment guide |
| `deployment/README.md` | Quick reference and file index |

## vLLM Compatibility

Requires upstream vLLM v0.25.1+. RHOAI 3.5 EA (v0.21.0) returns `"Unexpected message role"` because Codex sends `developer` role messages which that vLLM version does not support. RHOAI 3.6 EA (expected v0.26+) should include the fix (upstream PR #43590).

Harmony models (gpt-oss) reject namespace tools on all vLLM versions. Tracked: [vllm-project/vllm#49493](https://github.com/vllm-project/vllm/issues/49493).

| Model | Parser | vLLM | Tool Calling |
|---|---|---|---|
| Qwen3.6-27B | `qwen3_coder` | Upstream v0.25.1 | **Works** (tested end-to-end) |
| Qwen3-32B | `hermes` | Upstream v0.25.1 | **Works** (tested end-to-end) |
| Qwen3-32B | `qwen3_coder` | Upstream v0.25.1 | No (tool calls emitted as text) |
| Qwen3-8B | `qwen3_coder` | Upstream v0.25.1 | No (8B too small for tool calling) |
| Any model | any | RHOAI v0.21.0 | No (`"Unexpected message role"` — Codex uses `developer` role) |
| gpt-oss-120b | any | Any version | No (`"tool type namespace not supported"`) |

## Test plan

- [x] Image builds via npm on UBI 9 minimal (single Containerfile)
- [x] Container runs as non-root (UID 1001, restricted-v2 SCC)
- [x] Entrypoint generates config.toml from env vars (env vars override ConfigMap)
- [x] codex-run wrapper works via `bash -l` for both interactive and exec mode
- [x] MCP injection via MCP_CONFIG_JSON env var (tested, dedup on restart)
- [x] MCP injection via ConfigMap (tested, dedup on restart)
- [x] Git credential injection via GITHUB_PAT (tested)
- [x] API keys not persisted to PVC (env.sh uses single-quoted printf)
- [x] CODEX_API_KEY set via secretKeyRef for sleep infinity + oc exec mode
- [x] Config persists across pod restarts (PVC-backed)
- [x] Codex exec end-to-end with Qwen3.6-27B + qwen3_coder on upstream vLLM v0.25.1 (file creation works)
- [x] Codex exec end-to-end with Qwen3-32B + hermes on upstream vLLM v0.25.1 (file creation works)
- [x] Sanitized env var interpolation (TOML/shell injection prevention)
- [x] Agent listed in AGENTS.md non-standard section
- [ ] Codex with harmony models (gpt-oss) — blocked by vLLM namespace tool limitation
- [ ] Codex on RHOAI 3.5 EA vLLM (v0.21.0) — blocked by missing `developer` role support
- [x] No hardcoded URLs, models, or cluster-specific values


shricharan-ks (commented):
## Code Review — PR #251 (comprehensive, 2 review agents)

This PR adds a Codex CLI container image for RHOAI — multi-stage Containerfiles (Rust cross-compilation), an entrypoint script handling auth/config/MCP injection, OpenShift deployment manifests, and comprehensive README. 7 files changed, 1533 additions.

The container architecture is well-designed and the README is thorough. However, the entrypoint script has several injection vulnerabilities that should be addressed before merge, plus supply chain concerns in the Containerfiles.

---

### Findings (8, ranked by severity)

#### 1. 🔴 `entrypoint.sh:299` — Shell injection via PVC-persisted `env.sh` **(CONFIRMED)**

`OPENAI_BASE_URL` and `OPENAI_MODEL` are interpolated unsanitized into a generated `env.sh` script written to the PVC:
```bash
echo "export OPENAI_BASE_URL=\"${OPENAI_BASE_URL}\"" >> "${CODEX_DIR}/env.sh"
```

If `OPENAI_BASE_URL` contains `"` or backticks (e.g., via a misconfigured ConfigMap), the generated `env.sh` becomes an injection payload. Since `.bashrc` sources this file and it lives on the PVC, the payload survives pod restarts and executes on every `oc exec` session with access to all env vars including API keys.

**Fix**: Use `printf '%s' "$var"` with single quotes in the generated script, or write a non-executable env file parsed by the entrypoint rather than sourced by bash.

---

#### 2. 🔴 `entrypoint.sh:184` — TOML injection via model/URL env vars **(CONFIRMED)**

`OPENAI_MODEL` and `OPENAI_BASE_URL` are interpolated into a `config.toml` heredoc without escaping quotes or newlines. A crafted model name can inject a rogue `[model_providers]` section, routing all inference traffic (prompts + code context) to an attacker-controlled endpoint.

**Fix**: Sanitize values (strip quotes/newlines) or use a tool like `python3 -c "import toml; ..."` to generate the config safely.

---

#### 3. 🟡 `deployment.yaml:174` — API key in plaintext env var **(CONFIRMED)**

```yaml
- name: OPENAI_API_KEY
  value: "dummy"
```

When a user replaces `dummy` with a real key, it's stored in plaintext in the Deployment spec. Should use `valueFrom: secretKeyRef:` pointing to a Secret resource. CodeRabbit also flagged this.

---

#### 4. 🟡 `Containerfile:21` — Default BASE_IMAGE points to personal Quay registry **(CONFIRMED)**

```dockerfile
ARG BASE_IMAGE=quay.io/hmoghani/codex-base:latest
```

This is a personal namespace with a `:latest` floating tag. If the account is deleted or the image is overwritten, all downstream builds break. Should point to an org-owned registry (e.g., `quay.io/opendatahub/`) or clearly document that users must build the base locally first.

---

#### 5. 🟡 `entrypoint.sh:242` — MCP JSON-to-TOML conversion doesn't escape quotes **(CONFIRMED)**

MCP server args containing double-quotes (e.g., `--config={"key":"val"}`) produce malformed TOML. The `jq` wrapping adds outer quotes but doesn't escape inner ones, causing Codex to fail to parse its config and silently drop all MCP servers.

---

#### 6. 🟡 `Containerfile.base:31` — Rolling `rust:1-bookworm` builder tag **(PLAUSIBLE)**

The builder stage uses an unpinned `rust:1-bookworm` while the runtime is digest-pinned. A Rust/Debian update between rebuilds could introduce ABI mismatches with the pinned UBI9 runtime. Pin to at least a patch-level tag (e.g., `rust:1.87.0-bookworm`).

---

#### 7. 🟡 `README.md:1` — Agent not listed in AGENTS.md non-standard section **(CONFIRMED)**

This agent has no Makefile, no pyproject.toml, no FastAPI, no `/chat/completions`, no port 8080, and no tests/. It diverges from every standard convention. The PR should add a `codex/deployment` entry to the "Non-standard agents" section in AGENTS.md.

---

#### 8. 🟡 `README.md:63` — README says "Node.js 16" but actual version is 18+ **(CONFIRMED)**

`Containerfile.base` runs `microdnf install nodejs` without version pinning. UBI9 repos ship Node.js 18 (Node 16 reached EOL in September 2023). The README claim is inaccurate.

---

### Additional observations (not blocking)

- **`entrypoint.sh:313`**: Unquoted `${CODEX_EXTRA_ARGS}` breaks on model names with spaces or glob characters. Use a bash array instead.
- **`entrypoint.sh:210`**: MCP server names used unsanitized in grep/sed regex patterns — names with `.*` or `]` cause regex injection.
- **`entrypoint.sh:102`**: `GITHUB_PAT` written to `.git-credentials` is not scrubbed on exit — stale credentials persist on the PVC after rotation.

### What looks good

- Multi-stage build correctly separates Rust compilation from UBI9 runtime
- Non-root execution (UID 1001, restricted-v2 SCC)
- PVC-backed session persistence with config separation (env.sh excludes secrets — though the injection makes this moot)
- `codex-run` wrapper handles both interactive and exec modes cleanly
- vLLM compatibility matrix is well-documented with root cause analysis
- MCP dedup logic prevents duplicate injection on pod restarts

### Summary

**Findings #1 and #2 are blocking** — the shell and TOML injection vulnerabilities allow env var values to execute arbitrary commands or redirect inference traffic. Fix those, address the plaintext API key (#3) and personal registry (#4), and this is ready to merge.

agents/codex/README.md
```diff
@@ -0,0 +1,707 @@
...
+# Apply all resources (ImageStream, BuildConfig, ConfigMaps, PVC, Deployment)
+oc apply -f deployment.yaml
+
+# Build locally and push to the internal registry (recommended)
+#
+# Build locally (see Step 1), then push:
+oc registry login
+podman tag codex:latest \
+  $(oc registry info)/$(oc project -q)/codex:latest
+podman push $(oc registry info)/$(oc project -q)/codex:latest
```
bardenr: When deploying in our cluster, `oc registry info` gave me the intnernal registry route so I couldnt use this to tag and push an image from my laptop into the registry. I used this command to get the external regestry route:

```
oc get route -n openshift-image-registry -o custom-columns=HOST:.spec.host --no-headers
```

Also could be worth noting the user will need to login to this route with `podman login...`.

agents/codex/README.md
```diff
@@ -0,0 +1,707 @@
...
+
+The deployment runs `sleep infinity` so the pod stays alive for `oc exec` sessions. All examples below use `codex` as the deployment name.
+
+### The codex-run Wrapper
+
+The entrypoint generates a `codex-run` wrapper script at `~/.codex/codex-run` that includes all container-configured arguments (model, sandbox mode, provider config). Use it for all `oc exec` sessions:
+
+```bash
+oc exec deployment/codex -- bash -c '
+  codex-run exec "Your prompt here"
```
bardenr, line 185: I had to use the full path:

`~/.codex/codex-run exec "Your prompt here"`

to get this to work.

agents/codex/README.md
```diff
@@ -0,0 +1,707 @@
...
+'
+```
+
+### Interactive Mode
+
+For multi-turn conversations with a TTY:
+
+```bash
+oc exec -it deployment/codex -- bash -c '
+  codex-run
```
bardenr: Needed the full path here as well:

`~/.codex/codex-run`

bardenr (dismissed):
I was able to follow the readme to deploy on rosa. Only had issues with pushing a local image to the clusters image registry and some of the commands using `codex-run`

msager27:
**Tested on OpenShift — two issues found**

Deployed the PR on our cluster (built from source, deployed to OpenShift with OpenAI backend / gpt-5.6-sol). Found two issues:

**1. Missing `codex-code-mode-host` binary — Codex can't execute tools**

The `Containerfile.base` cargo build only compiles `codex` and `bwrap`:

```dockerfile
cargo build --release --target x86_64-unknown-linux-gnu --bin codex --bin bwrap
```

But Codex requires `codex-code-mode-host` for file operations and command execution. Without it, `codex exec` talks to the model but fails on any tool call:

```
ERROR codex_core::tools::router: error=failed to spawn code-mode host
/usr/local/bin/codex-code-mode-host: No such file or directory (os error 2)
```

Suggested fix — add to the cargo build and binary copy in `Containerfile.base`:

```dockerfile
cargo build --release --target x86_64-unknown-linux-gnu --bin codex --bin bwrap --bin codex-code-mode-host
```

And in Stage 2:
```dockerfile
COPY --from=builder /opt/codex-code-mode-host /usr/local/bin/codex-code-mode-host
```

**2. `CODEX_API_KEY` not set in `sleep infinity` mode**

The entrypoint propagates `OPENAI_API_KEY` → `CODEX_API_KEY` only when running `codex` directly (line 356-358), but the deployment uses `sleep infinity` + `oc exec`. In that path, `CODEX_API_KEY` is never set and `codex exec` fails with 401 Unauthorized.

Suggested fix — add to `deployment.yaml` env section:

```yaml
- name: CODEX_API_KEY
  valueFrom:
    secretKeyRef:
      name: codex-api-key
      key: api-key
```

🤖 *Generated with [Claude Code](https://claude.com/claude-code)*

agents/codex/deployment/Containerfile
```diff
@@ -0,0 +1,83 @@
...
+
+# Workspace and config directories
+RUN mkdir -p /workspace/projects /home/${USER_NAME}/.codex \
+    && chown -R ${USER_UID}:${USER_GID} /workspace /home/${USER_NAME} \
+    && chmod -R 775 /workspace /home/${USER_NAME}
+
+# Install Codex CLI via npm (pre-built binary, Apache 2.0)
+# The npm package is a thin wrapper that downloads the platform-specific
+# Rust binary. For a source-built alternative, see Containerfile.base.
+RUN npm install -g @openai/codex@${CODEX_VERSION}
```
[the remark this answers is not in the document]
hmoghani: not a concern in this PR anymore.  AGENTS.md is fixed.

msager27:
I did some further testing:

If I use ogx as my provider, I get tool: "namespace" errors (same as what @hmoghani encounters when using vLLM with gpt-oss models).

Digging into this some more, I think this is a Codex issue in that it uses a non-standard tool type.  There are a number of issues reported upstream including:

https://github.com/openai/codex/issues/32318
https://github.com/openai/codex/issues/31875
https://github.com/openai/codex/issues/31882

Maybe we can find a workaround until this is fixed upstream?

msager27:
Another issue I'm encountering is with the codex -> vllm (using Qwen3 8B) workflow.  While it can answer simple prompts, it doesn't handle tool calls.  For example, if I ask it to create a file, it will reason and report the number of tokens, but it won't create the file.  If I try the same test with an OpenAI backend, the file is created.  This issue might be specific to the model.

Note: I'm using the pre-built openai image for these latter tests.

msager27:
**Bug: `setup_model_provider()` appends top-level keys after TOML section headers**

During testing on OpenShift with Qwen 3.6-27B, `codex resume` crashes with:

```
Error loading config.toml: invalid type: string "qwen3.6-27b", expected u32
in `tui.model_availability_nux`
```

**Root cause:** `setup_model_provider()` in `entrypoint.sh` removes existing `model =` and `model_provider =` lines with `sed`, then appends them at the end of `config.toml` using `cat >>`. If the file already contains a `[section]` header (e.g. `[tui.model_availability_nux]` written by a previous interactive session), the appended `model = "..."` lands inside that section instead of at the top level. TOML assigns any key after a `[section]` header to that section until the next header.

**Suggested fix:** Insert `model =` and `model_provider =` at the top of the file (e.g. `sed -i '1i ...'`) instead of appending with `cat >>`, or write the entire config from scratch rather than patching a file that may have accumulated sections from interactive use.

🤖 *Generated with [Claude Code](https://claude.com/claude-code)*

agents/codex/README.md
```diff
@@ -0,0 +1,692 @@
...
+| Model | Parser | vLLM Version | Result |
+|---|---|---|---|
+| gpt-oss-120b | `openai` | RHOAI v0.21.0 | `"tool type namespace not supported"` |
+| gpt-oss-120b | `openai` | Upstream v0.25.1 | `"tool type namespace not supported"` |
+| gpt-oss-120b | `hermes` | Upstream v0.25.1 | `"tool type namespace not supported"` |
+| Any model | any | RHOAI v0.21.0 | `"Unexpected message role"` — Codex sends `developer` role which this version doesn't support |
+| Qwen3.6-27B | `qwen3_coder` | Upstream v0.25.1 | Works (full tool calling) |
+| Qwen3-32B | `hermes` | Upstream v0.25.1 | Works (full tool calling) |
+| Qwen3-8B | `qwen3_coder` | Upstream v0.25.1 | Namespace accepted, but no tool execution (8B too small) |
+
```
msager27, line 526: We should note somewhere that using OGX results in the tool type namespace error as well.

agents/codex/README.md
```diff
@@ -0,0 +1,692 @@
...
+  }'
+```
+
+**Namespace tool type compatibility:**
+
+Codex sends `namespace` tool types with every request. Whether vLLM accepts them depends on the **model architecture**, not the parser or vLLM version.
+
+| Model | Parser | vLLM Version | Result |
+|---|---|---|---|
+| gpt-oss-120b | `openai` | RHOAI v0.21.0 | `"tool type namespace not supported"` |
```
msager27: RHOAI v0.21.0 is confusing from a version perspective.  Maybe RHOAI 3.5 or RHAII 3.5, possibly including in parentheses the underlying base image.

agents/codex/README.md
```diff
@@ -0,0 +1,692 @@
...
+| gpt-oss-120b | `openai` | Upstream v0.25.1 | `"tool type namespace not supported"` |
+| gpt-oss-120b | `hermes` | Upstream v0.25.1 | `"tool type namespace not supported"` |
+| Any model | any | RHOAI v0.21.0 | `"Unexpected message role"` — Codex sends `developer` role which this version doesn't support |
```
msager27: Same as above comment

shricharan-ks (commented):
## Re-review — PR #251 (after 25 commits, addressing all reviewer feedback)

Comprehensive re-review checking all findings from msager27, bardenr, CodeRabbit, and my previous review.

---

### Resolved findings ✅ (10/13)

| Finding | Source | Status | Evidence |
|---------|--------|--------|----------|
| Missing `codex-code-mode-host` binary | msager27 | ✅ FIXED | Containerfile switched to npm install; Containerfile.base now builds `--bin codex-code-mode-host` |
| config.toml key ordering crash | msager27 | ✅ FIXED | `setup_model_provider()` now uses `sed -i "1i\\..."` to insert at top |
| OGX namespace tool errors | msager27 | ✅ FIXED | Documented in README and AGENTS.md |
| codex-run needs full path | bardenr | ✅ FIXED | README uses `bash -l` which sources .bashrc adding ~/.codex to PATH |
| oc registry info gives internal route | bardenr | ✅ FIXED | README uses `oc get route -n openshift-image-registry` for external route |
| Shell injection via env.sh | Previous review | ✅ FIXED | `sanitize_value()` strips `"`, `'`, `` ` ``, `$`, `\`, newlines; single-quoted printf; API keys not persisted |
| API key in plaintext deployment.yaml | Previous review | ✅ FIXED | Now uses `secretKeyRef` referencing K8s Secret `codex-api-key` |
| Personal registry in Containerfile | Previous review | ✅ FIXED | Main Containerfile and Containerfile.base use digest-pinned UBI9 |
| MCP config duplication on restart | CodeRabbit | ✅ FIXED | All MCP injection methods check `grep -q` before appending |
| Fallback model name "default" | CodeRabbit | ✅ FIXED | No default fallback exists; errors on missing model |

---

### Remaining findings (3)

#### 1. 🟡 `entrypoint.sh:~160` — TOML injection in fallback path **(CONFIRMED)**

The main `setup_model_provider()` path correctly calls `sanitize_value()` on `OPENAI_MODEL` (line ~173). But the **early-return branch** when `OPENAI_BASE_URL` is empty writes `OPENAI_MODEL` directly without sanitization:

```bash
echo "model = \"${OPENAI_MODEL}\"" >> "${config_file}"
```

A crafted `OPENAI_MODEL` with embedded `"` or newlines can inject rogue TOML keys. This is the same class of bug that was fixed in the main path but missed in this branch.

**Fix**: Call `sanitize_value` on `OPENAI_MODEL` in the fallback path too.

---

#### 2. 🟡 `entrypoint.sh:~245` — MCP JSON-to-TOML quoting still broken **(CONFIRMED)**

The jq expression `map("\"" + . + "\"")` naively wraps MCP args in double quotes without escaping inner quotes. An arg like `--config={"key":"val"}` produces:

```toml
args = ["--config={"key":"val"}"]
```

This is invalid TOML that crashes Codex on startup. From the original review — not yet addressed.

**Fix**: Use `jq`'s `@json` filter or escape inner double-quotes with `gsub`.

---

#### 3. 🟡 `Containerfile.openshell:9` — Personal registry still used **(CONFIRMED)**

```dockerfile
FROM quay.io/hmoghani/openshell-base:latest
```

The main Containerfile and Containerfile.base were fixed to use `registry.access.redhat.com`, but `Containerfile.openshell` still references a personal Quay.io repository.

---

### Summary

**10 of 13 findings from all reviewers are resolved** — impressive iteration over 25 commits. The remaining 3 are all non-blocking for merge (the TOML injection requires a crafted env var in a branch that rarely fires, the MCP quoting affects a narrow edge case, and Containerfile.openshell is experimental). Fix the sanitize_value gap in the fallback path (#1) as priority.

hmoghani:
@shricharan-ks thanks for the review! All issues are addressed and fixed.

aakankshaduggal (approved):
Tested end-to-end on ROSA cluster (`hmoghani-codex` namespace):

- **Image build**: `codex:v4-1785175052` built and deployed successfully from `Containerfile` (npm variant)
- **Entrypoint**: All setup stages work correctly — env validation, CODEX_HOME symlink, vLLM provider config, MCP dedup, `codex-run` wrapper, API key propagation
- **Model serving**: Validated with both Qwen3-8B and Qwen3.6-27B via vLLM (`--tool-call-parser qwen3_coder`). Tool calls generate correctly with the 27B model.
- **Security**: API keys via `secretKeyRef`, `sanitize_value()` prevents injection, non-root user (UID 1001), restricted-v2 SCC compatible
- **MLflow tracing**: Built `codex-mlflow` layer on top of this base image — traces export successfully with 10-span traces (AGENT + TOOL + LLM spans)

Minor non-blocking notes:
- `--tls-verify=false` still in README registry commands — consider `oc registry login` for production guidance
- Unquoted `${CODEX_EXTRA_ARGS}` in entrypoint exec calls — edge case with unusual model names

LGTM — well-structured, thoroughly reviewed, and production-tested.

msager27 (approved):

shricharan-ks (approved):
## Re-review — All findings resolved ✅

The latest commits (`c710d115`, `67387f9a`) address all 3 remaining findings from the previous review:

### Finding 1 (TOML injection in fallback path) — FIXED ✅
The no-BASE_URL fallback now sanitizes OPENAI_MODEL:
```bash
echo "model = \"$(sanitize_value "${OPENAI_MODEL}")\"" >> "${config_file}"
```

### Finding 2 (MCP JSON-to-TOML quoting) — FIXED ✅
Args now use jq's `@json` filter which properly escapes embedded quotes:
```
.args | map(@json) | join(", ")
```

### Finding 3 (Containerfile.openshell personal registry) — FIXED ✅
Changed from `quay.io/hmoghani/openshell-base:latest` to `quay.io/aipcc/agentic-ci/openshell:0.3.27` (org-owned registry, version-pinned tag).

### Full scorecard across all reviewers (13/13 resolved)

| Finding | Source | Status |
|---------|--------|--------|
| Missing codex-code-mode-host | msager27 | ✅ |
| config.toml key ordering crash | msager27 | ✅ |
| OGX namespace tool errors doc | msager27 | ✅ |
| codex-run full path | bardenr | ✅ |
| oc registry internal route | bardenr | ✅ |
| Shell injection via env.sh | Previous review | ✅ |
| TOML injection (main path) | Previous review | ✅ |
| TOML injection (fallback path) | Previous review | ✅ |
| API key plaintext → Secret | Previous review | ✅ |
| Personal registry (Containerfile) | Previous review | ✅ |
| Personal registry (openshell) | Previous review | ✅ |
| MCP JSON-to-TOML quoting | Previous review | ✅ |
| MCP config duplication | CodeRabbit | ✅ |

LGTM — ready to merge.

PR #252, opened 2026-07-13
fix(btest): use external MLflow route for workspace routing (RHAIENG-6128)

andrewdonheiser:
## Summary
- When `MLFLOW_WORKSPACE` is set and the discovered `MLFLOW_INTERNAL_URI` is an internal `.svc.cluster.local` URL, substitute the external route URL which supports workspace routing
- Add preflight validation that tests the resolved internal URI is reachable using the `/mlflow/health` endpoint (the experiments API requires workspace context that a simple curl check cannot provide)

Fixes the adapter pod MLflow enrichment failure where `mlflow_run_id` comes back null because the path-stripped internal service URL cannot route workspace-scoped API calls.

## Jira
- RHAIENG-6128
- Part of RHAIENG-6147 (Tune behavioral test suites for CI readiness)
- Tracking: RHAIENG-6227

## Test plan
- [x] Deploy EvalHub on cluster with workspace-enabled MLflow
- [x] Run `run-e2e.sh` — verify `mlflow_run_id` is non-null
- [x] Check preflight output shows workspace URI substitution message
- [ ] ~~Verify `tool_selection` scores are 0.8+ (not degraded 0.2)~~ — scores not available in eval results format; `mlflow_run_id` presence is the primary success signal

## Test results

Cluster: `agen-e2e-rhoai2` / namespace: `adonheis-testing`
Run command: `OC_NAMESPACE=adonheis-testing MLFLOW_INTERNAL_URI="https://mlflow.redhat-ods-applications.svc.cluster.local:8443" bash evals/evalhub_adapter/tests/run-e2e.sh`

Preflight confirmed workspace routing swap:
```
[OK]   MLflow internal URI (override): https://mlflow.redhat-ods-applications.svc.cluster.local:8443
[WARN] Internal URI uses .svc.cluster.local but MLFLOW_WORKSPACE=adonheis-testing; switching to external route for workspace routing
[OK]   MLflow internal URI (workspace-aware): https://rh-ai.apps.rosa.agen-e2e-rhoai2.p5ui.p3.openshiftapps.com/mlflow
[OK]   MLflow internal URI reachable (adapter)
```

All 10 standard agents completed with non-null `mlflow_run_id`:

| Agent | mlflow_run_id |
|-------|---------------|
| react_agent | `80c22d7f03c74b16b9f7b7691aa8c3f6` |
| openai_responses_agent | `4f54c6bd86fe41578a7fe9659fc95e42` |
| autogen_mcp_agent | `74ec4e54d3514475be83cc075754523f` |
| crewai_websearch_agent | `1891a0bb4a5a48b6905df08049f70802` |
| agentic_rag | `0c21e59a95ec4b7ab4928ad13948062a` |
| db_memory_agent | `637347cdc6264b9b90e34313c2cb4531` |
| llamaindex_websearch_agent | `9b9cb3ed57d441768413cf43f5350a6d` |
| hitl_agent | `9a5d8af061de4db6a3f3017b2624588c` |
| google_adk_agent | `375b3b771f1b4b2ca0f667e2ee82bffc` |
| a2a_langgraph_crewai | `47842c46918d41a8a6266757c722f7dd` |

Langflow eval submission failed (pre-existing issue, unrelated to this PR).


shricharan-ks (commented):
## Code Review — PR #252 (shell script, verified)

This PR fixes MLflow workspace routing in the EvalHub adapter's e2e test script. When `MLFLOW_INTERNAL_URI` is an internal `.svc.cluster.local` URL and workspace routing is needed, it swaps to the external route URL. It also adds a preflight health check against the resolved URI. 1 file changed, 29 lines added.

The fix is well-motivated (internal service URLs can't route workspace-scoped API calls) and test results show it works on the target cluster. However, there are some edge cases and consistency issues.

---

### Findings (4 survived verification)

#### 1. 🟡 `run-e2e.sh:329` — Workspace swap doesn't guard against empty `MLFLOW_TRACKING_URI` **(CONFIRMED)**

```bash
MLFLOW_INTERNAL_URI="${MLFLOW_TRACKING_URI}"
```

If `MLFLOW_TRACKING_URI` auto-discovery fails (the `preflight_fail` on line 296 logs a failure but does **not** exit — the script continues until the summary check at line ~475), and the user overrides `MLFLOW_INTERNAL_URI` with a `.svc.cluster.local` URL, the swap fires and sets `MLFLOW_INTERNAL_URI` to an **empty string**. All adapter pods get a blank tracking URI and fail silently.

**Fix**: Add a guard before the swap:
```bash
if [[ -z "${MLFLOW_TRACKING_URI:-}" ]]; then
  preflight_warn "Cannot swap to external route — MLFLOW_TRACKING_URI not discovered"
else
  MLFLOW_INTERNAL_URI="${MLFLOW_TRACKING_URI}"
  ...
fi
```

---

#### 2. 🟡 `run-e2e.sh:500` — Health check path wrong for path-stripped internal URIs **(CONFIRMED)**

```bash
_health_url="${_clean_uri}/mlflow/health"
[[ "${_clean_uri}" == */mlflow ]] && _health_url="${_clean_uri}/health"
```

When `MLFLOW_INTERNAL_URI` is the path-stripped internal URL (e.g. `https://mlflow.ns.svc.cluster.local:8443`), the health URL becomes `.../mlflow/health`. But the internal service has **no `/mlflow` prefix** — it was stripped at line 312 precisely because the service exposes the API at root. The actual health endpoint is `/health`.

This only fires when the workspace swap does NOT trigger (non-workspace deployments). The health check returns 404 and emits a misleading warning even though the service is healthy.

**Fix**: When the URI doesn't end with `/mlflow`, use `/health` instead of `/mlflow/health`:
```bash
[[ "${_clean_uri}" == */mlflow ]] && _health_url="${_clean_uri}/health" || _health_url="${_clean_uri}/health"
```
Or simply always use `/health` as the suffix.

---

#### 3. 🟡 `run-e2e.sh:327` — `OC_NAMESPACE` hardcoded default makes swap condition always true **(CONFIRMED)**

Line 17: `OC_NAMESPACE="${OC_NAMESPACE:-adonheis-btest1}"`. The swap condition checks `-n "${OC_NAMESPACE:-}"` which **always passes** because of the hardcoded default. This means the swap fires for ANY `.svc.cluster.local` URI, even on clusters with non-workspace MLflow deployments where the internal URI would work fine.

This may be acceptable for now (the script is developer-centric, always runs in a namespace context), but the condition doesn't actually check whether workspace routing is needed — it just checks whether a namespace is set.

---

#### 4. 🟡 `run-e2e.sh:302` — Stale comment contradicts new behavior **(CONFIRMED)**

Lines 302-304 state: *"The adapter pod cannot reach the external route (SSL cert mismatch)"*. After the workspace swap, `MLFLOW_INTERNAL_URI` IS the external route. The comment and the code now assert opposite things. A future developer reading the comment would believe the external route is off-limits for adapters.

**Fix**: Update the comment to reflect that external route is used when workspace routing is required (with `MLFLOW_TRACKING_INSECURE_TLS=true` to handle TLS).

---

### What looks good

- Clean pattern matching for `.svc.cluster.local` detection
- Trailing-slash edge case handled correctly in the health check URL construction
- Health check properly uses `/mlflow/health` (not the experiments API which requires workspace context)
- Warning-level severity for health check is appropriate — doesn't block on transient issues
- Well-documented commit history showing iterative refinement

### Summary

The core fix (workspace routing swap) works correctly in the intended scenario. The 4 findings are edge cases and consistency issues — no blocking bugs for the primary use case. Most actionable: add a guard for empty `MLFLOW_TRACKING_URI` (#1) and update the stale comment (#4).

andrewdonheiser:
Thanks for the thorough review @shricharan-ks! Addressed in e114a89 and 861165b.

**Finding 1 — Empty `MLFLOW_TRACKING_URI` guard** ✅ Fixed
Added a guard that emits a warning instead of silently setting `MLFLOW_INTERNAL_URI` to empty.

**Finding 2 — Health check path for path-stripped URIs** ❌ Not a bug (verified on cluster)
Tested both internal service and external route on `agen-e2e-rhoai2`:
- `https://mlflow.redhat-ods-applications.svc.cluster.local:8443/health` → **404**
- `https://mlflow.redhat-ods-applications.svc.cluster.local:8443/mlflow/health` → **200**

The MLflow service serves health at `/mlflow/health` regardless of whether it's internal or external. The original logic (`/mlflow/health` by default, collapsed to `/health` when URI already ends with `/mlflow`) is correct — left as-is.

**Finding 3 — `OC_NAMESPACE` default always triggers swap** 🔵 Acknowledged
Agreed this is acceptable for now. The script always runs in a namespace context. Noted for future if we need non-workspace support.

**Finding 4 — Stale comment** ✅ Fixed
Updated the comment block at the internal URI discovery to reflect that the external route is used for workspace-scoped deployments.

shricharan-ks (approved):
## Re-review — All findings addressed ✅

Commits `e114a89d` and `861165b9` resolve all 4 findings from the previous review:

### Finding 1 (empty MLFLOW_TRACKING_URI guard) — FIXED ✅
Now guarded: `if [[ -z "${MLFLOW_TRACKING_URI:-}" ]]; then preflight_warn ...`. Additionally, the `/mlflow` path-appending logic correctly ensures the external route always has the required prefix:
```bash
[[ "${_ext_uri}" != */mlflow && "${_ext_uri}" != */mlflow/ ]] && _ext_uri="${_ext_uri%/}/mlflow"
```
Edge cases verified: bare hosts, trailing slashes, URIs already ending in `/mlflow`.

### Finding 2 (health check path) — Intentionally kept ✅
Commit `861165b9` restores `/mlflow/health` as the correct health endpoint. The internal MLflow service exposes its endpoints under `/mlflow/...` — the path-stripping only applies to the SDK base URL, not health checks. The doubling guard (`*/mlflow` → `/health`) works correctly.

### Finding 3 (OC_NAMESPACE always set) — Acceptable as-is ✅
`MLFLOW_WORKSPACE` is always set to `OC_NAMESPACE` in the provider config, so workspace routing is always needed when the script runs. The tautological condition is harmless.

### Finding 4 (stale comment) — FIXED ✅
Updated from "SSL cert mismatch" to accurate description about workspace-scoped deployments and `MLFLOW_TRACKING_INSECURE_TLS`.

No new issues introduced by the fix commits. LGTM — ready to merge.

PR #253, opened 2026-07-13
fix(vanilla_python): add Chat Completions API fallback

andrewdonheiser:
## Summary
- Add automatic fallback from Responses API to Chat Completions API when the upstream model server (e.g. vLLM) returns HTTP 400/404
- Cache API detection result so subsequent calls skip the probe
- Handle both response formats in `_get_output_text_from_response()`
- **Fix**: Always fall back on 400/404, even if the Responses API previously succeeded (vLLM partially supports Responses API — simple requests succeed but multi-turn conversations fail)

The agent uses ReAct-style text patterns for tool calling (not structured API tool calling), so the fallback only needs to handle message/response format differences. The agentic loop (`query()`) required zero changes.

## Jira
- RHAIENG-6148
- Part of RHAIENG-6147 (Tune behavioral test suites for CI readiness)
- Tracking: RHAIENG-6227

## Test plan
- [x] Deploy agent with vLLM backend (no Responses API support)
- [x] Run behavioral tests — **24/27 pass** (previously 23/24 failed)
- [x] Verify tool calling works end-to-end via text patterns
- [x] Check logs show fallback to Chat Completions API on first call

## Test results (behavioral tests against vLLM on OpenShift)

**Cluster**: agen-e2e-rhoai2 (ROSA HCP), namespace: `adonheis-testing`
**Model**: qwen2-5-7b-instruct via vLLM

| Suite | Passed | Failed | Total |
|-------|--------|--------|-------|
| Cost/Latency | 2 | 0 | 2 |
| Reliability | 2 | 1 | 3 |
| Response Quality | 12 | 0 | 12 |
| Tool Usage | 8 | 2 | 10 |
| **Total** | **24** | **3** | **27** |

### Failures (pre-existing, not regressions)

1. **test_pass_at_k_multi_tool** / **test_multi_tool_selection["Which brand offers the best value"]** — Model flakiness: sometimes calls only `search_price` and skips `search_reviews` on comparison queries. Not related to API fallback.
2. **test_tool_not_called_for_greeting** — Agent crashes with `Unknown action: None` when model responds to "Hello" — the ReAct regex matches `Action: None` in the output. Pre-existing agent bug in action parsing.


shricharan-ks (commented):
## Code Review — PR #253 (medium effort, 6 angles, verified)

This PR adds automatic fallback from the Responses API to Chat Completions API when the upstream model server (e.g. vLLM) returns HTTP 400/404. It handles both response formats in `_get_output_text_from_response()`, caches the API detection result per-agent-instance, and adds comprehensive unit tests. 2 files changed, 192 additions, 7 deletions.

---

### Findings (6 survived verification, most severe first)

#### 1. 🔴 `agent.py:241` — `if temp != 0` silently drops temperature=0, API defaults to 1.0 **(CONFIRMED)**

```python
if temp != 0:
    kwargs["temperature"] = temp
```

The constructor defaults `temperature=0` for deterministic output. But `0` is falsy in Python, so `if temp != 0:` evaluates to `False` — temperature is never sent to the API. Both OpenAI's Chat Completions and Responses APIs default to `temperature=1.0` when omitted. **Every request runs non-deterministically** instead of the intended deterministic mode.

This is **pre-existing** in the Responses API path (line 283) but the PR propagates it to the new `_chat_completions_create`. Fix: `if temp is not None:` or always include `"temperature": temp` in kwargs.

---

#### 2. 🟡 `agent.py:195` — API detection cache resets every HTTP request **(CONFIRMED)**

`_AIAgentAdapter.run()` creates a new `AIAgent` per HTTP request. Since `_use_responses_api` is an instance variable initialized to `None`, the cache never persists across requests. Against a vLLM backend that doesn't support the Responses API, **every request** probes → fails → retries via Chat Completions, doubling first-turn latency and flooding logs with fallback warnings.

The cache works within a single `query()` call's multi-turn ReAct loop (2-5 turns), but not across HTTP requests. Consider lifting the detection state to `_AIAgentAdapter` (process-lifetime) or a module-level variable.

---

#### 3. 🟡 `test_agent.py` — No test for the multi-turn fallback scenario the PR fixes **(CONFIRMED)**

The PR description says: *"Fix: Always fall back on 400/404, even if the Responses API previously succeeded (vLLM partially supports Responses API — simple requests succeed but multi-turn conversations fail)."*

All 5 tests exercise **single, isolated** `_llm_create()` calls. No test calls `_llm_create` twice where the first succeeds (→ `True`) and the second gets 400 and must transition to `False`. This is the exact scenario motivating commit `d525731a` and is the one most likely to regress.

Missing test: mock `responses.create` to succeed on call 1, raise `APIStatusError(400)` on call 2, verify call 2 falls back to `chat.completions.create`.

---

#### 4. 🟡 `agent.py:293` — HTTP 400 is too broad for "endpoint not supported" **(PLAUSIBLE)**

The except block treats any 400 as "server doesn't support Responses API" and permanently sets `_use_responses_api = False`. But HTTP 400 also covers legitimate request errors (too-long context, malformed content, unsupported parameter). A single bad request on a server that **does** support the Responses API would permanently disable it.

Consider checking the error body/message for endpoint-not-found indicators, or limiting the permanent flip to 404 only.

---

#### 5. 🟡 `agent.py:137` — `choices[0].message` accessed without guard **(PLAUSIBLE)**

```python
choices = getattr(response, "choices", None)
if choices:
    return getattr(choices[0].message, "content", None) or ""
```

The Responses API path uses defensive `getattr()` throughout, but the new Chat Completions path directly accesses `.message` on `choices[0]`. An OpenAI-compatible server returning a non-standard choices schema (e.g., streaming `delta` instead of `message`) would raise an unhandled `AttributeError`.

---

#### 6. 🟡 `agent.py:233/273` — Parameter resolution duplicated across methods **(PLAUSIBLE)**

`_chat_completions_create` and `_llm_create` both have identical 3-line parameter resolution (`msg_list = messages if...`, `temp = ...`, `model_id = ...`) and the same `if temp != 0` guard. Adding a parameter to one without the other would cause silent behavioral divergence between the two API paths.

---

### What looks good

- **Fallback design**: The try/except structure correctly catches 400/404 and falls back, even after a previous Responses API success — the key fix for vLLM's partial support
- **Response format handling**: `_get_output_text_from_response` correctly dispatches between Chat Completions (`choices`) and Responses API (`output`) formats
- **Test coverage**: 5 well-structured unit tests covering success caching, 400/404 fallback, error propagation, and cached fast-path
- **Zero changes to `query()`**: The agentic loop is correctly insulated from the API abstraction change
- **Clean typing modernization**: `Optional[str]` → `str | None`, `Dict` → `dict` throughout

### Summary

Finding #1 (temperature=0) is the most impactful — affects every request. It's pre-existing but worth fixing in this PR since `_chat_completions_create` is new code. Finding #2 (cache scope) is a design improvement worth considering. Finding #3 (missing test) is easy to add and covers the exact scenario that motivated the fix.

andrewdonheiser:
## Review Response — `d58ca66`

Thanks for the thorough reviews @shricharan-ks and CodeRabbit! Pushed a commit addressing the actionable findings.

### Addressed

| # | Finding | Fix |
|---|---------|-----|
| 1 | 🔴 `if temp != 0` silently drops temperature=0 (API defaults to 1.0) | Changed to `if temp is not None:` in both `_chat_completions_create` and `_llm_create` |
| 2 | 🟡 HTTP 400 too broad for permanent cache flip | Only 404 now permanently sets `_use_responses_api = False`. On 400, falls back for the current call only without caching |
| 3 | 🟡 No test for multi-turn fallback (success → later failure) | Added `test_fallback_on_400_after_prior_success`, `test_fallback_on_404_after_prior_success`, and `test_temperature_zero_is_passed` |

### Not addressed (follow-up)

- **Cache scope resets per HTTP request** (#2): `_AIAgentAdapter.run()` creates a new `AIAgent` per request, so the detection cache only persists within one `query()` call's multi-turn loop. Lifting the cache to process-lifetime scope is a design improvement worth a separate PR.
- **`choices[0].message` without guard** (#5): Low risk — Chat Completions always returns `.message` in non-streaming mode; we never use streaming in `_llm_create`.
- **Parameter duplication** (#6): Minor DRY concern, not in scope of this fix PR.

All 12 unit tests pass, lint and formatting clean.

tarun-etikala (approved):
Clean PR — fallback logic is correct, the 400-vs-404 distinction is well-reasoned for vLLM's partial Responses API support, and test coverage is thorough.

Two suggestions (non-blocking):

1. **API detection cache doesn't persist across HTTP requests** — `_AIAgentAdapter.run()` creates a new `AIAgent` per request, so `_use_responses_api` resets to `None` each time. On a vLLM backend, every request pays for a wasted 404 probe before falling back. Consider lifting the flag to `_AIAgentAdapter` and syncing it back after each `query()` call. The log message "permanently falling back" is also misleading since the state is lost on the next request.

2. **Minor test style** — `test_temperature_zero_is_passed` uses `call_kwargs[1]["temperature"]`; `call_kwargs.kwargs["temperature"]` is slightly more readable.

PR #260, opened 2026-07-13
feat(opencode): add session persistence and skills injection

msager27:
## Description

- Redirect OpenCode config, data, and state directories to PVC-backed paths via XDG symlinks so sessions survive pod restarts
- Set workingDir to PVC mount (/opt/app-root/workspace) so workspace files also persist across restarts
- Add skills injection via ConfigMap mounted at /etc/opencode-skills/ and symlinked into the OpenCode config directory
- Set XDG_STATE_HOME explicitly to work around 755 permissions on ~/.local/state/ under OpenShift's random UID
- Add skills volume and mount to deployment.yaml
- Replace envsubst with jq for config template substitution (envsubst is not available in the container image)
- Write config to persistent file in CLI mode instead of env var
- Update DEPLOYMENT.md with session persistence and skills injection documentation

## Testing

- [ ] Ask OpenCode to create a file and verify it lands on the PVC (/opt/app-root/workspace/)
- [ ] Restart pod and verify workspace files survive
- [ ] List sessions before and after pod restart — confirm they match
- [ ] Resume a previous session with opencode --continue after pod restart
- [ ] Inject a skill via ConfigMap and verify OpenCode discovers and uses it (→ Skill in output)

### DIFF
--- a/agents/opencode/deployment/manifests/entrypoint.sh
+++ b/agents/opencode/deployment/manifests/entrypoint.sh
@@ -1,11 +1,128 @@
 #!/bin/bash
 set -euo pipefail
 
+# =============================================================================
+# OpenCode Data Directory Setup (session persistence)
+# =============================================================================
+# OpenCode stores config in ~/.config/opencode and session data in
+# ~/.local/share/opencode. This function redirects both to a PVC-backed
+# location so sessions persist across pod restarts.
+# =============================================================================
+setup_opencode_dirs() {
+    export OPENCODE_DATA_DIR="${OPENCODE_DATA_DIR:-/opt/app-root/workspace/.opencode}"
+
+    mkdir -p "${OPENCODE_DATA_DIR}/config/opencode"
+    mkdir -p "${OPENCODE_DATA_DIR}/data/opencode"
+    mkdir -p "${OPENCODE_DATA_DIR}/state/opencode"
+
+    export XDG_CONFIG_HOME="${OPENCODE_DATA_DIR}/config"
+    export XDG_DATA_HOME="${OPENCODE_DATA_DIR}/data"
+    export XDG_STATE_HOME="${OPENCODE_DATA_DIR}/state"
+
+    local home_config="${HOME}/.config"
+    local home_data="${HOME}/.local/share"
+    local home_state="${HOME}/.local/state"
+
+    # Symlink ~/.config/opencode -> persistent config
+    mkdir -p "${home_config}"
+    if [[ -L "${home_config}/opencode" ]]; then
+        local current_target
+        current_target=$(readlink "${home_config}/opencode")
+        if [[ "${current_target}" != "${XDG_CONFIG_HOME}/opencode" ]]; then
+            ln -sfn "${XDG_CONFIG_HOME}/opencode" "${home_config}/opencode"
+        fi
+    else
+        if [[ -d "${home_config}/opencode" ]]; then
+            if [[ -n "$(ls -A "${home_config}/opencode" 2>/dev/null)" ]]; then
+                cp -rn "${home_config}/opencode/"* "${XDG_CONFIG_HOME}/opencode/" 2>/dev/null || true
+            fi
+            mv "${home_config}/opencode" "${home_config}/opencode.migrated" 2>/dev/null || rm -rf "${home_config}/opencode" 2>/dev/null || true
+        fi
+        ln -sfn "${XDG_CONFIG_HOME}/opencode" "${home_config}/opencode" 2>/dev/null || true
+    fi
+
+    # Symlink ~/.local/share/opencode -> persistent data
+    mkdir -p "${home_data}"
+    if [[ -L "${home_data}/opencode" ]]; then
+        local current_target
+        current_target=$(readlink "${home_data}/opencode")
+        if [[ "${current_target}" != "${XDG_DATA_HOME}/opencode" ]]; then
+            ln -sfn "${XDG_DATA_HOME}/opencode" "${home_data}/opencode"
+        fi
+    else
+        if [[ -d "${home_data}/opencode" ]]; then
+            if [[ -n "$(ls -A "${home_data}/opencode" 2>/dev/null)" ]]; then
+                cp -rn "${home_data}/opencode/"* "${XDG_DATA_HOME}/opencode/" 2>/dev/null || true
+            fi
+            mv "${home_data}/opencode" "${home_data}/opencode.migrated" 2>/dev/null || rm -rf "${home_data}/opencode" 2>/dev/null || true
+        fi
+        ln -sfn "${XDG_DATA_HOME}/opencode" "${home_data}/opencode" 2>/dev/null || true
+    fi
+
+    # Symlink ~/.local/state/opencode -> persistent state
+    mkdir -p "${home_state}"
+    if [[ -L "${home_state}/opencode" ]]; then
+        local current_target
+        current_target=$(readlink "${home_state}/opencode")
+        if [[ "${current_target}" != "${XDG_STATE_HOME}/opencode" ]]; then
+            ln -sfn "${XDG_STATE_HOME}/opencode" "${home_state}/opencode"
+        fi
+    else
+        if [[ -d "${home_state}/opencode" ]]; then
+            if [[ -n "$(ls -A "${home_state}/opencode" 2>/dev/null)" ]]; then
+                cp -rn "${home_state}/opencode/"* "${XDG_STATE_HOME}/opencode/" 2>/dev/null || true
+            fi
+            mv "${home_state}/opencode" "${home_state}/opencode.migrated" 2>/dev/null || rm -rf "${home_state}/opencode" 2>/dev/null || true
+        fi
+        ln -sfn "${XDG_STATE_HOME}/opencode" "${home_state}/opencode" 2>/dev/null || true
+    fi
+
+    echo "[entrypoint] OpenCode data directory: ${OPENCODE_DATA_DIR}"
+}
+
+# =============================================================================
+# Skills Configuration
+# =============================================================================
+# Skills are staged at /etc/opencode-skills/ (read-only ConfigMap mount)
+# and symlinked into the config directory so OpenCode discovers them.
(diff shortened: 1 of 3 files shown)


andrewdonheiser (changes requested):
## Review Summary

Nice work on the persistence and skills injection architecture — the symlink migration logic is thorough and the `envsubst` → `jq` swap is the right call. Two issues worth addressing before merge:

### Important (2)

**1. `XDG_STATE_HOME` in deployment.yaml is dead code and creates a maintenance trap** `deployment.yaml:73-77`

The manifest hardcodes `XDG_STATE_HOME=/opt/app-root/workspace/.opencode/state`, but `setup_opencode_dirs()` in entrypoint.sh unconditionally overrides it at line 21:

```bash
export XDG_STATE_HOME="${OPENCODE_DATA_DIR}/state"
```

The deployment.yaml value never takes effect. If someone later sets `OPENCODE_DATA_DIR` to a custom path, the manifest value silently diverges. Additionally, `XDG_CONFIG_HOME` and `XDG_DATA_HOME` are only exported in the entrypoint (not the manifest), creating an inconsistency across the three XDG dirs.

**Suggested fix**: Remove `XDG_STATE_HOME` from deployment.yaml (the entrypoint handles all three XDG dirs consistently). Move the 755-permission explanation into the entrypoint's `setup_opencode_dirs` function where the actual workaround lives. Alternatively, if the intent is to support `oc exec` sessions that bypass the entrypoint, add all three XDG vars to the manifest for consistency.

**2. Config file with API key written without restrictive permissions** `entrypoint.sh:157`

```bash
echo "$CONFIG" > "${XDG_CONFIG_HOME}/opencode/opencode.json"
```

This writes the full config (including `apiKey`) to persistent storage with the default umask (typically 0022 → file mode 644). Under OpenShift's random UID, other processes or containers sharing the PVC could read the file.

**Suggested fix**:
```bash
(umask 077 && echo "$CONFIG" > "${XDG_CONFIG_HOME}/opencode/opencode.json")
```
Or: `install -m 0600 /dev/null "${XDG_CONFIG_HOME}/opencode/opencode.json"` before the write.

### Suggestions (not blocking)

- **Dotglob on migration copy** (`entrypoint.sh:40,58,76`): `cp -rn ".../opencode/"*` skips dotfiles due to bash glob defaults. Use `cp -rn ".../opencode/." "..."` instead. (CodeRabbit also flagged this — agreed.)
- **Docs gap**: "Customizing the Data Directory" section in DEPLOYMENT.md mentions `config/opencode/` and `data/opencode/` but omits `state/opencode/`.

### Strengths

- Symlink migration handles edge cases well (existing symlinks, stale targets, populated dirs, empty dirs)
- `jq` substitution is correct — verified against `config-template.json` that `.provider[]` iteration matches both providers using the same placeholders
- Skills injection via ConfigMap + symlink is clean and idiomatic for OpenShift
- DEPLOYMENT.md additions are thorough and well-structured
- `set -euo pipefail` with appropriate `|| true` on non-critical paths

agents/opencode/deployment/manifests/entrypoint.sh
```diff
@@ -1,39 +1,172 @@
...
+            if [[ -n "$(ls -A "${home_state}/opencode" 2>/dev/null)" ]]; then
+                cp -rn "${home_state}/opencode/"* "${XDG_STATE_HOME}/opencode/" 2>/dev/null || true
+            fi
+            mv "${home_state}/opencode" "${home_state}/opencode.migrated" 2>/dev/null || rm -rf "${home_state}/opencode" 2>/dev/null || true
+        fi
+        ln -sfn "${XDG_STATE_HOME}/opencode" "${home_state}/opencode" 2>/dev/null || true
+    fi
+
+    echo "[entrypoint] OpenCode data directory: ${OPENCODE_DATA_DIR}"
+}
```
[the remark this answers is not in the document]
msager27: Done

msager27:
@andrewdonheiser Thanks for the review.  I've addressed your comments.

agents/opencode/deployment/manifests/entrypoint.sh
```diff
@@ -1,39 +1,145 @@
...
+# OpenCode stores config in ~/.config/opencode and session data in
+# ~/.local/share/opencode. This function redirects both to a PVC-backed
+# location so sessions persist across pod restarts.
+#
+# XDG_STATE_HOME is exported here (not in the deployment manifest) so that
+# all three XDG dirs are managed consistently. This also works around
+# the container image creating ~/.local/state/ with 755 permissions,
+# which prevents symlink creation under OpenShift's random UID.
+# =============================================================================
+redirect_xdg_dir() {
```
aakankshaduggal, line 16: The jq substitution walks all providers via `.provider[]`, so both vLLM and OGX get the same BASE_URL, API_KEY, and model. The old `envsubst` approach replaced specific `${BASE_URL}` placeholders in the template — if those were placed selectively (e.g. different endpoints per provider), this is a behavioral change. Is setting all providers to the same endpoint intentional?

agents/opencode/deployment/manifests/entrypoint.sh
```diff
@@ -1,39 +1,145 @@
...
+        fi
+    else
+        if [[ -d "${source_dir}" ]]; then
+            if [[ -n "$(ls -A "${source_dir}" 2>/dev/null)" ]]; then
+                cp -rn "${source_dir}/." "${target_dir}/" 2>/dev/null || true
+            fi
+            mv "${source_dir}" "${source_dir}.migrated" 2>/dev/null || rm -rf "${source_dir}" 2>/dev/null || true
+        fi
+        ln -sfn "${target_dir}" "${source_dir}" 2>/dev/null || true
+    fi
```
aakankshaduggal, line 35: `mv ... || rm -rf ... || true` — if `mv` fails and the earlier `cp -rn` missed any files (e.g. permission issues on individual files), the `rm -rf` fallback could lose data. Consider logging when the rm fallback is hit so it is not silent:
```suggestion
            mv "${source_dir}" "${source_dir}.migrated" 2>/dev/null || { echo "[entrypoint] Warning: could not move ${source_dir}, removing"; rm -rf "${source_dir}" 2>/dev/null || true; }
```

agents/opencode/deployment/manifests/entrypoint.sh
```diff
@@ -1,39 +1,145 @@
...
+' /config-template/config-template.json)
 
 # Merge MCP config if mounted
 if [ -f /mcp-config/mcp-servers.json ]; then
   MCP_SERVERS=$(cat /mcp-config/mcp-servers.json)
   CONFIG=$(echo "$CONFIG" | jq --argjson mcp "$MCP_SERVERS" '. + {mcp: $mcp}')
 fi
 
-export OPENCODE_CONFIG_CONTENT="$CONFIG"
 unset API_KEY BASE_URL MODEL_NAME
```
aakankshaduggal, line 133: This references `deployment/opencode-cli -c opencode` but the deployment manifest still names the deployment `opencode-web` and the container `opencode-web`. The existing DEPLOYMENT.md also uses `deployment/opencode-web` for CLI mode. Should this match?
```suggestion
    echo "[entrypoint] Attach with: oc exec -it deployment/opencode-web -c opencode-web -- opencode"
```

agents/opencode/deployment/manifests/entrypoint.sh
```diff
@@ -1,39 +1,145 @@
...
-CONFIG=$(envsubst '${BASE_URL} ${API_KEY} ${MODEL_NAME}' < /config-template/config-template.json)
+# Exclude OpenCode internal data from git tracking
+if ! grep -q "^\.opencode$" .gitignore 2>/dev/null; then
+    echo ".opencode" >> .gitignore
+fi
+
+# Build OpenCode config from template (jq handles special chars safely)
+CONFIG=$(jq --arg base "$BASE_URL" --arg key "$API_KEY" --arg model "$MODEL_NAME" '
+  .provider[].options.baseURL = $base |
+  .provider[].options.apiKey = $key |
```
aakankshaduggal: Both `model` and `small_model` get set to the same value. This works, but users may not realize they can set a lighter model for small tasks (summarization, commit messages). A comment or doc note would help — or accept `SMALL_MODEL_NAME` as an optional env var that falls back to `MODEL_NAME`.

aakankshaduggal (commented):
Solid work — session persistence and skills injection address real user pain points. A few things to double-check below.

agents/opencode/deployment/manifests/entrypoint.sh
msager27: Good catch.  Unintentional.  I think since the template used the same BASE_URL, API_KEY, etc env vars across providers, the (AI-aided) change set all providers to the same values.  Instead, we want the original behavior.

The new approach (commit coming) will replace ${VAR} placeholders individually from environment variables. This matches the original behavior — each placeholder is replaced independently, so users can set per-provider endpoints by editing the template to use different placeholder names (e.g. ${VLLM_BASE_URL}, ${OGX_BASE_URL}) and setting the corresponding env vars.  No entrypoint changes needed.

agents/opencode/deployment/manifests/entrypoint.sh
msager27: Done

agents/opencode/deployment/manifests/entrypoint.sh
msager27: Done

agents/opencode/deployment/manifests/entrypoint.sh
msager27: Same pattern as the above-mentioned fix.  The template currently uses ${MODEL_NAME} for both model and small_model.  Users who want a lighter model for small tasks can edit the template to use ${SMALL_MODEL_NAME} for small_model and set that env var.   I've added a note about this in the doc.

msager27:
@aakankshaduggal Thanks for the review.  I've addressed your comments.  Please re-review and let me know what you think.

aakankshaduggal (approved):

andrewdonheiser (approved):
👍

PR #266, opened 2026-07-15
feat(opencode): add optional SMALL_MODEL_NAME env var

msager27:
## Summary

- Add `${SMALL_MODEL_NAME}` placeholder to `config-template.json` for the `small_model` field
- Default `SMALL_MODEL_NAME` to `MODEL_NAME` in the entrypoint when not set
- Allows users to configure a lighter model for summarization, commit messages, and other quick tasks while keeping the primary model for coding and reasoning

## Test plan

- [ ] Deploy without `SMALL_MODEL_NAME` set — verify `small_model` defaults to `MODEL_NAME` in the generated config
- [ ] Deploy with `SMALL_MODEL_NAME` set to a different model — verify `small_model` reflects the override
- [ ] Run opencode tasks and verify the model responds correctly

### DIFF
--- a/agents/opencode/deployment/manifests/config-template.json
+++ b/agents/opencode/deployment/manifests/config-template.json
@@ -29,6 +29,6 @@
     }
   },
   "model": "${MODEL_NAME}",
-  "small_model": "${MODEL_NAME}",
+  "small_model": "${SMALL_MODEL_NAME}",
   "enabled_providers": ["vllm", "ogx"]
 }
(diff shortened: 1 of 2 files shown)


andrewdonheiser (approved):
Looks good — clean, minimal, and backward-compatible. LGTM.

Two optional follow-ups (not blocking):

1. **Consistency nit:** `entrypoint.sh:131` does `unset API_KEY BASE_URL MODEL_NAME` after config generation. Consider adding `SMALL_MODEL_NAME` to that line to match the cleanup pattern.

2. **Provider model registry:** The provider `models` sections in `config-template.json` only register `${MODEL_NAME}`. If someone sets `SMALL_MODEL_NAME` to a different model, it won't appear in the provider's model list. Likely fine since vLLM/OGX pass model names through to the inference endpoint, but worth confirming opencode doesn't require the small model to be explicitly registered there.

PR #267, opened 2026-07-15
fix(opencode): register small model in provider model list when set

msager27:
## Summary

- Conditionally register `SMALL_MODEL_NAME` in provider model lists when it differs from `MODEL_NAME`, so OpenCode can resolve both models
- Add `SMALL_MODEL_NAME` to env var cleanup after config generation
- Add `SMALL_MODEL_NAME` documentation to the top-level OpenCode README

## Test plan

- [ ] Deploy without `SMALL_MODEL_NAME` set — verify `small_model` defaults to `MODEL_NAME` and only one model appears in the provider models list
- [ ] Deploy with `SMALL_MODEL_NAME` set to a different model — verify both models appear in the provider models list
- [ ] Run opencode tasks and verify the model responds correctly in both cases

### DIFF
--- a/agents/opencode/README.md
+++ b/agents/opencode/README.md
@@ -79,6 +79,8 @@ deployment/overlays/mlflow-tracing/  # Overlay: MLflow tracing integration
 
 Edit `manifests/kustomization.yaml` to configure the model endpoint, API key, model name, and storage class. Apply overlays with `oc apply -k overlays/<mode>`.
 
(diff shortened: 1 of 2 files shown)


aakankshaduggal (approved):

andrewdonheiser (commented):
👍

andrewdonheiser (approved):
👍

PR #268, opened 2026-07-16
feat(google-adk): add OpenShell sandbox integration

rrbanda:
## Summary

- Add `Containerfile.openshell` for running the Google ADK agent inside an NVIDIA OpenShell sandbox
- Add `OPENSHELL.md` with build/run instructions, network policy configuration, and testing guide

Follows the official [BYOC pattern](https://github.com/NVIDIA/OpenShell/tree/main/examples/bring-your-own-container) from NVIDIA/OpenShell.

## Motivation

All other sandboxed agent deployments in this repo (Claude Code, Codex, OpenCode, OpenClaw) already have `Containerfile.openshell` files. The Google ADK agent currently deploys only as a standard FastAPI service via Helm. Adding OpenShell support enables:

- **Security isolation**: Sandboxed execution with Landlock filesystem policies and L7 network enforcement
- **Credential management**: API keys injected via OpenShell providers (never written to disk)
- **Egress control**: Only the configured LLM endpoint is reachable from within the sandbox

## Key Design Decisions

- Uses UBI9/python-312 base (same as the standard ADK Dockerfile) rather than `openshell-base` since ADK is a Python service
- Installs `iproute` + `nftables` for network namespace isolation and bypass detection
- No baked-in sandbox user (OpenShell compute driver injects UID at runtime per BYOC docs)
- `/sandbox` workdir follows OpenShell convention
- CMD documents the start command but is replaced by the supervisor at runtime

## Test plan

Tested on OpenShift 4.18 cluster with OpenShell v0.0.80 gateway:

- [x] Image builds successfully via `oc start-build` (95 packages installed)
- [x] Agent starts: `/health` returns `{"status":"healthy","agent_initialized":true}`
- [x] End-to-end chat completion with `gemini/models/gemini-2.5-flash` via Llama Stack endpoint:
  - Agent called `dummy_web_search` tool
  - Generated final answer from tool result
  - Full response: `"Red Hat OpenShift is a comprehensive platform for building, deploying, and managing applications in a hybrid cloud environment."`

### DIFF
--- a/agents/google/templates/adk/Containerfile.openshell
+++ b/agents/google/templates/adk/Containerfile.openshell
@@ -3,7 +3,8 @@
 # BYOC (Bring Your Own Container) image following the OpenShell pattern:
 # https://github.com/NVIDIA/OpenShell/tree/main/examples/bring-your-own-container
 #
-# Build:
+# Build (copy repo-level images/ into context first):
+#   cp -r ../../../../images ./images && trap 'rm -rf ./images' EXIT
 #   podman build --platform linux/amd64 -t adk-sandbox:latest -f Containerfile.openshell .
 #
 # Run (OpenShell replaces CMD -- pass the start command after --):
@@ -29,11 +30,12 @@ WORKDIR /sandbox
 # Install Python dependencies using the image's Python 3.12
 COPY --chown=1001:0 pyproject.toml .
 COPY --chown=1001:0 src/ ./src/
-RUN uv pip install --python /opt/app-root/bin/python3 --no-cache "."
+RUN uv pip install --python /opt/app-root/bin/python3 --no-cache ".[tracing]"
 
-# Application code
+# Application code and assets
 COPY --chown=1001:0 main.py .
 COPY --chown=1001:0 playground/ ./playground/
+COPY --chown=1001:0 images/ ./images/
 
 USER 1001
 
--- a/agents/google/templates/adk/Containerfile.openshell
+++ b/agents/google/templates/adk/Containerfile.openshell
@@ -1,0 +1,48 @@
+# Google ADK agent for NVIDIA OpenShell sandbox
+#
+# BYOC (Bring Your Own Container) image following the OpenShell pattern:
+# https://github.com/NVIDIA/OpenShell/tree/main/examples/bring-your-own-container
+#
+# Build:
+#   podman build --platform linux/amd64 -t adk-sandbox:latest -f Containerfile.openshell .
+#
+# Run (OpenShell replaces CMD -- pass the start command after --):
+#   openshell sandbox create --from adk-sandbox:latest --forward 8080 \
+#     -e API_KEY=<key> -e BASE_URL=<url> -e MODEL_ID=<model> \
+#     -- uvicorn main:app --host 0.0.0.0 --port 8080
+
+FROM registry.access.redhat.com/ubi9/python-312:latest
+
+USER 0
+
+# iproute: required by OpenShell for network namespace management
+# nftables: optional, enables bypass detection (log + reject for direct connections)
+RUN dnf install -y --nodocs iproute nftables && dnf clean all && rm -rf /var/cache/dnf
+
+# uv for fast dependency installation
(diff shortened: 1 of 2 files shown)


kami619:
@rrbanda the lint seems to be failing, do you have a JIRA ticket reference for this feature ?

kami619:
@rrbanda Here are some changes that are needed from the PR.,
 
 1. Important: agents/google/templates/adk/Containerfile.openshell installs only . instead of .[tracing]. The standard ADK image intentionally includes the tracing extra, and the app will raise at startup
     whenever MLFLOW_TRACKING_URI is set but mlflow is not installed.

   agents/google/templates/adk/Dockerfile lines 20-25

```
  RUN uv pip install --no-cache ".[tracing]"
  Copy the application entrypoint, playground UI, and images
  COPY main.py .
  COPY playground/ ./playground/
  COPY images/ ./images/
```
   agents/google/templates/adk/src/adk_agent/tracing.py lines 119-131
```
  tracking_uri: Optional[str] = getenv("MLFLOW_TRACKING_URI")
  if not tracking_uri:
      logger.info("[Tracing] MLFLOW_TRACKING_URI not set. Tracing is disabled.")
      return
  try:
      import mlflow
      import mlflow.litellm
  except ModuleNotFoundError as e:
      raise ModuleNotFoundError(
          "MLFLOW_TRACKING_URI is set but mlflow is not installed. "
          "Install it with: uv sync --extra tracing"
      ) from e
```
  In other words, the OpenShell image currently drops a documented runtime capability and can fail to boot for traced deployments. Fix by installing .[tracing] in the new image, or explicitly documenting
  that tracing is unsupported there.

  2. Minor: the new image copies main.py and playground/, so it still serves the / playground UI, but it does not include the repo-level images/ assets that the UI expects.

   agents/google/templates/adk/main.py lines 450-467
```
  _PLAYGROUND_HTML = _BASE_DIR / "playground" / "templates" / "index.html"
  In Docker the images are copied to /opt/app-root/src/images; locally they live at the repo root
  _IMAGES_DIR = _BASE_DIR / "images"
  if not _IMAGES_DIR.is_dir():
      _IMAGES_DIR = _BASE_DIR.parent.parent.parent / "images"
  @app.get("/", response_class=HTMLResponse, include_in_schema=False)
  async def playground():
      """Serve the playground chat UI."""
      return FileResponse(
          _PLAYGROUND_HTML,
          headers={"Cache-Control": "no-cache, no-store, must-revalidate"},
      )
  @app.get("/images/{filename:path}", include_in_schema=False)
  async def serve_image(filename: str):
```
   agents/google/templates/adk/playground/templates/index.html lines 240-243
```
  <div class="chat-container" id="chat">
      <div class="watermark">
          <img src="/images/rh_logo.svg" alt="">
      </div>
```
  As shipped, the OpenShell variant will 404 those image requests and the playground loads partially broken. Fix by copying images/ into the OpenShell image too, or make the variant API-only and omit the
  playground assets/routes.

kami619 (dismissed):
please let me know when you make these changes.

andrewdonheiser (changes requested):
## Re-Review

Both issues from the previous review are resolved — `.[tracing]` install and `images/` copy are correct. CI is fully green.

### Request: Pin the base image digest

`Containerfile.openshell:15` uses `registry.access.redhat.com/ubi9/python-312:latest`, but the standard ADK `Dockerfile` pins to a specific digest:

```dockerfile
# Current (floating)
FROM registry.access.redhat.com/ubi9/python-312:latest

# Expected (pinned, matching Dockerfile)
FROM registry.access.redhat.com/ubi9/python-312@sha256:e95978812895b9abb2bdc109b501078da2a47c8dbb9fa23758af40ed50ab6023
```

Floating `:latest` means builds can silently break when the base image updates. Please pin to the same digest as the standard Dockerfile for reproducibility.

### Everything else looks good

- BYOC pattern is correct (`iproute` + `nftables`, `/sandbox` workdir, UID 1001)
- uv digest is pinned and matches the standard Dockerfile
- `COPY --chown=1001:0` is clean — avoids the extra `chown -R` pass
- OPENSHELL.md is thorough and well-structured
- Security posture is correct (USER 0 scoped to installs only)

andrewdonheiser (approved):
👍

PR #269, opened 2026-07-16
feat(langgraph): add OpenShell sandbox integration for react_agent

rrbanda:
## TL;DR

Adds OpenShell support for the LangGraph ReAct template with one new BYOC container definition and one runbook. This does not change the existing Helm/OpenShift deployment path for `react_agent`.

## Reviewer Guide

Review in this order:

1. `agents/langgraph/templates/react_agent/Containerfile.openshell`
2. `agents/langgraph/templates/react_agent/OPENSHELL.md`

The diff is intentionally small: one file defines the image/runtime adaptation, and the second documents build, launch, policy, and verification steps.

## Summary

- Add `Containerfile.openshell` for running the LangGraph ReAct agent inside an NVIDIA OpenShell sandbox
- Add `OPENSHELL.md` with build/run instructions, network policy configuration, and verification steps

## Motivation

None of the LangGraph templates currently include an OpenShell BYOC path. Adding it to `react_agent` provides the first example on the most mature LangGraph template.

## Key Design Decisions

- Use the same pinned UBI9 Python 3.12 base digest as the standard `Dockerfile`
- Install `iproute` and `nftables` for OpenShell network namespace support and optional bypass detection
- Pre-install `components/auth` before the main package so the existing `tool.uv.sources` path does not fail from `/sandbox`
- Run Python dependency installs as UID `1001` so installed files are owned by the runtime user
- Use explicit cleanup commands in the docs so the build snippet is safe to paste into an interactive shell

## Reviewer Focus

- `Containerfile.openshell`: verify the `/sandbox` layout, dependency install order, and non-root runtime setup
- `OPENSHELL.md`: verify the build/run flow and policy example against current OpenShell CLI behavior
- Open question: the example command uses direct `-e` environment injection in `openshell sandbox create`; if we want provider-backed secret handling, that should be documented explicitly with `openshell provider create` and `--provider` rather than implied by the current wording

## Test Plan

Validated manually on OpenShift 4.18 with OpenShell v0.0.80 gateway:

- [x] Image builds successfully via `oc start-build`
- [x] `GET /health` returns `{\"status\":\"healthy\",\"agent_initialized\":true}`
- [x] `GET /images/rh_logo.svg` returns 200
- [x] `POST /chat/completions` works end-to-end with `gemini/models/gemini-2.5-flash` via a Llama Stack endpoint

No automated tests were added because this PR introduces an alternate container/deployment path plus documentation, and validation was performed as an in-cluster manual run.

### DIFF
--- a/agents/langgraph/templates/react_agent/Containerfile.openshell
+++ b/agents/langgraph/templates/react_agent/Containerfile.openshell
@@ -1,0 +1,60 @@
+# LangGraph ReAct agent for NVIDIA OpenShell sandbox
+#
+# BYOC (Bring Your Own Container) image following the OpenShell pattern:
+# https://github.com/NVIDIA/OpenShell/tree/main/examples/bring-your-own-container
+#
+# Build (copy repo-level images/ and components/auth/ into context first):
+#   cp -r ../../../../images ./images
+#   mkdir -p ./components && cp -r ../../../../components/auth ./components/auth
+#   trap 'rm -rf ./images ./components' EXIT
+#   podman build --platform linux/amd64 -t langgraph-sandbox:latest -f Containerfile.openshell .
+#
+# Run (OpenShell replaces CMD -- pass the start command after --):
+#   openshell sandbox create --from langgraph-sandbox:latest --forward 8080 \
+#     -e API_KEY=<key> -e BASE_URL=<url> -e MODEL_ID=<model> \
+#     -- uvicorn main:app --host 0.0.0.0 --port 8080
+
+FROM registry.access.redhat.com/ubi9/python-312@sha256:e95978812895b9abb2bdc109b501078da2a47c8dbb9fa23758af40ed50ab6023
+
+USER 0
+
+# iproute: required by OpenShell for network namespace management
+# nftables: optional, enables bypass detection (log + reject for direct connections)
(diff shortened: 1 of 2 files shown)


kami619:
@rrbanda 

Review Findings:

  • Medium: agents/langgraph/templates/react_agent/OPENSHELL.md misdescribes -e as provider-backed secret injection. The new guide says -e "injects environment variables via OpenShell providers (never
    written to disk)" and the comparison table repeats that mapping, but OpenShell’s current CLI docs distinguish sandbox create -e/--env from provider-backed credentials attached with --provider. That
    means the example command is documenting stronger credential-handling semantics than it actually uses, and anyone trying to follow the provider model will be sent down the wrong path. This should
    either switch to openshell provider create ... plus openshell sandbox create --provider ..., or be reworded to describe -e as direct env injection.

  • Low: agents/langgraph/templates/react_agent/Containerfile.openshell still carries the old trap 'rm -rf ./images ./components' EXIT build snippet in its header comment, even though OPENSHELL.md was
    updated to explicit cleanup after the follow-up review. That leaves two conflicting build instructions in the same PR and preserves the exact interactive-shell footgun the markdown guide was trying to
    remove.

kami619:
@rrbanda please take a look at my previous comment at your convenience.

kami619 (approved):
lgtm

PR #284, opened 2026-07-28
feat(codex): add MLflow tracing for Codex CLI on RHOAI

aakankshaduggal:
## Summary

- Adds MLflow experiment tracking layer for Codex CLI deployments on RHOAI
- Each interactive agent turn is exported as a trace to MLflow via the `notify` hook, capturing inputs, outputs, token usage, and child spans for LLM calls and tool invocations
- Uses `mlflow-codex-hook.sh` wrapper to ensure SA token auth works regardless of how codex is launched (e.g., `oc exec` bypasses `.bashrc`)
- Validated end-to-end on ROSA with Qwen3.6-27B — multi-span traces with tool call child spans

## Files

| File | Description |
|------|-------------|
| `Containerfile.mlflow` | Image layer extending base Codex with Python 3.12, Node.js 22, MLflow SDK, @mlflow/codex |
| `setup-mlflow.sh` | Post-entrypoint setup: SA token auth, experiment creation, notify hook registration, config fixups |
| `mlflow-codex-hook.sh` | Notify hook wrapper — loads SA token + experiment ID before calling @mlflow/codex |
| `deployment-mlflow-patch.yaml` | Strategic merge patch adding MLflow env vars to existing deployment |
| `mlflow-tracing.md` | Architecture docs, trace schema, env vars, known issues |
| `screenshots/vllm-codex-trace.png` | MLflow trace screenshot showing successful tool call spans |

## Key design decisions

- **TOML scoping**: `notify` key must be top-level in `config.toml` — inserting after `[section]` headers causes TOML to nest it, silently breaking hook dispatch. `setup-mlflow.sh` inserts before the first `[section]`.
- **Auth via wrapper**: `mlflow-codex-hook.sh` reads the SA token from disk on every invocation rather than relying on env vars, since `oc exec` bypasses `.bashrc`.
- **RBAC/TLS**: Uses shared `docs/mlflow-openshift-auth-and-tls.md` patterns — `mlflow-integration` ClusterRole binding + service-ca.crt for TLS.

## Dependencies

- Depends on #251 (base Codex container image) merging first

## Jira

- Epic: RHAIENG-5925 — MLflow Tracing for Codex CLI Agent Runtimes
- RHAIENG-5926 — Validate MLflow tracing with vLLM backends
- RHAIENG-5927 — Prototype trace schema for tool calls
- RHAIENG-5928 — Document tracing setup

## Test plan

- [x] Build `codex-mlflow` image on-cluster via BuildConfig
- [x] Deploy with `setup-mlflow.sh` chain, verify pod logs show MLflow setup
- [x] Run interactive codex with a coding prompt (merge sort, fibonacci, hello world)
- [x] Verify traces appear in MLflow UI with child spans (LLM + TOOL)

Generated with [Claude Code](https://claude.com/claude-code)


Nehanth:
Thanks for putting this together @aakankshaduggal! A few things to fix to align with the OpenCode and Claude Code MLflow patterns:

**1. Drop the codex-traced-exec.sh wrapper.**
It works around `codex exec` not firing notify hooks — but interactive Codex (`oc exec -it ... codex`) fires the notify hook natively and traces export automatically, no wrapper needed. That's how Claude Code and OpenCode work on RHOAI. The embedded Python JSONL parser also hardcodes Codex's transcript record shapes (response_item, input_text, session_meta), so it silently breaks if Codex changes its session format. If interactive is the primary use case, drop the wrapper and rely on the native notify hook.

**2. RBAC and TLS — reuse and extend the existing shared doc.**
There's a canonical reference at docs/mlflow-openshift-auth-and-tls.md that covers the `mlflow-integration` ClusterRole (the actual experiment/trace write permission) and proper TLS via `MLFLOW_TRACKING_SERVER_CERT_PATH` (`/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt`). Right now rbac-mlflow.yaml only grants serviceaccounts/token (not the MLflow write role), and the docs use insecure TLS (`MLFLOW_TRACKING_INSECURE_TLS=true` + `NODE_TLS_REJECT_UNAUTHORIZED=0`). Instead of a custom RBAC file + insecure TLS, point to that shared doc and extend it with any Codex-specific notes so all agents stay consistent.

**3. Missing trace screenshots.**
The PR says "validated end-to-end, 10-span traces" but no images are committed. OpenCode has screenshots/vllm-tetris-trace.png + ogx-tetris-trace.png; Claude Code has several. Add trace screenshots.

**4. Hardcoded values that won't port across clusters:**
deployment-mlflow-patch.yaml pins `MLFLOW_EXPERIMENT_ID: "54"` and the image path `hmoghani-codex/codex-mlflow` — both namespace/cluster-specific. Use placeholders, or drop the experiment ID and let it resolve from the name.

The auth token injection (MLFLOW_TRACKING_TOKEN from the SA token) and the v3.14.0 plugin build both match the OpenCode/Claude Code approach — those look good. The OpenCode and Claude Code implementations are good references for structure and depth if useful.

robinnarsinghranabhat:
Adding to what @Nehanth pointed out. 

Could we cross-reference best practices for RBAC and TLS from that global docs.
I had updated docs from `openclaw`, `claude` and `langgraph-react` agent to use same reference.
 
My hope is make navigation coherent for readers : 
- https://github.com/red-hat-data-services/agentic-starter-kits/pull/241
- https://github.com/red-hat-data-services/agentic-starter-kits/pull/226

aakankshaduggal:
Thanks Robin — already done in the latest push. Deleted the custom `rbac-mlflow.yaml`, replaced insecure TLS with `service-ca.crt` + `NODE_EXTRA_CA_CERTS`, and the RBAC/TLS section now points to `docs/mlflow-openshift-auth-and-tls.md` with Codex-specific notes. Should be consistent with what you did in #241 and #226.

agents/codex/deployment/build-mlflow/Containerfile.mlflow
```diff
@@ -0,0 +1,51 @@
...
+        tar \
+        xz \
+    && ln -sf /usr/bin/python3.12 /usr/bin/python3 \
+    && ln -sf /usr/bin/pip3.12 /usr/bin/pip3 \
+    && microdnf clean all \
+    && rm -rf /var/cache/yum
+
+# Node.js 20 (RHEL 9 ships v16, but @mlflow/codex requires >= 18)
+RUN curl -sL https://nodejs.org/dist/v20.19.2/node-v20.19.2-linux-x64.tar.xz \
+    | tar -xJ -C /usr/local --strip-components=1 \
```
tarun-etikala:
```suggestion
RUN curl -sL https://nodejs.org/dist/v22.23.1/node-v22.23.1-linux-x64.tar.xz \
    | tar -xJ -C /usr/local --strip-components=1 \
```
Node.js 20 reached end-of-life on April 30, 2026. The pinned v20.19.2 was the last release and has [known CVEs](https://www.herodevs.com/blog-posts/node-js-end-of-life-dates-you-should-be-aware-of) (CVE-2026-21637, CVE-2025-55132, CVE-2025-59465). Node.js 22 LTS (v22.23.1, Maintenance until April 2027) or Node.js 24 LTS (Active until April 2028) should be used instead.

agents/codex/deployment/build-mlflow/setup-mlflow.sh
```diff
@@ -0,0 +1,119 @@
...
+        log_info "SA token loaded from ${sa_token_path}"
+    else
+        log_warn "SA token not found at ${sa_token_path}"
+    fi
+
+    # 2. Auto-create experiment
+    python3 -c "
+import mlflow, os
+mlflow.set_tracking_uri(os.environ['MLFLOW_TRACKING_URI'])
+name = '${experiment_name}'
```
tarun-etikala: ${experiment_name} is interpolated directly into inline Python via bash string substitution. If MLFLOW_EXPERIMENT_NAME contains a single quote (e.g. codex'traces), the Python code breaks. Worse, a crafted value could inject arbitrary Python. Use os.environ inside the Python code instead:


Instead of:  name = '${experiment_name}'
Use:
name = os.environ.get('MLFLOW_EXPERIMENT_NAME', 'codex-traces')
Same pattern at line 57 where ${experiment_name} is interpolated into the experiment ID lookup.

agents/codex/deployment/build-mlflow/Containerfile.mlflow
aakankshaduggal: Addressed!

agents/codex/deployment/build-mlflow/setup-mlflow.sh
aakankshaduggal: Done

Nehanth:
Did a clean-room run of this following only the doc — build, RBAC, patches, and setup all worked. Dynamic experiment resolution and cert-path TLS both work cleanly (no `MLFLOW_TRACKING_INSECURE_TLS` needed).

One real issue: **end-to-end tracing fails silently via the documented Usage command.**

The Usage example runs codex directly (`oc exec -it deployment/codex -- codex …`), which bypasses `~/.bashrc`, so `.mlflow-env` is never sourced and `MLFLOW_TRACKING_TOKEN` isn't in codex's environment. `@mlflow/codex` doesn't read `MLFLOW_TRACKING_AUTH=kubernetes-namespaced` (Python-SDK-only), so with no token the export gets **HTTP 401 `UNAUTHENTICATED`** — but the notify hook still **exits 0**. The turn looks normal and no trace lands, with no error surfaced.

I left everything deployed per your doc in namespace **`codex-mlflow-doctest`** on the **agentic-mcp** cluster (experiment id **`59`**) so you can repro directly:

```bash
oc project codex-mlflow-doctest
POD=$(oc get pod -l app=codex -o jsonpath='{.items[0].metadata.name}')

# 1. Baseline count
oc exec $POD -- bash -lc 'python3 -c "import mlflow,os; mlflow.set_tracking_uri(os.environ[\"MLFLOW_TRACKING_URI\"]); print(len(mlflow.MlflowClient().search_traces(experiment_ids=[\"59\"],max_results=20)))"'

# 2. Fire the notify hook WITHOUT the token (what `oc exec -it -- codex …` does) → 401, exit 0
oc exec $POD -- env MLFLOW_TRACKING_TOKEN= bash -c \
  'mlflow-codex notify-hook "{\"type\":\"agent-turn-complete\",\"turn-id\":\"t1\",\"input-messages\":[\"hi\"],\"last-assistant-message\":\"4\"}"; echo EXIT=$?'
# → "HTTP 401: Unauthorized ... UNAUTHENTICATED" then EXIT=0

# 3. Same hook WITH the token sourced → exports successfully
oc exec $POD -- bash -lc \
  'source "$CODEX_HOME/.mlflow-env"; mlflow-codex notify-hook "{\"type\":\"agent-turn-complete\",\"turn-id\":\"t1\",\"input-messages\":[\"hi\"],\"last-assistant-message\":\"4\"}"; echo EXIT=$?'

# 4. Re-count → increased by 1 only from step 3
oc exec $POD -- bash -lc 'python3 -c "import mlflow,os; mlflow.set_tracking_uri(os.environ[\"MLFLOW_TRACKING_URI\"]); print(len(mlflow.MlflowClient().search_traces(experiment_ids=[\"59\"],max_results=20)))"'
```

Step 2 (no token) = 401 / exit 0 / no trace; step 3 (token) = trace lands. So the export path is solid — the gap is purely that the documented launch doesn't get the SA token into codex's env.

Two other things I hit while reproing:
- **`codex exec` doesn't fire the notify hook at all** (confirmed no trace even with the token present) — only interactive codex does.
- The "Querying Traces" curl example (`…/api/2.0/mlflow/traces/search`) returned `ENDPOINT_NOT_FOUND`; the Python `search_traces` example right below it works.

Nehanth:
Hey @aakankshaduggal! I did a full clean-room test run of this on a fresh namespace to try it out, but I couldn't get traces to show up and wanted to check whether I'm missing a step or something's breaking silently. Here's exactly what I did, command for command:

**Environment:** ROSA `agentic-mcp` cluster · fresh namespace `codex-trace-demo` · base Codex image = codex-cli **0.144.0** · backend `vllm-qwen36-27b` · MLflow at `redhat-ods-applications`.

### Test plan (followed the doc exactly)

**1. Base Codex deployment (PR #251 prereq)** — fresh namespace, api-key secret, PVC, deploy pointed at vLLM:
```bash
oc new-project codex-trace-demo
oc tag hmoghani-codex/codex:latest codex-trace-demo/codex:latest
# copy the api-key secret
KEY=$(oc get secret codex-api-key -n hmoghani-codex -o jsonpath='{.data.api-key}' | base64 -d)
oc create secret generic codex-api-key --from-literal=api-key="$KEY"
# PVC + base deployment (entrypoint.sh, backend env -> vLLM)
oc apply -f - <<'EOF'
apiVersion: v1
kind: PersistentVolumeClaim
metadata: { name: codex-workspace }
spec: { accessModes: ["ReadWriteOnce"], resources: { requests: { storage: 5Gi } } }
---
apiVersion: apps/v1
kind: Deployment
metadata: { name: codex, labels: { app: codex } }
spec:
  replicas: 1
  selector: { matchLabels: { app: codex } }
  template:
    metadata: { labels: { app: codex } }
    spec:
      containers:
        - name: codex
          image: image-registry.openshift-image-registry.svc:5000/codex-trace-demo/codex:latest
          command: ["entrypoint.sh"]
          args: ["sleep", "infinity"]
          env:
            - { name: HOME, value: /home/codex-agent }
            - { name: CODEX_HOME, value: /workspace/.codex }
            - { name: OPENAI_BASE_URL, value: http://vllm-qwen36-27b.hmoghani-codex.svc.cluster.local/v1 }
            - { name: OPENAI_MODEL, value: qwen3.6-27b }
            - { name: OPENAI_API_KEY, valueFrom: { secretKeyRef: { name: codex-api-key, key: api-key } } }
            - { name: CODEX_API_KEY, valueFrom: { secretKeyRef: { name: codex-api-key, key: api-key } } }
            - { name: CODEX_SANDBOX, value: danger-full-access }
          volumeMounts: [{ name: workspace, mountPath: /workspace }]
      volumes: [{ name: workspace, persistentVolumeClaim: { claimName: codex-workspace } }]
EOF
oc rollout status deploy/codex   # -> answers prompts fine
```

**2. Build the MLflow image** (your ImageStream + BuildConfig verbatim):
```bash
oc apply -f - <<'EOF'
apiVersion: image.openshift.io/v1
kind: ImageStream
metadata: { name: codex-mlflow }
---
apiVersion: build.openshift.io/v1
kind: BuildConfig
metadata: { name: codex-mlflow }
spec:
  source: { type: Binary }
  strategy:
    type: Docker
    dockerStrategy:
      dockerfilePath: Containerfile.mlflow
      from: { kind: ImageStreamTag, name: codex:latest }
  output: { to: { kind: ImageStreamTag, name: codex-mlflow:latest } }
EOF
cd agents/codex/deployment/build-mlflow
oc start-build codex-mlflow --from-dir=. --follow
# -> built clean; MLflow 3.15.1, `mlflow-codex --help` OK in-build
```

**3. RBAC:**
```bash
oc get clusterroles | grep mlflow-integration
oc create rolebinding codex-mlflow-integration \
  --clusterrole=mlflow-operator-mlflow-integration \
  --serviceaccount=codex-trace-demo:default
```

**4. Deployment patches** (both, `<namespace>` -> `codex-trace-demo`):
```bash
oc patch deployment codex --type=json -p '[
  {"op":"replace","path":"/spec/template/spec/containers/0/image",
   "value":"image-registry.openshift-image-registry.svc:5000/codex-trace-demo/codex-mlflow:latest"},
  {"op":"replace","path":"/spec/template/spec/containers/0/args",
   "value":["setup-mlflow.sh","sleep","infinity"]}
]'
sed 's/<namespace>/codex-trace-demo/g' deployment-mlflow-patch.yaml \
  | oc patch deployment codex --patch-file /dev/stdin --type=strategic
oc rollout status deploy/codex
```

**5. Setup looked perfect** — pod logs:
```
[mlflow-setup] SA token loaded from /var/run/secrets/kubernetes.io/serviceaccount/token
[mlflow-setup] Experiment ready: codex-traces
[mlflow-setup] mlflow-tracing.json written (experiment 60)
[mlflow-setup] Notify hook added to user-level config.toml
[mlflow-setup] MLflow tracing setup complete
```
`MLFLOW_WORKSPACE=codex-trace-demo`, `notify = ["mlflow-codex-hook.sh"]` in `config.toml`, no warnings.

**6. Usage** — your exact command:
```bash
oc exec -it deployment/codex -- codex --model qwen3.6-27b -c 'model_provider="vllm"' "explain what 2+2 equals"
# -> "2 + 2 equals 4."  waited ~3s, Ctrl-C to quit
```

**7. Check for the trace:**
```bash
oc exec deployment/codex -- python3 -c "
import mlflow, os
mlflow.set_tracking_uri(os.environ['MLFLOW_TRACKING_URI'])
print('TRACES:', len(mlflow.MlflowClient().search_traces(experiment_ids=['60'], max_results=20)))"
# -> TRACES: 0
```

### Result: turn runs fine, but **no trace ever lands** — experiment 60 stays at 0 across multiple runs.

That's my full test plan above — could you share yours? Specifically the exact steps/commands I should run to replicate a successful trace the way a customer would. Want to make sure I'm not missing a step somewhere. I've left the `codex-trace-demo` namespace up if you want to poke at it. Thanks!

Nehanth:
One more thing — in the attached screenshot (`vllm-codex-trace.png`), the `tool_exec_command` outputs actually show the tool calls failing:

> `bwrap: No permissions to create a new namespace, likely because the kernel does not allow non-privileged user namespaces...`
> `I'm unable to write files in this environment — the sandbox is read-only and shell execution is blocked.`

So it looks like codex can't actually write files / run shell in that trace. Could you please take a look and fix that so the screenshot shows a clean successful run? Thanks!

aakankshaduggal:
Thanks for the thorough clean-room test @Nehanth — that's exactly what surfaced the bug.

**Root cause:** `setup-mlflow.sh` was appending `notify = ["mlflow-codex-hook.sh"]` to the end of `config.toml`, which placed it after `[mcp_servers.*]` section headers. In TOML, any key after a `[section]` header belongs to that section — so Codex parsed it as `mcp_servers.<name>.notify` instead of the top-level `notify` it reads for hook dispatch. The hook was never registered, traces silently never landed.

Your `codex doctor` output would have shown: `Ignored unsupported project-local config keys ... notify` — that was the clue.

**Fix (f7aeb1e):**
- `setup-mlflow.sh` now inserts `notify` before the first `[section]` header via `sed`, keeping it top-level
- Also fixed: `realpath` guard for the `cp` crash when `~/.codex` is symlinked to `CODEX_HOME`, experiment ID export to `.mlflow-env`, and full-line `sed` replacement instead of prepending to existing notify arrays
- Screenshot replaced with a clean merge sort trace (all spans OK, no bwrap errors)

Your exact deploy steps are correct — the only change is in `setup-mlflow.sh` itself. Rebuilding the image with this commit and redeploying should give you traces on the first interactive run. Happy to jump on a call if you want to verify together on `codex-trace-demo`.

Nehanth:
Confirmed — the fix works end-to-end. 🎉

I deleted the namespace, pulled `6094e8f`, and re-ran the **exact same steps** on a brand-new `codex-trace-demo` (only change vs. my last run is your `setup-mlflow.sh` fix):

- `notify` is now a **top-level** key (inserted before the first `[section]` header), and `codex --strict-config` no longer flags it as an unknown field.
- Traces land on the first interactive run:
  - `explain what 2+2 equals` → 2 spans (AGENT + LLM), state OK
  - A real coding task (`write a tetris game …`) → **10-span trace** (1 AGENT + 5 LLM + 4 TOOL), every `tool_exec_command` **OK**, and it actually wrote an 8.5 KB `tetris.py` to disk. So the tool calls genuinely execute — matches your new merge-sort screenshot (`Process exited with code 0`, no bwrap error).

**Note / heads-up (base-image, not this PR):** on the very first launch I hit `config.toml:4:1: duplicate key model` and codex refused to load. The base `entrypoint.sh` re-appends the `model` / `[model_providers.vllm]` block to `config.toml` on every container start, and since `config.toml` lives on the persistent PVC, the doc's two-patch flow (which triggers multiple rollouts) accumulates duplicate keys. Clearing `/workspace/.codex` + one clean restart gets past it. This is a base-image (PR #251) thing, **not a blocker for this PR — we can address it in a follow-up** (idempotency guard in the entrypoint), maybe with a quick note in the doc here in the meantime.

Nice work tracking down the TOML scoping bug — LGTM on the tracing side.

Nehanth (approved):

tarun-etikala (approved):
LGTM — nice work addressing all the feedback. Two suggestions before merging:

**1. Update the PR description** — The Files table still lists `codex-traced-exec.sh` and `rbac-mlflow.yaml` (both removed), says "Node.js 20" (now 22), and the test plan references `codex-traced-exec.sh`. The new `mlflow-codex-hook.sh` isn't listed. Worth updating so the git history reads correctly.

**2. Source build of `@mlflow/codex` may no longer be needed** — The Containerfile comment says "npm package lacks workspace header fix, mlflow#23927" but that fix shipped in v3.14.0 (June 17), and `@mlflow/codex` v0.3.0 was [published on npm July 7](https://www.npmjs.com/package/@mlflow/codex) — after the fix. If the npm package works with RHOAI's kubernetes-namespaced auth, `npm install -g @mlflow/codex@0.3.0` would replace the entire source build block and eliminate the `git`/`make`/`gcc-c++` build dependencies, significantly reducing image size. Worth testing as a follow-up.

aakankshaduggal:
Thanks for the review @tarun-etikala!

1. **PR description** — Updated. Removed stale references to `codex-traced-exec.sh` and `rbac-mlflow.yaml`, corrected Node.js 20→22, added `mlflow-codex-hook.sh` and key design decisions section.

2. **`@mlflow/codex` npm package** — Good catch, v0.3.0 shipping the workspace header fix would let us drop the entire source build block (git, make, gcc-c++). Will test `npm install -g @mlflow/codex@0.3.0` against RHOAI's kubernetes-namespaced auth as a follow-up — if it works, that's a nice image size win.

PR #287, opened 2026-07-30
feat(agentic-rag): add OpenShell sandbox Containerfile

Wojciech-Rebisz:
## Summary
  - Add Containerfile.openshell for building the agentic RAG agent as an OpenShell sandbox image
  - Add `auth_wrapper.py` - K8s ServiceAccount token authentication via TokenReview API, protecting `/chat/completions` without modifying agent source code
  - Add `Makefile` targets for the full OpenShell sandbox lifecycle: `setup-gateway`, `build-openshell`, `deploy-openshell`, `teardown-gateway`, and individual sub-targets (`create-sandbox`, `wait-sandbox`, `setup-egress`,
  `load-docs-sandbox`, `start-agent`)
  - Add playground-sandbox/ - Flask web UI that auto-fetches the agent URL and SA token from OpenShift for interactive testing
  - Add `docs/agentic-rag-openshell-sandbox.md` -> step-by-step deployment guide
  - Security: TLS verification for TokenReview API calls using in-cluster K8s CA bundle, SA identity authorization (`ALLOWED_SA_USERNAME`), .env excluded from `oc start-build` uploads

Playground:
Screenshot 2026-08-04 at 10 01 37

Assisted with < *Claude Code* >


agents/langgraph/templates/agentic_rag/auth_wrapper.py
```diff
@@ -0,0 +1,99 @@
...
+                "apiVersion": "authentication.k8s.io/v1",
+                "kind": "TokenReview",
+                "spec": {"token": token},
+            },
+            headers={
+                "Authorization": f"Bearer {_K8S_REVIEWER_TOKEN}",
+                "Content-Type": "application/json",
+            },
+            verify=False,
+            timeout=10,
```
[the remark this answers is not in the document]
Wojciech-Rebisz: Is that fixed? check the other comments to see if they are still valid

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -182,6 +182,199 @@ dry-run: _check-env ## Render Helm templates without deploying
...
+	  echo "==> Installing openShell gateway $(OPENSHELL_VERSION)..." && \
+	  helm install openshell oci://ghcr.io/nvidia/openshell/helm-chart \
+	    --version $(OPENSHELL_VERSION) \
+	    --namespace openshell \
+	    --set securityContext.runAsUser=null \
+	    --set securityContext.runAsNonRoot=false \
+	    --set securityContext.allowPrivilegeEscalation=true \
+	    --set securityContext.capabilities=null \
+	    --set server.auth.allowUnauthenticatedUsers=true \
+	    --set "pkiInitJob.serverDnsNames={*.openshell.$$APPS_DOMAIN,openshell.$$APPS_DOMAIN,openshell-gateway-openshell.$$APPS_DOMAIN}" && \
```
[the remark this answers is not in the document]
Wojciech-Rebisz: Won't fix — privileged SCC is required by the OpenShell sandbox runtime

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -182,6 +182,199 @@ dry-run: _check-env ## Render Helm templates without deploying
...
+	  GATEWAY_URL="https://openshell-gateway-openshell.$$APPS_DOMAIN" && \
+	  echo "==> Registering gateway in CLI..." && \
+	  openshell gateway remove openshift 2>/dev/null || true && \
+	  MTLS_DIR=~/.config/openshell/gateways/openshift/mtls && mkdir -p $$MTLS_DIR && \
+	  oc -n openshell get secret openshell-client-tls -o jsonpath='{.data.ca\.crt}'  | base64 -d > $$MTLS_DIR/ca.crt && \
+	  oc -n openshell get secret openshell-client-tls -o jsonpath='{.data.tls\.crt}' | base64 -d > $$MTLS_DIR/tls.crt && \
+	  oc -n openshell get secret openshell-client-tls -o jsonpath='{.data.tls\.key}' | base64 -d > $$MTLS_DIR/tls.key && \
+	  openshell gateway add "$$GATEWAY_URL" --local --name openshift && \
+	  echo "==> Verifying connection..." && \
+	  export OPENSHELL_GATEWAY_INSECURE=true && \
```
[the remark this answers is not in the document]
Wojciech-Rebisz: Won't fix — the OpenShell gateway uses a self-signed cert from the PKI init job behind a passthrough Route.

docs/agentic-rag-openshell-sandbox.md
```diff
@@ -0,0 +1,182 @@
...
+curl -sk "$AGENT_URL/chat/completions" -X POST \
+  -H "Content-Type: application/json" \
+  -d '{"messages":[{"role":"user","content":"hello"}]}'
+# Output (error):
+# {"error": "Missing API key (use X-Api-Key or Authorization: Bearer header)"}
+```
+
+```bash
+# With SA token — RAG query
+curl -sk "$AGENT_URL/chat/completions" -X POST \
```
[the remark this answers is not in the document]
Wojciech-Rebisz: See comments above

Wojciech-Rebisz:
Hi there, this PR is `size/xl` mainly due to the playground web UI (files copied from playground and modified a little) which includes HTML templates CSS etc, and the step-by-step deployment guide `docs/agentic-rag-openshell-sandbox.md`. The core logic is relatively small ->  auth_wrapper.py, Containerfile.openshell, and new Makefile targets.

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -220,6 +220,212 @@ dry-run: _check-env ## Render Helm templates without deploying
 undeploy: ## Remove deployment from cluster
 	helm uninstall $(AGENT_NAME)
 
+## ── openShell sandbox targets ─────────────────────────────────────────────────
+OPENSHELL_SANDBOX ?= $(shell . ./.env 2>/dev/null && echo $${OPENSHELL_SANDBOX_NAME:-rag-sandbox})
+OPENSHELL_IMAGE   := openshell-agentic-rag
+OPENSHELL_VERSION ?= 0.0.80
```
jwm4: The AgentOps team's [opencode-vertex-tracing](https://github.com/r3v5/agent-ops/tree/opencode-in-openshell-with-mlflow-on-openshift-demo/demos/opencode-vertex-tracing) demo uses `0.0.85`. Worth checking what the current release is and updating the default.

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -220,6 +220,212 @@ dry-run: _check-env ## Render Helm templates without deploying
...
+
+setup-gateway: ## Install openShell gateway, create Route, register CLI (one-time)
```
jwm4, line 228: This installs OpenShell as the default StatefulSet, then patches `gateway.toml` with `sed` to remove `client_ca_path` and deletes `openshell-0` to restart. The AgentOps team takes a different approach: Deployment mode with PostgreSQL and `server.disableTls=true`, which avoids the post-install configmap patching. I don't have strong feelings either way, just noting the discrepancy in case the AgentOps team's approach was chosen for reasons worth adopting here too.

agents/langgraph/templates/agentic_rag/auth_wrapper.py
```diff
@@ -0,0 +1,112 @@
...
+
+_AUTH_ENABLED = bool(_K8S_API_URL and _K8S_REVIEWER_TOKEN)
+
+
+async def _validate_k8s_token(token: str) -> bool:
+    if not (_K8S_API_URL and _K8S_REVIEWER_TOKEN):
+        return False
+    try:
+        tls_verify: str | bool = _K8S_CA_PATH if Path(_K8S_CA_PATH).is_file() else True
+        async with httpx.AsyncClient(verify=tls_verify, timeout=10.0) as client:
```
jwm4: This creates a new `httpx.AsyncClient` for each token validation request, meaning every call to `/chat/completions` opens a new TCP connection and TLS handshake to the K8s API. A module-level or cached client with connection pooling would be more efficient. I recognize this is a demo and doesn't need to work at scale, but if part of the purpose is to teach best practices for someone who will build something that does need to scale, this is probably worth addressing.

agents/langgraph/templates/agentic_rag/playground-sandbox/app.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+CURRENT_NAMESPACE = get_current_namespace()
+AGENT_URL = get_agent_url()
+AGENT_TOKEN = get_agent_token()
+AGENT_CA_BUNDLE = getenv("AGENT_CA_BUNDLE", "")
+if AGENT_CA_BUNDLE and Path(AGENT_CA_BUNDLE).is_file():
+    VERIFY_TLS = AGENT_CA_BUNDLE
+else:
+    VERIFY_TLS = False
```
jwm4, line 151: When `AGENT_CA_BUNDLE` is not set (the default), TLS verification is disabled for all requests to the agent, with only a `logger.warning()` that's easy to miss in startup logs. Since part of the value of this demo is teaching secure sandbox deployment practices, it might be worth either making TLS verification the default (requiring an explicit opt-out) or making the warning more prominent. The AgentOps team sidesteps this by using `oc port-forward` for local access, so TLS isn't needed.

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -220,6 +220,212 @@ dry-run: _check-env ## Render Helm templates without deploying
...
+	  CLIENT_TOKEN=$$(oc create token agent-client --duration=168h) && \
+	  oc delete secret agent-client-token 2>/dev/null || true && \
+	  oc create secret generic agent-client-token --from-literal=token="$$CLIENT_TOKEN" && \
+	  echo "==> Generating sandbox reviewer token..." && \
+	  K8S_API_URL=$$(oc whoami --show-server) && \
+	  K8S_REVIEWER_TOKEN=$$(oc create token openshell-sandbox -n openshell --duration=168h) && \
+	  ALLOWED_SA_USERNAME="system:serviceaccount:$$NS:agent-client" && \
+	  echo "==> Starting agent (VECTOR_STORE_ID=$$VSID)..." && \
+	  (openshell sandbox exec -n "$$SANDBOX" \
+	    --env "BASE_URL=$$BASE_URL" \
```
jwm4: Env vars are passed via `openshell sandbox exec --env` on each command. The AgentOps team passes them via `openshell sandbox create --env`, which persists them in the sandbox. With the per-exec approach, if someone enters the sandbox interactively with `openshell term` to debug, the env vars won't be present.

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -220,6 +220,212 @@ dry-run: _check-env ## Render Helm templates without deploying
...
+	    --env "EMBEDDING_DIMENSION=$${EMBEDDING_DIMENSION:-1024}" \
+	    --env "VECTOR_STORE_ID=$$VSID" \
+	    --env "VECTOR_STORE_PROVIDER=$$VECTOR_STORE_PROVIDER" \
+	    --env "API_KEY=$$API_KEY" \
+	    --env "K8S_API_URL=$$K8S_API_URL" \
+	    --env "K8S_REVIEWER_TOKEN=$$K8S_REVIEWER_TOKEN" \
+	    --env "ALLOWED_SA_USERNAME=$$ALLOWED_SA_USERNAME" \
+	    --env "PORT=8080" \
+	    --workdir /sandbox \
+	    -- uvicorn auth_wrapper:app --host 127.0.0.1 --port 8080 >/dev/null 2>&1 &) && \
```
jwm4, line 389: This runs uvicorn in the background with `&` inside the sandbox with no restart mechanism or supervision. If this demo is meant to teach deployment best practices, consider adding a process supervisor or at minimum a health-check retry loop. Otherwise, a comment in the Makefile or docs describing what a production setup would look like (e.g. a process supervisor, or relying on Kubernetes restart policies) would help readers bridge the gap from demo to production.

jwm4 (changes requested):
Good addition. Reviewed against the AgentOps team's [opencode-vertex-tracing](https://github.com/r3v5/agent-ops/tree/opencode-in-openshell-with-mlflow-on-openshift-demo/demos/opencode-vertex-tracing) demo, which is the reference implementation for OpenShell on our product. A few notes below, mostly around alignment with that reference. Comment from Bill Murdock with assistance from Claude Code.

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -220,6 +220,212 @@ dry-run: _check-env ## Render Helm templates without deploying
...
+	  echo "    Waiting 10s for agent startup..." && \
+	  sleep 10 && \
+	  HEALTH=$$(openshell sandbox exec -n "$$SANDBOX" -- curl -sf http://127.0.0.1:8080/health 2>/dev/null) && \
+	  if [ -z "$$HEALTH" ]; then echo "ERROR: agent health check failed"; exit 1; fi && \
+	  echo "    $$HEALTH" && \
+	  echo "==> Exposing service URL..." && \
+	  openshell service expose "$$SANDBOX" 8080 agent && \
+	  SVC_HOST="$$SANDBOX--agent.openshell.$$APPS_DOMAIN" && \
+	  printf 'apiVersion: route.openshift.io/v1\nkind: Route\nmetadata:\n  name: openshell-rag-agent\n  namespace: openshell\nspec:\n  host: "%s"\n  to:\n    kind: Service\n    name: openshell\n  port:\n    targetPort: 8080\n  tls:\n    termination: passthrough\n' "$$SVC_HOST" | oc apply -f - && \
+	  echo "" && \
```
andrewdonheiser, line 399: **Blocking: Route hostname missing `default--` namespace prefix**

The route is constructed as `$SANDBOX--agent.openshell.$APPS_DOMAIN` but `openshell service expose` generates URLs with a `default--` prefix: `default--$SANDBOX--agent.openshell.$APPS_DOMAIN`. The agent deploys and is healthy but is completely unreachable via this URL.

Confirmed during manual deployment testing — all requests returned "Service endpoint not found" from the openshell gateway until the route was manually deleted and recreated with `default--rag-sandbox--agent.openshell.<apps-domain>`.

**Fix**: Capture the actual hostname from `openshell service expose` output, or prepend `default--` to the hostname construction.

docs/agentic-rag-openshell-sandbox.md
```diff
@@ -0,0 +1,229 @@
...
+# openshell-rag-agent   rag-sandbox--agent.openshell.apps.rosa.example.com         openshell   8080   passthrough   None
+```
+
+## Step 5 — Test
+
+```bash
+AGENT_URL="https://rag-sandbox--agent.openshell.$(oc get ingresses.config cluster -o jsonpath='{.spec.domain}')"
+
+# Get the auto-generated token from the Secret
+TOKEN=$(oc get secret agent-client-token -o jsonpath='{.data.token}' | base64 -d)
```
andrewdonheiser, line 118: **Blocking: Doc URL pattern also missing `default--` prefix**

Same issue as the Makefile — this URL should be `https://default--rag-sandbox--agent.openshell.$(...)`. A user copy-pasting from the doc will get "Service endpoint not found" even if the Makefile route is fixed, because this URL is constructed independently.

The `oc get route` example output on line 109 also shows the wrong hostname (`rag-sandbox--agent.openshell.apps.rosa.example.com` instead of `default--rag-sandbox--agent...`).

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -220,6 +220,212 @@ dry-run: _check-env ## Render Helm templates without deploying
...
+	  oc adm policy add-scc-to-user privileged -z openshell-sandbox -n openshell 2>/dev/null || true && \
+	  oc adm policy add-scc-to-user privileged -z openshell -n openshell 2>/dev/null || true && \
+	  echo "==> Granting TokenReview RBAC to sandbox SA..." && \
+	  oc create clusterrolebinding openshell-tokenreview \
+	    --clusterrole=system:auth-delegator \
+	    --serviceaccount=openshell:openshell-sandbox 2>/dev/null || true && \
+	  echo "==> Installing openShell gateway $(OPENSHELL_VERSION)..." && \
+	  helm install openshell oci://ghcr.io/nvidia/openshell/helm-chart \
+	    --version $(OPENSHELL_VERSION) \
+	    --namespace openshell \
```
andrewdonheiser, line 243: **`helm install` is not idempotent — re-runs silently break**

If the helm release already exists (common when re-running or debugging), `helm install` fails with "cannot reuse a name that is still in use." The `&&`-chained steps after it (mTLS disable, route creation, pod restart) are silently skipped, but the final `echo "Gateway ready"` still prints because it's past the `&&` chain.

**Fix**: Use `helm upgrade --install` instead of `helm install`.

docs/agentic-rag-openshell-sandbox.md
```diff
@@ -0,0 +1,229 @@
...
+| Variable | Example | Description |
+|---|---|---|
+| `API_KEY` | `eyJhbG...` | OGX API key (JWT from Keycloak) |
+| `BASE_URL` | `https://server-ogx.<apps-domain>/v1` | OGX endpoint |
+| `MODEL_ID` | `maas-llm/qwen3-8b-fp8-dynamic` | LLM model |
+| `EMBEDDING_MODEL` | `maas-embedding/redhataibge-m3` | Embedding model (must be in OGX `allowed_models`) |
+| `VECTOR_STORE_PROVIDER` | `milvus-remote` | Vector store backend |
+
+## Step 2 — Install openShell gateway and connect CLI
+
```
andrewdonheiser, line 36: **`.env.example` / doc mismatch for `VECTOR_STORE_PROVIDER`**

This table shows `milvus-remote` as the example value, but `.env.example` line 18 has `VECTOR_STORE_PROVIDER=milvus`. A user following the docs would set a different value than the template default. These should be consistent — either update the doc or the `.env.example`.

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -220,6 +220,212 @@ dry-run: _check-env ## Render Helm templates without deploying
...
+	  oc -n openshell get secret openshell-client-tls -o jsonpath='{.data.tls\.crt}' | base64 -d > $$MTLS_DIR/tls.crt && \
+	  oc -n openshell get secret openshell-client-tls -o jsonpath='{.data.tls\.key}' | base64 -d > $$MTLS_DIR/tls.key && \
+	  openshell gateway add "$$GATEWAY_URL" --local --name openshift && \
+	  echo "==> Verifying connection..." && \
+	  export OPENSHELL_GATEWAY_INSECURE=true && \
+	  openshell status && \
+	  echo "" && \
+	  echo "Gateway ready. Next: make build-openshell"
+
+teardown-gateway: ## Uninstall openShell gateway and clean up
```
andrewdonheiser, line 275: **False success message when preceding commands fail**

If `helm install` or any of the `&&`-chained steps above fail, this line still executes and prints "Gateway ready." The user sees success and moves to `make build-openshell`, only to hit failures later.

Same pattern occurs in `build-openshell` at line 294 ("Image ready").

**Fix**: Either wrap the entire recipe body in a single `&&` chain so the echo only runs on success, or use `set -e` at the start of the recipe (already using `set -e` in `setup-gateway` but it's after the problematic echo).

agents/langgraph/templates/agentic_rag/Makefile
```diff
@@ -220,6 +220,212 @@ dry-run: _check-env ## Render Helm templates without deploying
...
+	  helm install openshell oci://ghcr.io/nvidia/openshell/helm-chart \
+	    --version $(OPENSHELL_VERSION) \
+	    --namespace openshell \
+	    --set securityContext.runAsUser=null \
+	    --set securityContext.runAsNonRoot=false \
+	    --set securityContext.allowPrivilegeEscalation=true \
+	    --set securityContext.capabilities=null \
+	    --set server.auth.allowUnauthenticatedUsers=true \
+	    --set "pkiInitJob.serverDnsNames={*.openshell.$$APPS_DOMAIN,openshell.$$APPS_DOMAIN,openshell-gateway-openshell.$$APPS_DOMAIN}" && \
+	  echo "==> Waiting for gateway to be ready..." && \
```
andrewdonheiser, line 250: **Minor: `allowUnauthenticatedUsers=true` on gateway is worth documenting**

This makes the gateway world-accessible without authentication. The auth_wrapper protects individual endpoints inside the sandbox, so the overall security posture is fine for a demo. But it's worth a comment in the docs (e.g., in the "How auth works" section) noting this is intentional — otherwise someone auditing the deployment may flag it.

andrewdonheiser (changes requested):
Tested this PR end-to-end on the `agen-e2e-rhoai2` cluster (ROSA HCP, openshell CLI v0.0.97). The auth architecture is solid and the agent worked correctly once the route hostname was fixed. Notes below focus on areas not already covered by Bill's review.

**Blocking issues** are the route hostname mismatch and the doc URL pattern — without fixing those, no user following the docs can reach the deployed agent. The rest are important but non-blocking.

Review from Andrew Donheiser with assistance from Claude Code.

agents/langgraph/templates/agentic_rag/Makefile
Wojciech-Rebisz: Updated to 0.0.99, as that looks like latest version.

agents/langgraph/templates/agentic_rag/Makefile
Wojciech-Rebisz: Good point, and well, I chose statefulset because its the upstream default and doesn't require a postgres dependency. I will think if we can improve it to align with AgentOps team, leaving it as open for now..

agents/langgraph/templates/agentic_rag/auth_wrapper.py
Wojciech-Rebisz: I implemented module-level client with lazy init as suggested.

agents/langgraph/templates/agentic_rag/playground-sandbox/app.py
Wojciech-Rebisz: So I decided to add more "visible" warning. As for port forwarding, I wanted to avoid it at all costs -> my goal was for users to be able to launch the agent using just 3–4 commands, without needing additional terminals.

agents/langgraph/templates/agentic_rag/Makefile
Wojciech-Rebisz: Static env vars moved to `sandbox create --env`

agents/langgraph/templates/agentic_rag/Makefile
Wojciech-Rebisz: Added a note to the docs explaining this limitation and the restart workaround. A process supervisor might be a little complicated -> the sandbox itself is ephemeral and `make start-agent` takes only 10 seconds.

agents/langgraph/templates/agentic_rag/Makefile
Wojciech-Rebisz: Prefix `default--` added to hostname construction.

docs/agentic-rag-openshell-sandbox.md
Wojciech-Rebisz: Added `default--` to urls as well.

agents/langgraph/templates/agentic_rag/Makefile
Wojciech-Rebisz: After changes that one should use `helm upgrade --install`

docs/agentic-rag-openshell-sandbox.md
Wojciech-Rebisz: Fair point, my `ogx` setup was using `milvus-remote`, and that was the problem. I have updated it to use just `milvus`.

agents/langgraph/templates/agentic_rag/Makefile
Wojciech-Rebisz: Sure, I suspect that fix with `helm upgrade --install` will help here, so it won't fail on re-run.

agents/langgraph/templates/agentic_rag/Makefile
Wojciech-Rebisz: Documented in the `how auth works` section.

agents/langgraph/templates/agentic_rag/Makefile
Wojciech-Rebisz: Is that okey?

andrewdonheiser (approved):

jwm4 (approved):
My concerns are all addressed now, so this looks good to me.

PR #288, opened 2026-07-30
fix(opencode): fix stale mlflow-tracing overlay and update Containerfile base image

msager27:
## Summary

- Fix `overlays/mlflow-tracing/entrypoint-patch.yaml`: the patch was stale and
  contained only the MLflow-specific portion of `entrypoint.sh`. Kustomize
  strategic merge on a ConfigMap string value replaces the entire value, so the
  patch must carry the full current base entrypoint. Updated to include the
  complete base entrypoint with the MLflow tracing block inserted before the
  API key unset.
- Fix `Containerfile.mlflow` base image: `quay.io/opendatahub/odh-opencode-rhel9:1.4.4`
  no longer exists in the registry. Updated to `v1.18.3` (OpenCode 1.18.3, July 2026).

## Test plan

- [ ] Build `Containerfile.mlflow` — verify the build succeeds with the `v1.18.3` base image
- [ ] Deploy with the `mlflow-tracing` overlay — verify the pod starts with the full
      entrypoint (XDG dir setup, skills config, and MLflow tracing block all present)
- [ ] With `MLFLOW_TRACKING_URI` and `MLFLOW_EXPERIMENT_NAME` set, verify the plugin
      installs and the experiment ID is resolved on startup
- [ ] Run opencode and verify traces appear in the MLflow UI

### DIFF
--- a/agents/opencode/deployment/Containerfile.mlflow
+++ b/agents/opencode/deployment/Containerfile.mlflow
@@ -4,7 +4,7 @@
 # - mlflow[kubernetes] for experiment creation and SA token auth
 # - Pre-built @mlflow/core from v3.14.0 (workspace header fix)
 
-FROM quay.io/opendatahub/odh-opencode-rhel9:1.4.4
+FROM quay.io/opendatahub/odh-opencode-rhel9:v1.18.3
 
 USER 0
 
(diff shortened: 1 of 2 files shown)


aakankshaduggal (approved):
LGTM, thanks!

tarun-etikala (approved):
LGTM — the fix is correct and well-scoped.

Verified that the new patch carries a byte-identical copy of the current base `entrypoint.sh` with only the MLflow tracing block inserted. Base image `v1.18.3` confirmed present in the registry.

**One suggestion for follow-up:** Consider adding a CI check or Makefile target that diffs the base entrypoint against the non-MLflow portions of this patch to catch future drift — this is the same class of staleness bug this PR fixes.

Nehanth (approved):

PR #289, opened 2026-07-30
feat(guardrails): layer NIM safety models per guardrail role in guardrailed_agent

kami619:
## Summary

Splits `guardrailed_agent`'s NeMo Guardrails config into two profiles under `guardrails/config/`, matching two maturity levels of the same rail layering:

- **`local`** — single "main" model, NeMo Guardrails' built-in self-check rails (self_check_input/output). Zero extra setup; good for demoing the proxy pattern (`make guardrails-server-local`).
- **`nemoguard`** — layered `content_safety`/`topic_control` rails, each backed by a purpose-built NemoGuard/Nemotron model, with per-role model/endpoint/key/engine overrides (`MAIN_*`, `CONTENT_SAFETY_*`, `TOPIC_CONTROL_*`) and an `engine=nim` path via `ChatNVIDIA` for NVIDIA's NIM-hosted models (`make guardrails-server-nemoguard`).

The dedicated `nvidia/llama-3.1-nemoguard-8b-topic-control` NIM is currently returning 500s (TensorRT-LLM CUDA crash) on NVIDIA's hosted free tier, and its would-be successor reached end-of-life the same day it was tested. Wires `topic_control` to `nvidia/nemotron-3.5-content-safety` instead, which classifies against a free-text policy rather than NeMo Guardrails' built-in on-topic/off-topic prompt -- needs its own action/flow (`actions.py`, `topic_policy.co`), auto-selected by `generate_config.py` based on the configured model id.

Also addresses findings from two independent code reviews of the above:

- Fixes ambient-env leakage in `test_generate_config.py` via an autouse env-clearing fixture, so ambient shell state (e.g. a previously-sourced `.env`) can't affect test assertions.
- Guards the `topic_control` custom-policy flow swap against silently no-op'ing if the expected built-in flow entry isn't found.
- Hardens `actions.py`'s topic-policy verdict parsing to look for an explicit `User Safety: safe|unsafe` field (instead of a raw substring check) and fail closed on unparseable output.
- Adds `tests/test_guardrails_smoke.py`, loading both profiles through NeMo Guardrails' real `RailsConfig`/`LLMRails` (no LLM calls) to catch rails/action/prompt wiring bugs that structural YAML checks alone would miss.
- Documents (in `generate_config.py`, `.env.example`, and the README) that `MAIN_MODEL_ID` has no effect on real proxied traffic -- verified against the installed `nemoguardrails==0.21.0` source: its server always overrides the "main" model's id from the client's OpenAI `model` field, which this agent always sets to `MODEL_ID`.
- Minor: DRYs the per-role engine override helper, wraps malformed `config.yaml` errors with file/role context, fixes an inaccurate `.env.example` comment, and other small polish.

## JIRA
https://redhat.atlassian.net/browse/RHAIENG-6263

## Test plan

- [x] `make test` (63 tests, including the new `RailsConfig`/`LLMRails` smoke tests for both profiles) passes
- [x] `ruff check` / `ruff format --check` clean
- [x] All pre-commit hooks pass
- [x] Manually verified `make guardrails-server-local` and `make guardrails-server-nemoguard` generate valid configs and start the proxy


tarun-etikala (approved):
LGTM. Thanks Kamesh!

PR #303, opened 2026-08-05
fix(btest): use per-agent MLflow experiment names in test runner (RHAIENG-6743)

andrewdonheiser:
## Summary

https://redhat.atlassian.net/browse/RHAIENG-6743

- Fix `run-btests-pytest.sh` to detect per-agent `MLFLOW_EXPERIMENT_NAME` from each agent's deployment instead of using a single global value from the first deployment

## Root cause

All agents in `adonheis-testing` shared `MLFLOW_EXPERIMENT_NAME=adonheis-testing`. `MLflowTraceClient.get_latest_trace()` queries by `experiment_id` only, so it returned traces from whichever agent ran most recently — not the one being tested. This caused `tool_selection_accuracy` cross-contamination across all agents.

## Test plan

- [x] Deployed all 11 agents with per-agent experiment names (`adonheis-testing/<agent-name>`)
- [x] Verified `[Tracing Enabled]` with unique experiment names in all agent startup logs
- [x] Ran full btest suite: 188 passed, 16 failed — zero cross-contamination


shricharan-ks (approved):
## Code Review — PR #303

Clean, well-targeted fix for MLflow trace cross-contamination in the btest runner.

**Root cause confirmed**: Line 269 discovers `MLFLOW_EXPERIMENT_NAME` from the first deployment and sets it globally. Line 353 (old) exports this stale global for ALL agents, so `MLflowTraceClient.get_latest_trace()` returns traces from whichever agent ran most recently — not the one being tested.

**Fix is correct**: The new code re-detects the experiment name from the current agent's specific deployment inside `run_tests()`:
```bash
agent_experiment=$(timeout 30 oc get deployment "${deploy}" -n "${NAMESPACE}" \
  -o jsonpath='{...MLFLOW_EXPERIMENT_NAME...}' 2>/dev/null || true)
agent_experiment="${agent_experiment:-${MLFLOW_EXPERIMENT_NAME:-}}"
```

**Error handling**: `timeout 30` + `|| true` + fallback to global value — gracefully handles missing deployments or timeout.

**Verified**: `${deploy}` comes from `agent_deploy "$agent_tuple"` — the correct per-agent deployment name. The `oc get deployment` pattern matches the existing discovery at line 269 but is scoped to the current agent.

LGTM — ready to merge.

PR #305, opened 2026-08-06
ci: prevent CI Health Pages deploys from cancelling each other

tarun-etikala:
## Problem

The `CI Health Pages` workflow intermittently fails with `Deployment cancelled.` — e.g. [run 31105557581](https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/31105557581).

Root cause: the workflow shares a static `pages` concurrency group with `cancel-in-progress: true`. When two eligible runs overlap — which happens when multiple monitored CI workflows (Code Quality, Agent Tests, etc.) finish close together on a `main` push — the newer run cancels the in-progress one mid-deploy, and the newer run then fails at its own `actions/deploy-pages` step with `Deployment cancelled.`

Observed collision:

| Run | Created | Ended | Result |
|-----|---------|-------|--------|
| [31105455073](https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/31105455073) | 13:20:29 | 13:22:23 | cancelled |
| [31105557581](https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/31105557581) | 13:21:49 | 13:23:00 | failure |

Impact is cosmetic only: the CI dashboard page occasionally doesn't refresh. No product or deployment pipeline is affected, and the next successful run self-heals the page.

## Fix

Switch to GitHub's recommended Pages concurrency pattern (`cancel-in-progress: false`) so an in-progress deployment always completes and later runs queue instead of cancelling each other.


shricharan-ks (approved):
## Code Review — PR #305

One-line CI fix: `cancel-in-progress: true` → `false` for the Pages concurrency group.

**Root cause is well-documented** — two overlapping CI Health Pages runs cancel each other mid-deploy when multiple monitored workflows finish close together on `main`. Specific failing run IDs provided.

**Fix is correct** — matches [GitHub's recommended Pages concurrency pattern](https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages). With `cancel-in-progress: false`, an in-progress deployment completes and later runs queue instead of cancelling each other.

**No downside** — the only trade-off is slightly longer queue times when runs overlap, which is negligible for a dashboard page that takes ~2 minutes to deploy.

LGTM — ready to merge.

kami619:
great work and nice quality of improvement @tarun-etikala thank you.

PR #313, opened 2026-08-12
fix(opencode): remove Kagenti references from A2A deployment docs and template

msager27:
## Summary

- Kagenti was used only as a test vehicle for A2A agent discovery and is
  not part of RHOAI — remove all references while keeping the A2A
  integration intact
- Rename `kagenti-agent.yaml` → `opencode-a2a-template.yaml`; remove the
  `AgentRuntime` CRD object (requires Kagenti installed) and
  `protocol.kagenti.io/a2a` labels (Kagenti watch labels) from the
  OpenShift Template — the Service, Deployment, probes, env vars, and
  security context are unchanged
- Drop the "Enable Kagenti Discovery" section (Helm-based namespace
  enrollment), "Check Kagenti UI" verification step, and "Agent not
  appearing in Kagenti UI" troubleshooting from `README-a2a.md`
- Replace Kagenti-specific "Check AgentRuntime and AgentCard Creation"
  verification block with standard `oc get deployment/pods` commands
- Update `entrypoint-a2a.sh` comments: "Kagenti standard names" →
  neutral `LLM_*` variable descriptions (no logic changes)
- Update `Containerfile.a2a` comment (no build changes)
- Update variant table, TOC, section heading, directory tree, and links
  in `agents/opencode/README.md`

## Test plan

- [ ] `oc process -f opencode-a2a-template.yaml -p NAMESPACE=... -p LLM_API_BASE=... -p LLM_MODEL=... | oc apply -f -` applies cleanly (no unknown CRD errors)
- [ ] Pod starts and readiness probe passes (`GET /.well-known/agent-card.json` returns 200)
- [ ] `curl http://opencode-a2a.<namespace>.svc.cluster.local:8000/.well-known/agent-card.json` returns a valid agent card
- [ ] `cat ~/.config/opencode/opencode.json` inside the pod shows correct provider config from `LLM_*` env vars
- [ ] `grep -r kagenti agents/opencode/` returns no matches

### DIFF
--- a/agents/opencode/README.md
+++ b/agents/opencode/README.md
@@ -23,7 +23,7 @@ Deploy [OpenCode](https://opencode.ai), an open-source terminal-based coding age
   - [View Traces](#view-traces)
   - [Key Notes](#key-notes)
   - [OpenShell + MLflow Tracing](#openshell--mlflow-tracing)
-- [A2A / Kagenti Discovery](#a2a--kagenti-discovery)
+- [A2A Discovery](#a2a-discovery)
 - [Security](#security)
 - [Architecture](#architecture)
 - [Image Reference](#image-reference)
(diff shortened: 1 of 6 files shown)


Nehanth (approved):

tarun-etikala (approved):
lgtm 

**Minor docs suggestion:** could we drop the RHAIENG-* ticket IDs from the user-facing READMEs and describe capability status in plain language instead? Jira is public, but external readers can't easily jump to the ticket without a URL. Something like “A2A Protocol execution is not yet implemented” would be clearer.

PR #315, opened 2026-08-17
fix(evals): remove EvalHub adapter from behavioral testing (RHAIENG-6979)

andrewdonheiser:
## Summary
- Removes the `evals/evalhub_adapter/` package (adapter, config, evaluations, Containerfile, run-e2e.sh, and all tests)
- Deletes per-agent `evalhub/tool_use.yaml` fixtures from all 12 agents (pytest uses separate `golden_queries.yaml`)
- Removes `eval-hub-sdk` dependency from `pyproject.toml` and the `evalhub` optional dep group
- Restructures `eval-gating.yml` workflow: removes the `deterministic` job, inlines cluster-btests config logic
- Removes `docs/adding-evalhub-agent-integration.md` and updates README/CONTRIBUTING references

The evaluation harness (`evals/harness/`), pytest behavioral tests, and `run-btests-pytest.sh` cluster runner are all preserved.

## Test plan
- [x] `grep -r evalhub` returns no hits in tracked files
- [x] `uv sync --extra test --extra test-mlflow` succeeds without eval-hub-sdk
- [x] 68/68 harness unit tests pass (`pytest evals/harness/tests`)
- [x] Cluster btests ran against live cluster (203 passed, 15 failed from pre-existing flakes — latency thresholds, LLM response quality)
- [x] eval-gating.yml validates as correct YAML


andrewdonheiser (commented):
### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kami619:
@andrewdonheiser there are conflicts on this branch.

Also, where would this removed feature go ? do we have a place for it to land or are we dropping this completely ? Do you have any references for it in your blog that you published recently ?

andrewdonheiser:
> @andrewdonheiser there are conflicts on this branch.
> 
> Also, where would this removed feature go ? do we have a place for it to land or are we dropping this completely ? Do you have any references for it in your blog that you published recently ?

Hi @kami619 i'll clean up the conflicts now. the functionality is just deleted and there are no references in my blog.

kami619:
### Findings

  1. High: .github/workflows/eval-gating.yml stops being a real gate once deterministic is deleted. After this diff, the only remaining job is cluster-btests, and every substantive step in it is guarded by
     steps.cluster-btests-config.outputs.run_cluster_btests == 'true'. When cluster vars/secrets are unavailable, especially on fork PRs, the workflow can go green after only resolving config without
     running any tests. The PR also keeps the surviving local suite in pyproject.toml (evals/harness/tests) but does not replace the deleted deterministic execution path with any always-on check. That turns
     “Inner Loop Gating” into an optional cluster launcher rather than an actual gate. The description change in .github/scripts/generate_ci_health_page.py makes the mismatch worse by still advertising
     this as “Behavioral pytest gating.” The clean move is to add a small always-on harness job, or split cluster btests into a separate opt-in workflow instead of keeping this one as a gate.

  2. Medium: .github/workflows/eval-gating.yml spreads one control-flow decision across six separate steps. Repeating the same if: on checkout, cluster setup, dependency install, test execution, artifact
     upload, and logout is workflow spaghetti: it’s easy for a future edit to add an unguarded step, and a green job no longer means “tests ran.” A cleaner structure would resolve config in a tiny
     precursor job or reusable action, then gate the whole cluster-btests job once via needs outputs.

kami619:
@andrewdonheiser WDYT about those findings ?

andrewdonheiser:
> @andrewdonheiser WDYT about those findings ?

Not sure this is why I wanted you to look at it. I didn't realize CI touched this (it was an abandoned feature) when I started the task

.github/workflows/code-quality.yml
```diff
@@ -66,7 +66,21 @@ jobs:
           GH_TOKEN: ${{ github.token }}
         run: |
           if [ "${{ github.event_name }}" = "pull_request" ]; then
-            changed=$(gh pr diff ${{ github.event.pull_request.number }} --name-only | grep '\.md$' | tr '\n' ' ' || true)
+            if ! diff_files=$(gh pr diff ${{ github.event.pull_request.number }} --name-only); then
```
kami619, line 75: can you please add a note denoting this change for future reference.

evals/harness/mlflow_client.py
```diff
@@ -88,9 +88,9 @@ def verify_connection(self) -> bool:
             if "401" in msg or "403" in msg or "Authorization" in msg:
                 logger.error(
                     "MLflow auth failed (%s). Ensure MLFLOW_TRACKING_TOKEN is "
-                    "set in the adapter pod environment. EvalHub does not "
-                    "support secretKeyRef — pass the token as a literal value "
-                    "in the provider runtime Env.",
+                    "set in the environment. Pass the token as a literal "
+                    "value or refresh via: "
+                    "export MLFLOW_TRACKING_TOKEN=$(oc whoami -t)",
```
kami619: lets make sure the `MLFLOW_TRACKING_TOKEN` is being asked to set in the right way in this error blob.

kami619 (dismissed):
lets look at these and make sure we have a dry run against the live cluster before we merge.

.github/workflows/code-quality.yml
andrewdonheiser: Added in c264ae7. The note covers both reasons: `gh pr diff --name-only` also lists deleted files, so lychee was being handed the .md paths this PR removes and failing on them; and the `gh pr diff` failure is fatal on purpose since the old pipeline-wide `|| true` let the job go green without checking any links.

evals/harness/mlflow_client.py
andrewdonheiser: Good catch — I dropped the recipe entirely in c264ae7. It is not settled which token is correct here: `run-btests-pytest.sh` refreshes with `oc whoami -t`, while `tracing.md` documents `oc create token default`. Rather than bake in a guess, the error now just says `MLFLOW_TRACKING_TOKEN` must be set to a valid token. I also removed the "and re-register the provider" line — that was the EvalHub provider this PR deletes, so it was unactionable.

andrewdonheiser:
Both inline comments addressed in c264ae7.

On the gating findings: the `deterministic` job only ever ran `evals/evalhub_adapter/tests`, so it goes away with EvalHub — there is nothing being un-gated that was gating anything else. `eval-gating.yml` was created in #185 with both jobs at once, so there is no earlier state to restore either; removing EvalHub just leaves the cluster-btests job. If you want an always-on job added (the only cluster-free suites are `evals/harness/tests` and `tests/behavioral/deterministic/test_run_btests_pytest.py`, neither of which has ever run in CI), that would be new work — happy to file a follow-up.

Same for the repeated `if:` guards on the cluster-btests steps: that structure is from #185 and this PR only deleted the job above it.

Will run the btests against the live cluster and post results before merge.

kami619 (approved):
lgtm

andrewdonheiser:
## Live cluster dry run

Ran `tests/behavioral/deterministic/run-btests-pytest.sh` against `agen-e2e-rhoai2` (namespace `adonheis-testing`) on c264ae7. MLflow tokens refreshed and tracing confirmed enabled on all agents beforehand.

**202 passed / 31 failed** across 11 agents.

| Agent | Result |
|---|---|
| autogen/templates/mcp_agent | 18 passed, 0 failed |
| langgraph/templates/human_in_the_loop | 16 passed, 0 failed |
| crewai/templates/websearch_agent | 18 passed, 1 failed |
| langgraph/templates/react_agent | 22 passed, 3 failed |
| langgraph/templates/react_with_database_memory | 17 passed, 2 failed |
| llamaindex/templates/websearch_agent | 19 passed, 1 failed |
| vanilla_python/templates/openai_responses_agent | 24 passed, 3 failed |
| google/templates/adk | 21 passed, 1 failed |
| langflow/templates/simple_tool_calling_agent | 28 passed, 1 failed |
| a2a/templates/langgraph_crewai_agent | 18 passed, 1 failed |
| langgraph/templates/agentic_rag | 1 passed, 18 failed |

**agentic_rag is a broken environment, not a regression.** All 18 failures are the same error:

```
HTTP 500: Vector_store 'vs_f91e065e-92a4-4ce8-a06d-ec976e85d18e' not found
```

Its `VECTOR_STORE_ID` points at a store that no longer exists on the cluster. Fix is `make load-docs` + redeploy; unrelated to this PR.

The remaining 13 failures are the usual flakes — two latency thresholds marginally over (8.33s > 8.0s, 20.19s > 20.0s), one 504 gateway timeout, sync/stream tool-set mismatches, and LLM response-quality judgments. Same profile as the earlier run noted in the PR description (203/15).

Nothing in the failures touches the harness or workflow code this PR changes.

kami619:
@andrewdonheiser thanks a lot.

Lets log a JIRA bug for this issue and we can pull that in. Lets work towards enabling the entire suite or a sub-set of these tests in the nightly CI. We can talk more offline.

 > agentic_rag is a broken environment, not a regression. All 18 failures are the same error: HTTP 500: Vector_store 'vs_f91e065e-92a4-4ce8-a06d-ec976e85d18e' not found
Its VECTOR_STORE_ID points at a store that no longer exists on the cluster. Fix is make load-docs + redeploy; unrelated to this PR.

PR #317, opened 2026-08-17
ci: extract QG4 deployment tests into composite action (RHAIENG-6540)

mpk-droid:
## Description

Extracts the QG4 agent deployment health-check logic from `agent-deployment-test.yaml` into a reusable composite action at `.github/actions/run-qg4/action.yml`. This enables the orchestrator workflow to call QG4 as a step while preserving the standalone workflow for ad-hoc/debug runs.

The composite action follows the same convention as `run-qg1` and `run-qg7`: it contains only the gate logic (`make test-integration`). The caller handles checkout, setup-cluster, artifact upload, and logout — so an orchestrator can share a single cluster session across multiple gates and decide how/when to collect artifacts.

**Changes:**
- **New:** `.github/actions/run-qg4/action.yml` — composite action with one input (`agent-dir`). Env vars (API keys, DB config, etc.) are passed via step-level `env:` on the calling step, matching the `notify-slack` pattern — no pass-through input declarations needed.
- **Updated:** `.github/workflows/agent-deployment-test.yaml` — `test-agent` job now calls the composite action; matrix, triggers, `verify-cluster-connection`, and `notify-slack` jobs are untouched.

**Minor behavior change:** the `if` guard for autogen-mcp-agent (skip when `MCP_SERVER_URL` is unset) now covers the composite action step (test only), not just the inline test step. Previously, an unset `MCP_SERVER_URL` still ran cluster setup, artifact upload of a nonexistent `results.xml`, and logout; now setup-cluster and logout still run but the test and upload are skipped. This is an intentional improvement.

## Jira Ticket

RHAIENG-6540

## Testing

Triggered `workflow_dispatch` runs on this branch against upstream:
- **Run 1 (initial):** https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/32169430475 — all 11 agents passed ✅
- **Run 2 (review fixes — slim action, step-level env):** https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/32179183012 — all 11 agents passed ✅
- **Run 3 (move upload to caller):** https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/32183358684 — 10/11 passed ✅, a2a-langgraph-crewai failed due to Docker Hub rate limit (`toomanyrequests` on `python:3.12-slim` pull) — unrelated to this PR

## Checklist

## Review Guidance

- The composite action is run-only: just `make test-integration` with `working-directory`. Caller handles checkout, setup-cluster, upload-artifact, and logout — matching QG1/QG7 convention.
- Env vars use step-level `env:` instead of input declarations — adding a new env var is a one-line change in the caller
- `if-no-files-found: warn` on upload-artifact for consistency

### DIFF
--- a/.github/actions/run-qg4/action.yml
+++ b/.github/actions/run-qg4/action.yml
@@ -1,5 +1,5 @@
 name: "Run QG4 Agent Deployment Test"
-description: "Deploy an agent to OpenShift, run its integration test suite, and upload results."
+description: "Run an agent's integration test suite and upload results. Caller handles checkout, setup-cluster, and logout."
 
 inputs:
   agent-name:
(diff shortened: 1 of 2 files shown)


andrewdonheiser (changes requested):
Reviewed the branch directly and compared against the existing `run-qg1` / `notify-slack` conventions in this repo. The refactor is behavior-preserving for the nightly run and the dispatch validation backs that up — nice clean extraction. One thing I'd like changed before merge; the rest are non-blocking.

## Please fix: `run-qg4` bundles setup + logout, unlike `run-qg1`

`.github/actions/run-qg4/action.yml` (`Setup cluster tools`, `Logout` steps)

The convention established by QG1 (`.github/workflows/qg1-cluster-readiness.yml`) is that the caller does checkout → `setup-cluster` → `run-qgN` → upload → logout, and the QG action is *just the gate*. `run-qg4` instead swallows `setup-cluster`, `upload-artifact`, and `oc logout`.

That matters specifically for the orchestrator this PR exists to enable: a job running QG1 then QG4 will re-install python/uv/oc/helm and re-`oc login` (wasteful), and then `run-qg4`'s trailing `oc logout` tears down the session for any gate sequenced *after* QG4 in the same job.

Could you hoist `setup-cluster` and `Logout` back into the calling workflow so QG4 matches QG1's shape? Happy to leave the artifact upload inside if you prefer.

---

## Non-blocking notes

**1. Twelve pass-through inputs where step-level `env:` would do.** Every input is a 1:1 passthrough to an identically-named env var. There's precedent right in this workflow: the `notify-slack` step sets `env: SLACK_WEBHOOK_URL` on the `uses:` line and `notify.sh` reads it from the environment — env on a `uses:` step is visible to a composite action's `run` steps. As written, adding one env var for a new agent means editing three places (input decl, env map, caller `with:`). Not worth blocking on, but something to consider if this action grows.

**2. Secret inputs widen into `INPUT_*` for every step in the action.** `oc-token`, `postgres-password`, `nvidia-api-key`, and `api-key` become `INPUT_*` env vars on all of the composite's run steps, including `oc logout`; previously each was scoped to the single step that needed it. Values stay masked in logs so this is low severity, and note 1's approach would also resolve it.

**3. "No behavior change" isn't quite accurate for the autogen skip path.** The `if: matrix.agent.name != 'autogen-mcp-agent' || vars.MCP_SERVER_URL != ''` guard moved from the test step up to the whole composite step. Previously an unset `MCP_SERVER_URL` still ran cluster setup, an artifact upload of a nonexistent `results.xml`, and logout; now the whole thing is skipped. That's an improvement, not a regression — just worth a line in the description. Also worth noting the validation run had `MCP_SERVER_URL` set, so that path wasn't exercised.

**4. Small stuff.**
- `required: true` on `oc-token` / `cluster-api-url` isn't enforced by GitHub for composite actions — omitting one yields `oc login --token=""` rather than a clear failure. Pre-existing in `setup-cluster` too.
- QG1's upload sets `if-no-files-found: warn` explicitly, QG4's doesn't. Default is `warn`, so no functional difference — just inconsistent.
- Dropping the job-level `env: API_KEY/BASE_URL/MODEL_ID` block is correct (it was shadowed by the step env). No issue.

mpk-droid:
Thanks for the thorough review, Andrew. All addressed in aa86057:

**Setup + logout bundling (please fix):** Done — hoisted `setup-cluster` and `oc logout` back into the calling workflow. `run-qg4` now only does `make test-integration` + `upload-artifact`, matching QG1's shape. The orchestrator can share a single cluster session across gates.

**1. Pass-through inputs → step-level `env:`:** Adopted. Removed all 12 env-var inputs plus `oc-token`/`cluster-api-url`. The action now has just two inputs (`agent-name`, `agent-dir`). Env vars are passed via `env:` on the calling step, same pattern as `notify-slack`. Adding a new env var is a one-line change.

**2. Secret `INPUT_*` scope:** Resolved by the above — secrets are now step-level `env:` on the caller, scoped to the composite action steps only (which is just the test + upload).

**3. Autogen skip path:** Updated the PR description to note the behavior change explicitly. You're right that the validation run had `MCP_SERVER_URL` set so the skip path wasn't exercised — the `autogen-mcp-agent` job ran and passed, which validates the happy path. The skip path is a pure no-op (step is skipped entirely), so I'm comfortable it's correct.

**4. Small stuff:**
- `if-no-files-found: warn` added to `upload-artifact` for consistency with QG1.
- `required: true` enforcement gap is pre-existing in `setup-cluster` — not addressing here.

Validation run on the updated code: https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/32179183012 — all 11 agents passed.

andrewdonheiser (approved):

PR #319, opened 2026-08-18
fix(ci): assume dedicated QG1 service account for readiness checks

kami619:
## Description
  This PR fixes the immediate `QG1: Cluster Readiness` RBAC failure by keeping the existing shared `OC_TOKEN` only as a bootstrap identity, then minting a short-lived token for a dedicated `qg1-readiness` service account before the checker runs.
  It adds:
  - a dedicated QG1 assume-action in `.github/actions/assume-qg1-service-account/action.yml`
  - a minimal read-only cluster RBAC manifest in `.github/cluster/qg1-readiness-rbac.yaml`
  - workflow wiring in `.github/workflows/qg1-cluster-readiness.yml`
  - contract tests to lock in the new workflow/action/RBAC behavior
  This is the short-term working fix. The longer-term identity model is tracked separately in `RHAIENG-6993`, which covers replacing secret/bootstrap-based auth with a GitHub OIDC-based short-lived credential flow.
  ## Jira Ticket
  [RHAIENG-6993](https://redhat.atlassian.net/browse/RHAIENG-6993)
  ## Testing
  Additional verification performed:
  - `uv run pytest .github/scripts/tests/test_qg1_workflow_contract.py .github/scripts/tests/test_qg1_cluster_readiness.py -q`
  - Applied `.github/cluster/qg1-readiness-rbac.yaml` to the cluster
  - Verified effective permissions for `system:serviceaccount:ci-testing:qg1-readiness`:
    - can list `clusterversions.config.openshift.io`
    - can list `nodes`
    - can list `namespaces`
  - Verified `system:serviceaccount:ci-testing:github-actions` can mint `serviceaccounts/token` in `ci-testing`
  - Ran the QG1 checker end to end with a freshly minted short-lived `qg1-readiness` token:
    - `api_health: PASS`
    - `cluster_version: PASS`
    - `gpu_nodes: PASS`
    - `required_namespaces: PASS`
  ## Checklist
  ## Review Guidance
  Please review these together as one contract:
  - `.github/workflows/qg1-cluster-readiness.yml`
  - `.github/actions/assume-qg1-service-account/action.yml`
  - `.github/cluster/qg1-readiness-rbac.yaml`
  - `.github/scripts/tests/test_qg1_workflow_contract.py`
  Key design choice:
  - This PR intentionally keeps the short-term model on a bootstrap login plus a short-lived dedicated QG1 token.
  - It does **not** attempt to solve the longer-term GitHub OIDC identity model in this branch; that plan is tracked in `RHAIENG-6993`.
  ## Related PRs
  None.

[RHAIENG-6993]: https://redhat.atlassian.net/browse/RHAIENG-6993?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/actions/assume-qg1-service-account/action.yml
+++ b/.github/actions/assume-qg1-service-account/action.yml
@@ -1,0 +1,38 @@
+name: "Assume QG1 Service Account"
+description: "Mint a short-lived token for the QG1 readiness service account and re-login with that identity."
+
+inputs:
+  service-account:
+    description: "Service account to assume for cluster-readiness checks"
+    required: false
+    default: "qg1-readiness"
+  namespace:
+    description: "Namespace that owns the service account"
+    required: false
+    default: "ci-testing"
+
+runs:
+  using: "composite"
+  steps:
+    - name: Mint token and login as dedicated QG1 service account
+      shell: bash
+      env:
+        SERVICE_ACCOUNT: ${{ inputs.service-account }}
+        NAMESPACE: ${{ inputs.namespace }}
+      run: |
+        set -euo pipefail
+
+        server="$(oc whoami --show-server)"
+        qg1_token="$(oc create token "${SERVICE_ACCOUNT}" -n "${NAMESPACE}" --duration=20m)"
+
+        oc login \
+          --token="${qg1_token}" \
+          --server="${server}" \
+          --namespace="${NAMESPACE}" >/dev/null
+
+        expected_identity="system:serviceaccount:${NAMESPACE}:${SERVICE_ACCOUNT}"
+        actual_identity="$(oc whoami)"
+        if [[ "${actual_identity}" != "${expected_identity}" ]]; then
+          echo "Expected ${expected_identity}, got ${actual_identity}" >&2
+          exit 1
+        fi
--- a/.github/actions/assume-service-account/action.yml
+++ b/.github/actions/assume-service-account/action.yml
@@ -1,9 +1,9 @@
-name: "Assume QG1 Service Account"
-description: "Mint a short-lived token for the QG1 readiness service account and re-login with that identity."
+name: "Assume Service Account"
+description: "Mint a short-lived token for a service account and re-login with that identity."
 
 inputs:
   service-account:
-    description: "Service account to assume for cluster-readiness checks"
+    description: "Service account to assume; defaults to the qg1-readiness convention used by QG1"
     required: false
     default: "qg1-readiness"
   namespace:
@@ -14,7 +14,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-    - name: Mint token and login as dedicated QG1 service account
+    - name: Mint token and login as service account
       shell: bash
       env:
         SERVICE_ACCOUNT: ${{ inputs.service-account }}
@@ -24,6 +24,7 @@ runs:
 
         server="$(oc whoami --show-server)"
         qg1_token="$(oc create token "${SERVICE_ACCOUNT}" -n "${NAMESPACE}" --duration=20m)"
+        echo "::add-mask::${qg1_token}"
 
         oc login \
           --token="${qg1_token}" \
--- a/.github/cluster/qg1-readiness-rbac.yaml
+++ b/.github/cluster/qg1-readiness-rbac.yaml
@@ -1,0 +1,33 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: qg1-readiness
+  namespace: ci-testing
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: qg1-readiness-reader
+rules:
+  - apiGroups: ["config.openshift.io"]
+    resources: ["clusterversions"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: [""]
+    resources: ["namespaces"]
+    verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: qg1-readiness-reader
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: qg1-readiness-reader
+subjects:
+  - kind: ServiceAccount
+    name: qg1-readiness
+    namespace: ci-testing
--- a/.github/cluster/qg1-readiness-rbac.yaml
+++ b/.github/cluster/qg1-readiness-rbac.yaml
@@ -11,13 +11,13 @@ metadata:
 rules:
   - apiGroups: ["config.openshift.io"]
     resources: ["clusterversions"]
-    verbs: ["get", "list", "watch"]
+    verbs: ["get", "list"]
   - apiGroups: [""]
     resources: ["nodes"]
-    verbs: ["get", "list", "watch"]
+    verbs: ["get", "list"]
   - apiGroups: [""]
     resources: ["namespaces"]
-    verbs: ["get", "list", "watch"]
+    verbs: ["get", "list"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
--- a/.github/scripts/tests/test_qg1_workflow_contract.py
+++ b/.github/scripts/tests/test_qg1_workflow_contract.py
@@ -8,7 +8,11 @@ import yaml
 
 REPO_ROOT = Path(__file__).resolve().parents[3]
 ACTION_PATH = REPO_ROOT / ".github" / "actions" / "run-qg1" / "action.yml"
+ASSUME_ACTION_PATH = (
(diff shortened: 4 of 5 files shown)


kami619:
A working run using this branch: https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/32112000896/job/95633274834

shricharan-ks (dismissed):
## Multi-Perspective Review — PR #319

Reviewed from four independent angles: **security/RBAC**, **CI workflow correctness**, **test quality**, and **design/architecture**.

---

### Verdict: APPROVE

The bootstrap-then-assume pattern is a clean short-term fix that improves the security posture by scoping QG1 checks to a dedicated, least-privilege service account. The RBAC manifest is minimal, the action fails cleanly on all error paths, and contract tests lock in the critical invariants. The long-term OIDC-based approach is correctly deferred to RHAIENG-6993.

---

### What this PR does well

- **Correct least-privilege RBAC** — ClusterRole is required (clusterversions, nodes are cluster-scoped) and grants only read verbs for the three resources the checker actually queries
- **Clean failure on all error paths** — `set -euo pipefail` + identity verification catch every failure mode: SA doesn't exist, bootstrap lacks permission, login fails, wrong identity post-login
- **No concurrency issues** — `oc create token` mints stateless JWTs; parallel runs get independent tokens in separate runner filesystems
- **Token TTL (20m) safely exceeds job timeout (15m)** — 5-minute margin for clock skew
- **Contract tests validate step ordering** — setup-cluster < assume < run-qg1 assertion catches reordering regressions
- **Generic action implementation** — despite the QG1-specific name, the action already accepts `service-account` and `namespace` inputs, making it reusable

---

### Suggestions

#### 1. [Security] Add `::add-mask::` for the minted token

The token is captured in a variable and not echoed, but if anyone adds `set -x` for debugging, it would leak. Adding GitHub's built-in masking is a one-liner defense-in-depth:

```bash
qg1_token="$(oc create token "${SERVICE_ACCOUNT}" -n "${NAMESPACE}" --duration=20m)"
echo "::add-mask::${qg1_token}"    # <-- add this line
```

This ensures the token is redacted from all subsequent log output regardless of debug flags.

#### 2. [Security] Drop `watch` verb from ClusterRole

The QG1 checker runs point-in-time `oc get` commands — it never sets up a watch stream. Removing `watch` from all three rules tightens the ClusterRole to true least-privilege:

```yaml
verbs: ["get", "list"]    # instead of ["get", "list", "watch"]
```

#### 3. [Test gap] No assertion for `--duration=20m`

The 20-minute TTL is a critical security parameter — if someone removes it, tokens default to 1 hour. Add an assertion in `test_assume_qg1_service_account_action_mints_token_and_relogs`:

```python
assert any("--duration=20m" in step.get("run", "") for step in bash_steps)
```

#### 4. [Test gap] No negative RBAC assertion for scope creep

The RBAC test verifies the three expected resource rules but doesn't assert there are _only_ three. A fourth rule granting `pods/exec` would pass undetected:

```python
assert len(rule_map) == 3  # catch accidental RBAC scope creep
```

Similarly, consider asserting that no verb outside `{"get", "list", "watch"}` appears across all rules (or `{"get", "list"}` if suggestion 2 is taken).

#### 5. [Test gap] No assertion for identity verification

The action's post-login `oc whoami` check is the fail-safe that catches login-as-wrong-identity bugs. The test checks for `oc whoami --show-server` (the server lookup before login) but never asserts the identity comparison logic is present. Consider adding:

```python
assert any("oc whoami" in step.get("run", "") for step in bash_steps 
           if "--show-server" not in step.get("run", ""))
```

or more directly checking for the `expected_identity` / `actual_identity` comparison.

#### 6. [Design] Consider renaming action to `assume-service-account`

The action's implementation is fully generic — only the directory name and `name:` field couple it to QG1. Renaming now avoids a redirect later when QG2/QG3 adopt it. If renamed, the `qg1-readiness` default should either be removed (making `service-account` required) or documented as the convention.

---

### No concerns found in

- **Failure modes**: `set -euo pipefail` + the identity check covers all error paths. `oc login >/dev/null` suppresses stdout but stderr still surfaces.
- **Logout behavior**: `oc logout` correctly targets the current (QG1) identity; the bootstrap secret is unaffected in GitHub Secrets.
- **RBAC manifest location**: `.github/cluster/` is appropriate for CI-owned infrastructure. The manifest is not auto-applied — cluster admin applies manually, which is correct.
- **Bootstrap privilege escalation**: The shared `OC_TOKEN` can mint tokens for any SA in `ci-testing`. This is a pre-existing trust boundary, not introduced by this PR.
- **Concurrent runs**: Independent, stateless JWT minting. No shared state.

LGTM — solid least-privilege improvement with clean error handling.

kami619:
@shricharan-ks I have implemented all the suggestions, there are some good catches in there. Thanks for the detailed review. If you can take another look, that would be great.

### Pushed commit:
  • b1f1d81 — fix(ci): harden QG1 service-account assumption flow

### What changed:
  • added ::add-mask:: for the minted service-account token
  • removed watch from the QG1 ClusterRole verbs
  • extended the contract test to assert:
    • --duration=20m
    • the post-login identity verification logic
    • exact RBAC rule count and verb set
  • renamed the generic assume action from assume-qg1-service-account to assume-service-account
  • updated the workflow and contract tests to use the new action path/name

### Verification:
  • uv run pytest .github/scripts/tests/test_qg1_workflow_contract.py -q -> 17 passed
  • ruff check .github/scripts/tests/test_qg1_workflow_contract.py -> passed

shricharan-ks (approved):
## Fresh Multi-Perspective Review — PR #319 (post-revision)

Reviewed from four new angles: **operational readiness**, **reusability/future consumers**, **security deep-dive**, and **test completeness**. This supplements the previous review which focused on RBAC, CI correctness, test quality, and design.

All six suggestions from the previous review have been addressed (token masking, `watch` verb removal, duration test, scope creep assertion, identity verification test, generic rename). This review looks at what remains.

---

### Verdict: APPROVE

The revision is solid. The items below are non-blocking improvements.

---

### New findings

#### 1. [Operational — Medium] No onboarding docs for the RBAC prerequisite

The RBAC manifest at `.github/cluster/qg1-readiness-rbac.yaml` must be manually applied by a cluster admin before this flow works. No README, Makefile target, or inline comment documents this step. If merged without the SA existing on the cluster, the next nightly run will fail at `oc create token` with a raw API error (`serviceaccounts "qg1-readiness" not found`) and no breadcrumb pointing to the manifest. Add a one-liner comment in the workflow or a `## Cluster Prerequisites` section in the RBAC manifest itself.

#### 2. [Reusability — Low] Stale `qg1_token` variable name in generic action

The action was renamed to `assume-service-account` but the bash variable is still `qg1_token`. Rename to `sa_token` for consistency with the generic naming — it's a one-line change that avoids confusing future consumers.

#### 3. [Reusability — Low] Consider exposing `--duration` as an input

The 20m TTL is hardcoded. Different consumers may need different TTLs. Exposing it as an input with `default: "20m"` costs one line and makes the action fully parameterizable.

#### 4. [Security — Low] `list` on namespaces exposes cluster structure

The ClusterRole grants `list` on namespaces cluster-wide, revealing all tenant/team namespace names. If the QG1 checker only needs to verify specific known namespaces exist (e.g., `ci-testing`), `get` with `resourceNames` would be tighter. Worth evaluating in a follow-up.

#### 5. [Security — Low] Kubeconfig residue on job cancellation

`oc logout` runs `if: always()` but doesn't run on hard cancellation/timeout. The token is written to `~/.kube/config` and accessible to subsequent steps. The 20m TTL is sufficient mitigation for GitHub-hosted (ephemeral) runners. For self-hosted runners, this would be a real risk — composite actions don't support `post:` cleanup, so a workflow-level mitigation would be needed.

#### 6. [Test — Low] No test for `::add-mask::` presence

The token masking line was a security recommendation from the previous review and was implemented. But no contract test asserts `::add-mask::` appears in the action's bash step. If someone removes it, no test fails.

#### 7. [Operational — Low] Failure modes are indistinguishable in Slack alerts

When the assume step fails, the QG1 checker never runs, so no `summary.json` is produced. The Slack notification fires on job failure but can't differentiate "cluster unhealthy" from "assume-SA setup broken". The artifact upload silently warns "no files found" — a future improvement could emit a distinguishable error artifact.

---

### Not a concern

- **Token TTL timing**: The 20m clock starts when `oc create token` executes (step runtime), not job start. Runner queue time doesn't consume TTL.
- **`::add-mask::` ordering**: `oc create token` writes the token only to stdout (captured by `$()`). Stderr warnings would appear unmasked but don't contain the token.
- **ClusterRoleBinding scope**: Nodes, clusterversions, and namespaces are all cluster-scoped resources — a namespace-scoped RoleBinding cannot grant access to them. The ClusterRoleBinding is architecturally necessary.
- **Bash variable exfiltration**: The variable is local to the composite step's shell process, not exported or accessible to later steps. The real exfiltration vector is `~/.kube/config` (covered in finding 5).
- **Rollback**: Straightforward — revert the workflow change; the RBAC manifest and action are inert without the workflow reference.

LGTM — previous feedback well-addressed, remaining items are all non-blocking follow-ups.

PR #321, opened 2026-08-18
ci: extract run-qg7 action (RHAIENG-6541)

andrewdonheiser:
Extract the QG7 behavioral eval logic from eval-gating.yml into a reusable composite action at .github/actions/run-qg7/action.yml.

This enables the orchestrator workflow to call QG7 as a step while preserving the standalone workflow for push/PR triggers and ad-hoc runs.

### Acceptance criteria
- .github/actions/run-qg7/action.yml exists with inputs for agent name
- eval-gating.yml updated to call the composite action (no behavior change)
- Existing push/PR and workflow_dispatch triggers still work

Ref: RFC Discussion #285
Jira: [RHAIENG-6541](https://redhat.atlassian.net/browse/RHAIENG-6541)

[RHAIENG-6541]: https://redhat.atlassian.net/browse/RHAIENG-6541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/actions/run-qg7/action.yml
```diff
@@ -0,0 +1,25 @@
+name: "Run QG7 Behavioral Eval"
+description: "Run shared cluster behavioral pytest suites after checkout and cluster setup."
+
+inputs:
+  agent-name:
```
mpk-droid: **nit: pass-through input → step-level env**

`agent-name` exists solely to set the `QG7_AGENT_NAME` env var inside the action. In the updated QG4 action (#317), we removed pass-through inputs and kept only structural ones (like `agent-dir` for `working-directory:`). For consistency, consider removing the input and having the caller pass the env var directly:

```yaml
# action.yml — no inputs needed
runs:
  using: "composite"
  steps:
    - name: Execute shared QG7 runner
      shell: bash
      run: |
        set -euo pipefail
        if [[ -n "${QG7_AGENT_NAME:-}" ]]; then
          read -r -a selected_agents <<< "${QG7_AGENT_NAME}"
          bash "tests/behavioral/deterministic/run-btests-pytest.sh" "${selected_agents[@]}"
        else
          bash "tests/behavioral/deterministic/run-btests-pytest.sh"
        fi
```

```yaml
# eval-gating.yml caller
- name: Run QG7 behavioral eval
  uses: ./.github/actions/run-qg7
  env:
    QG7_AGENT_NAME: ${{ needs.deterministic.outputs.cluster_btests_agents }}
```

This way adding a new env var is a one-line change in the caller — no action.yml update needed.

.github/actions/run-qg7/action.yml
```diff
@@ -0,0 +1,25 @@
+name: "Run QG7 Behavioral Eval"
+description: "Run shared cluster behavioral pytest suites after checkout and cluster setup."
```
mpk-droid: **nit:** QG4's updated description explicitly states caller responsibilities: *"Caller handles checkout, setup-cluster, artifact upload, and logout."* Consider aligning:

```yaml
description: "Run shared cluster behavioral pytest suites. Caller handles checkout, setup-cluster, artifact upload, and logout."
```

kami619:
@andrewdonheiser how would it intersect with the other PR you have https://github.com/red-hat-data-services/agentic-starter-kits/pull/315 ? would they conflict each other based on the merge time ?

.github/actions/run-qg7/action.yml
andrewdonheiser: Done in 74413c7 — dropped the `agent-name` input; the caller now sets `QG7_AGENT_NAME` as a step-level `env`.

One adjustment to your snippet: with the input (and its `default: ""`) gone, the variable can now be genuinely unset, so the guard has to be `${QG7_AGENT_NAME:-}` rather than `${QG7_AGENT_NAME}` — otherwise `set -u` aborts the step on the run-all path, which is the common case. I also left a short comment block in `action.yml` documenting the env contract, since without an `inputs:` block there is nowhere else for a caller to discover the variable exists.

Worth flagging for the wider convention: `run-qg1` and `run-qg2` on `main` are both pure pass-through inputs mapped to `env:` — i.e. the pattern you are moving away from. Happy to align all of them in a follow-up if we want env pass-through as the standard.

.github/actions/run-qg7/action.yml
andrewdonheiser: Done in 74413c7 — adopted the QG4 wording verbatim.

.github/actions/run-qg7/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    - name: Execute shared QG7 runner
+      shell: bash
+      env:
+        QG7_AGENT_NAME: ${{ inputs.agent-name }}
+      run: |
+        set -euo pipefail
+
+        if [[ -n "${QG7_AGENT_NAME}" ]]; then
+          read -r -a selected_agents <<< "${QG7_AGENT_NAME}"
+          bash "tests/behavioral/deterministic/run-btests-pytest.sh" "${selected_agents[@]}"
```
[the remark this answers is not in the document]
andrewdonheiser: Fixed in 74413c7, in the `Resolve cluster btests config` step where the value originates:

```bash
_agents=()
read -r -a _agents <<< "$(tr '\n' ' ' <<< "${cluster_btests_agents}")"
cluster_btests_agents="${_agents[*]:-}"
```

This collapses newlines and repeated whitespace and trims, so only a single-line space-separated value reaches `GITHUB_OUTPUT`. Verified against your repro: a value of `$'a b\ncluster_btests_agents=INJECTED\nrun_cluster_btests=true'` now writes one output line and does not override `run_cluster_btests`. Empty input still yields `""` so the action keeps taking the run-all branch.

Note the `${_agents[*]:-}` rather than `${_agents[*]}` — expanding an empty array under `set -u` is only safe on bash 4.4+, and the runners have bash 5.x, but the `:-` makes it version-independent.

For severity calibration: `EVAL_CLUSTER_BTESTS_AGENTS` is an admin-set repo variable, so this is malformed-input robustness rather than a security boundary — anyone able to set it already has broader access.

andrewdonheiser:
@kami619 Good catch — they will conflict, but only textually and it is a small fix.

Both PRs edit the same step. #315 deletes the `deterministic` job, folds `Resolve cluster btests config` into the `cluster-btests` job, and rewrites `Run shared btest runner` to read `steps.cluster-btests-config.outputs.cluster_btests_agents`. This PR replaces that same step with `uses: ./.github/actions/run-qg7`. Same lines, same file, so whichever lands second hits a conflict. Nothing else overlaps — this PR otherwise only adds one new file and two `paths:` entries.

Resolution is mechanical — keep #315's single-job structure and swap the inlined runner for the composite action, preserving its `if:` guard:

```yaml
- name: Run QG7 behavioral eval
  if: steps.cluster-btests-config.outputs.run_cluster_btests == 'true'
  uses: ./.github/actions/run-qg7
  env:
    QG7_AGENT_NAME: ${{ steps.cluster-btests-config.outputs.cluster_btests_agents }}
```

One thing to watch: the newline-normalization I just added for the CodeRabbit finding lives in `Resolve cluster btests config`, which #315 moves. It needs to come along with the step.

Suggest merging #315 first since it is much larger and deletion-heavy, then I rebase this one onto it — the rebase here is a handful of lines. Happy to go the other way if you prefer, just wanted to flag the ordering.

mpk-droid (approved):
lgtm

PR #325, opened 2026-08-20
ci(qg7): deploy and tear down btest agents within the gate (RHAIENG-7057)

andrewdonheiser:
QG7 previously assumed its agents had been deployed out of band. This branch gives the gate its
own deploy and teardown: it resolves the agent selection once, deploys that set, runs the shared
btest runner over it, and tears it down in an `if: always()` step inside the composite action.

The original change shipped with static verification only — ruff, unit tests, YAML parse,
`--print-selection`. Every cluster-dependent step was unverified. This PR is that work finished:
five stages against a live cluster and a real GitHub Actions runner, which turned up **seven
false-greens**, all fixed and re-verified here.

---

# Prerequisites — all satisfied

Nothing outside this PR is outstanding. `MLFLOW_TRACKING_URI` is now set as a repository variable
(**RHAIENG-7096**, closed) to the CI cluster's data-science-gateway endpoint, verified live with a
200 + JSON from `/api/3.0/mlflow/server-info`. Fix 6 makes CI read it, and the namespace-scoped CI
SA cannot discover the route itself — routes are cluster-scoped — so without it every agent's
tracing verification would have failed.

Note for anyone re-deriving that value: the standalone `mlflow-redhat-ods-applications.apps.…`
route resolves but **404s** the API. On RHOAI 3.5 MLflow is a cluster-scoped singleton served
through the data-science gateway under `/mlflow`, and namespaces join it as *workspaces* via the
`mlflow-tracking=enabled` label rather than getting their own install. Picking the standalone route
is false-green 3 below.

Everything else degrades gracefully:

| Situation | What the gate does |
|---|---|
| `OPENAI_*` unset (**RHAIENG-7097**, current state) | Skips `openai_responses_agent`, logs which variables to set, runs the other **8** and goes green |
| `mcp_agent` / `agentic_rag` broken (**RHAIENG-7094**, **RHAIENG-7095**) | Excluded via `EXCLUDED_AGENTS`, logged |

So the gate covers **9 of 12 agents** with `OPENAI_*` set, **8 of 12** without. No ticket is closed
by merging this: 7094 and 7095 track *fixing* the two agents, which this PR only excludes, and
7097 tracks creating the variables.

### `openai_responses_agent` is misconfigured on `main` too — QG4 has the same bug

Worth flagging because it is not a QG7 defect and QG7 is only where it became visible.
`OPENAI_BASE_URL` and `OPENAI_MODEL_ID` do not exist as repo variables (`gh variable list`), and
QG4 resolves that agent with the same fallback (`agent-deployment-test.yaml:106-107`):

```yaml
API_KEY:  ${{ matrix.agent.name == 'vanilla-python-openai-responses-agent' && secrets.OPENAI_API_KEY || vars.API_KEY }}
BASE_URL: ${{ matrix.BASE_URL || vars.BASE_URL }}   # matrix.BASE_URL = vars.OPENAI_BASE_URL
```

Both collapse to the cluster vLLM (`http://vllm-svc.llama-serving.svc:8000/v1`, `not_needed`), so
QG4 has been health-checking this agent against the wrong endpoint nightly and passing. QG4 only
probes `/health`, which returns 200 regardless — the misconfiguration is invisible there. QG7 runs
behavioral tests, where it would show.

Neither gate *detected* this; it was found reading the expression. Aligning QG4 is deliberately
left out of scope — see RHAIENG-7097.

### Why two agents are excluded rather than fixed here

Both are **pre-existing defects on `main` that this PR merely exposes**, invisible until now
because nothing in CI ever built these agents fresh and then probed them — QG4 was health-checking
stale pods (this is false-green 1, below). Neither can be tested honestly today, and both are
cluster/dependency problems rather than QG7 problems. `EXCLUDED_AGENTS` already exists for exactly
this situation and already holds `guardrailed_agent` (RHAIENG-7081) on the same reasoning: skip
and ticket rather than fake green.

Both tickets carry an acceptance criterion to remove the agent from `EXCLUDED_AGENTS` and confirm
it passes QG7, so the exclusion cannot quietly become permanent. The `AGENTS` array keeps all 12
entries and stays the single source of truth; exclusion is a filter applied on top, and
`test_excluded_agents_are_all_in_the_runner_array` guards against the two lists drifting.

## Commits

| Commit | What |
|---|---|
| `5162d31` | `ci(qg7)`: deploy and tear down btest agents within the gate — branch base |
| `442dbbf` | `fix(scripts)`: guard empty namespace array under bash 3.2 — **separable, not this branch's bug** |
| `f04755f` | `fix(qg7)`: poll for the tracing marker instead of reading once |
| `69f7237` | `fix(qg7)`: close four false-greens found on a live cluster |
| `8bc1c59` | `fix(qg7)`: close two false-greens a namespace-scoped CI token exposes |
| `810a0e9` | `fix(qg7)`: exclude two untestable agents and close the alias fallback |
| `b3d940b` | `fix(qg7)`: skip an agent whose aliased env vars are unset, don't fail the gate |

`442dbbf` fixes `scripts/openshift-build-config.sh` under macOS bash 3.2, where
`ns_flag[@]: unbound variable` broke `make build-openshift` for every agent. Pre-existing since
#306 (`57a59fd`); CI on ubuntu/bash 5 is unaffected. It is deliberately its own commit so it can
be split out or cherry-picked to `main` independently.

---

# Testing record

Cluster `agen-e2e-rhoai2`, RHOAI 3.5. Stages 0–3 in `adonheis-testing` as cluster-admin;
Stage 4 on GitHub Actions in a private throwaway repo against a throwaway namespace with a
dedicated namespace-scoped ServiceAccount.

## Stage 0 — static gate: PASS

`ruff check`, `ruff format --check`, `pytest tests/behavioral/deterministic/ -q`,
`--print-selection`. Re-run after the exclusions and the env skip: **57 passed**, ruff clean, all
three YAML files parse, and `--print-selection` emits **8** ids with `mcp_agent`, `agentic_rag`,
`guardrailed_agent` and `openai_responses_agent` all absent — the last of those skipped with the
warning naming `OPENAI_API_KEY, OPENAI_BASE_URL, OPENAI_MODEL_ID`, and it returns to **9** once
those are set. (Stages 1–4 below were run pre-exclusion, against 11 agents.)

## Stage 1 — react_agent round trip: PASS

- `MLFLOW_TRACKING_TOKEN` on the deployment is a `valueFrom.secretKeyRef`, not a literal.
- `MLFLOW_EXPERIMENT_NAME=adonheis-testing/langgraph-react-agent`, `MLFLOW_WORKSPACE=adonheis-testing`.
- Re-deploy produced no `conflict with "kubectl-patch"` — Helm field ownership survives the token
  refresh (`--field-manager=helm` server-side-apply holds).
- `.env` stash/restore round-trips cleanly; no `.env.qg7.bak` left behind (md5 verified).
- Runner Phase 2 printed a real `MLFLOW_TRACKING_URI=…`, not the "not found in deployment env
  vars" warning. **This is the exact failure the change exists to fix, and it is fixed.**
- Runner 22 passed / 3 failed — **pre-existing**, identical test names and `tool_calls` warning
  count to the prior hand-deployed run.
- Build + deploy ≈ 4.5 min.

## Stage 2 — non-standard shapes: PASS

- **a2a:** one `make deploy` yielded both `a2a-crew-agent` and `a2a-langgraph-agent`, both probed
  via `/.well-known/agent-card.json`, both removed by a single `--undeploy`. `CONTAINER_IMAGE`
  alone correctly drove the `:crew`/`:langgraph` tagging. ≈ 5.5 min.
- **langflow:** probe-only via `/health_check`. `helm list -n langflow-agent` byte-identical
  before and after deploy + undeploy — nothing built, deployed, or torn down.

### Timing — the headline number

| Phase | Wall clock |
|---|---|
| Full-set deploy (11 agents, 4-wide pool) | **20.1 min** (1204s) |
| Runner over all 11 | **8.2 min** (491s) |
| Full teardown | **17s** |
| **Total** | **≈ 28 min against a `timeout-minutes: 150` budget** |

The `~20 min/agent` estimate behind that budget is off by roughly 7x.

### Selection plumbing

All 11 selected agents reached the runner. **No `NO_ROUTE` row**; `guardrailed_agent` absent from
both the deploy pass and the runner argv.

```
[FAIL] crewai/templates/websearch_agent               (11 passed,  6 failed, 2 skipped)
[FAIL] langgraph/templates/react_agent                (22 passed,  3 failed, 0 skipped)
[FAIL] langgraph/templates/agentic_rag                ( 1 passed, 18 failed, 0 skipped)
[FAIL] langgraph/templates/react_with_database_memory (16 passed,  3 failed, 0 skipped)
[FAIL] autogen/templates/mcp_agent                    (15 passed,  1 failed, 2 skipped)
[FAIL] llamaindex/templates/websearch_agent           (18 passed,  2 failed, 0 skipped)
[FAIL] vanilla_python/templates/openai_responses_agent(24 passed,  3 failed, 0 skipped)
[FAIL] langgraph/templates/human_in_the_loop          (15 passed,  1 failed, 0 skipped)
[FAIL] google/templates/adk                           (21 passed,  1 failed, 0 skipped)
[PASS] langflow/templates/simple_tool_calling_agent   (29 passed,  0 failed, 0 skipped)
[FAIL] a2a/templates/langgraph_crewai_agent           (12 passed,  6 failed, 1 skipped)
```

These are agent-quality results, not QG7 results — QG7's job is to get them deployed and traced,
which it did. `react_agent`'s 22/3 matches every other baseline exactly.

### Teardown, including partial failure

- Full `--undeploy` removed all 10 deployed agents in 17s.
- **Partial-failure case:** deployed 2 agents, `kill -9` mid-probe. `--undeploy` then read
  `${RUNNER_TEMP:-/tmp}/qg7-deployed.txt`, removed **exactly the 2 recorded agents** (not the
  11-agent selection), deleted the record, *and* restored both `.env` files from the orphaned
  `.qg7.bak` backups the hard kill left behind. `git status` clean afterwards.

### Correction to the plan's expectation

**The pool is not fail-fast.** `autogen/templates/mcp_agent` failed and the remaining agents still
deployed to completion. The plan predicted one bad agent would stop the rest; it does not.

## Stage 4 — CI wiring on a real runner: PASS

Nothing in Stages 1–3 executes `run-qg7/action.yml`. Stage 4 ran the real workflow on
`ubuntu-latest` in a private throwaway repo, logging in with a **dedicated ServiceAccount holding
only the namespaced `admin` role** on a throwaway `qg7-ci-test` namespace — deliberately *not* a
cluster-admin token, and the shape real CI actually has. Repo, SA, and namespace were destroyed
afterwards.

All four objectives confirmed:

| Objective | Evidence |
|---|---|
| `$(...)` capture yields clean ids (script logs to stderr, ids to stdout) | `QG7_SELECTED_AGENTS: langgraph/templates/react_agent` — no stderr leak into the value |
| Runner receives `QG7_AGENT_NAME` from `steps.selection.outputs.agents` | `QG7_AGENT_NAME: langgraph/templates/react_agent` on the runner step |
| **`if: always()` teardown actually fires** | Fired after a **failing runner** step, and again after a **failing deploy** step with the runner skipped. Namespace empty, zero helm releases, both times |
| Runner authenticates with no token flag (`oc whoami -t` → `OC_TOKEN`) | Build, deploy, route, and a live MLflow token all succeeded from the SA login |

Full round trip under the SA: build 1m43s → deploy → `/health` 200 → `[Tracing Enabled]` →
runner **22 passed / 3 failed** (matching every other baseline) → teardown 1.1s, namespace empty.
Deployment env verified on-cluster: `MLFLOW_EXPERIMENT_NAME=qg7-ci-test/langgraph-react-agent`,
`MLFLOW_WORKSPACE=qg7-ci-test`, and `MLFLOW_TRACKING_TOKEN` as a `secretKeyRef` rather than a
literal.

**Teardown guard:** a `run_cluster_btests=false` run completed in 3s executing *only*
`Resolve cluster btests config`. Checkout, setup-cluster, and `run-qg7` were never entered, so
the teardown cannot run against a logged-out runner.

---

# Seven false-greens found and fixed

Every one is a case where the gate reported success against something that was not working.

### 1. `deploy_agent` never waited for the rollout

`make deploy` returns while the previous generation's pod is still Ready and still serving the
route, so the readiness probe *and* the tracing gate validate the **old** pod.

Caught red-handed: `autogen-mcp-agent`'s new image crash-looped on
`ImportError: cannot import name 'RequestContext' from 'mcp.shared.context'` while a 27-hour-old
pod answered `/health` with `200 OK`. Both gates passed.

**Fix:** `wait_for_rollout()` + `DEFAULT_ROLLOUT_TIMEOUT`, called for every deployment between
`make deploy` and route resolution.

### 2. Tracing markers were exact-string matches

`_TRACING_OK = "[Tracing Enabled]"` never matches a2a, which logs `[Tracing Enabled LangGraph]` /
`[Tracing Enabled CrewAI]`. Both a2a deployments always fell through to the token-only fallback —
which proves the token works, not that the agent traces.

**Fix:** compiled regexes `\[Tracing Enabled[^\]]*\]` and `\[Tracing[^\]]*\] Failed to configure`.

### 3. Route discovery picked a dead route

With the namespace emptied, resolution fell through to `_tracking_uri_from_route()`, which chose
`https://mlflow-redhat-ods-applications.apps.…` — that route exists but **404s**. RHOAI 3.5 serves
MLflow through the data-science gateway at `https://rh-ai.apps.…/mlflow`. Every agent got a dead
URI and nothing traced; the tracing gate then blamed the token.

The full-set run only worked because link 2 of the chain read the correct URI off a pre-existing
deployment. Route discovery had never actually been exercised.

**Fix:** `_tracking_uri_candidates()` offers gateway-with-`/mlflow` first, then the standalone
route, and `_tracking_uri_from_route()` picks the first whose `/api/3.0/mlflow/server-info`
returns JSON.

### 4. `agentic_rag` got the shared `BASE_URL`

Its vector store lives behind the llama-stack/OGX endpoint, so retrieval 404'd on
`/v1/vector-io/query` and 18 of 19 tests failed. `/health` never touches the vector store, so
every deploy gate passed.

**Fix:** `ENV_SOURCE_ALIASES["langgraph/templates/agentic_rag"]` maps `BASE_URL`/`MODEL_ID` to
`OGX_*`, plus the matching entries in the `eval-gating.yml` env block. (This fix originally let an
unset source fall back to the shared value — that fallback is itself false-green 7, below.)

### 5. Forbidden CRD read was reported as "RHOAI < 3.5" *(found in Stage 4)*

CRDs are cluster-scoped and a namespace-scoped CI ServiceAccount cannot read them.
`mlflow_operator_present()` collapsed "not permitted" into "absent", logged
`MLflow operator CRD absent (RHOAI < 3.5) — skipping namespace label` **on a 3.5 cluster**, and
skipped the label. Agents then come up untraced with `Workspace '<ns>' not found`, long after the
deploy gate went green. This is invisible to any local cluster-admin run.

**Fix:** the probe is tri-state (`True` / `False` / `None` for unknown). When presence is unknown
the namespace's own label is checked — readable by a namespace admin even when CRDs are not — and
an unlabelled namespace that cannot be labelled now raises with the exact out-of-band command to
run.

*Verified live:* with the label removed, the run failed in **3 seconds** with
`Namespace qg7-ci-test is not labelled mlflow-tracking=enabled … Pre-label the namespace out of
band: oc label namespace qg7-ci-test mlflow-tracking=enabled`. With the label restored, it logged
`namespace qg7-ci-test is already labelled` and deployed normally.

### 6. Route discovery is unavailable to the CI token at all *(found in Stage 4)*

`oc get route -A` is also cluster-scoped and forbidden to the CI SA, so fix 3's route discovery
returns no candidates in CI. That leaves the entire resolution chain resting on link 2 — reading
the URI off a pre-existing deployment — so the **first run against an empty `ci-testing` would
fail the whole deploy pass**, with an error blaming a missing route rather than the credential.

**Fix:** the forbidden case now warns explicitly, the resolution error names the cause, and
`eval-gating.yml` passes `MLFLOW_TRACKING_URI` so CI does not depend on discovery. Unset still
falls back to discovery, which is correct for a cluster-admin.

### 7. An empty aliased variable fell back to the shared endpoint

**This one reverses a deliberate earlier decision in this branch, so it deserves a close read.**

`build_env_map`'s `read()` helper resolved an `ENV_SOURCE_ALIASES` entry and, when the alias was
empty, fell through to the plain variable — mirroring QG4's `${{ matrix.BASE_URL || vars.BASE_URL }}`.
The test that covered it said so in its docstring.

That `||` is the false green. GitHub substitutes an **empty string** for an undefined variable,
silently. None of `OPENAI_API_KEY` / `OPENAI_BASE_URL` / `OPENAI_MODEL_ID` exist as repo
variables, so `vanilla_python/templates/openai_responses_agent` resolved to the shared cluster
vLLM and **passed** — an agent whose entire point is that it talks to OpenAI, greened against an
endpoint it is not meant to use. Confirmed live in Stage 4's logs.

An alias entry exists *precisely because* the shared value is wrong for that agent, so falling
back to it is never the right recovery.

**Fix:** `read()` returns `(value, source_name)` and an empty alias is reported as missing under
**the alias's own name** — the failure now says `missing required env vars: OPENAI_API_KEY,
OPENAI_BASE_URL, OPENAI_MODEL_ID`, naming what to set rather than the `agent.yaml` var. Two
existing tests asserted the old fallback and were rewritten to assert the failure; a third covers
the empty-string case distinctly from the absent-var case, since GitHub produces the former.

This does not remove the need for the `OPENAI_*` variables (RHAIENG-7097) — it makes their
absence loud instead of silent. `b3d940b` then made it *non-fatal*: selection skips such an agent
rather than failing the run, matching QG4's `if: … || vars.MCP_SERVER_URL != ''`. The raise stays
as a backstop, so an agent that somehow reaches deploy unconfigured still cannot fall back.

---

# Follow-ups — all ticketed

Nothing here stays only in a PR comment. Every item has a Jira issue under RHAIENG-6074.

| Ticket | Status | Item |
|---|---|---|
| RHAIENG-7096 | **Done** | `MLFLOW_TRACKING_URI` repo variable — set and verified. Fix 6 makes CI read it; without it the gate falls back to discovery, which the CI SA cannot perform. |
| **RHAIENG-7097** | Not blocking | `OPENAI_API_KEY` (secret), `OPENAI_BASE_URL`, `OPENAI_MODEL_ID` do not exist, so `openai_responses_agent` is skipped and the gate runs 8 agents. **QG4 has the same misconfiguration** and passes it nightly against the cluster vLLM — aligning QG4 is tracked here too. |
| RHAIENG-7094 | Excluded, ticketed | `autogen/templates/mcp_agent` crash-loops on a fresh build — `mcp.shared.context.RequestContext` is gone from the installed `autogen_ext`. Pre-existing on `main`, masked by QG4 health-checking a stale pod. |
| RHAIENG-7095 | Excluded, ticketed | `agentic_rag` retrieval fails — `VECTOR_STORE_ID` names a store the OGX llama-stack does not have (`Vector_store 'vs_f91e065e-…' not found`); it currently has zero stores. Cluster data, not code. `ENV_SOURCE_ALIASES` already routes it to the OGX endpoint and is retained. |
| RHAIENG-7098 | Deferred | The other seven variables the `env:` block references (`MCP_SERVER_URL`, `EMBEDDING_*`, `VECTOR_STORE_*`, `OGX_*`). **Not blockers:** grepping `agent.yaml` across `agents/` shows each is consumed only by one of the two excluded agents, so none can affect the 9 that run. Blocks 7094 and 7095. |
| RHAIENG-7099 | Deferred | `ci-testing` has `postgres` but no `mcp-automl`. Blocks re-enabling 7094. |
| RHAIENG-7100 | Deferred | a2a's two deployments share one MLflow experiment name — both write to `…/a2a-langgraph-agent`, cross-contaminating traces. Same class as the closed RHAIENG-6743. |
| RHAIENG-7101 | Deferred | QG4/QG7 shared concurrency group for `ci-testing`. QG4 (nightly 03:00 UTC) deploys identical deployment names and undeploys in a `finally`, so an overlapping run tears QG7's agents down mid-suite. Requires a QG4 edit, outside this PR's scope bar. **Recognisable signature:** agents that passed deploy and health probes reporting `NO_ROUTE` partway through the suite. |

`gh variable list` currently has only `API_KEY`, `BASE_URL`, `MODEL_ID`, `LANGFLOW_AGENT_URL`,
`POSTGRES_{HOST,PORT,DB,USER}` — ten of the workflow's referenced variables are missing, split
across RHAIENG-7097 (three, blocking) and RHAIENG-7098 (seven, not blocking).


mpk-droid:
**Token subsystem: compounding credential leak risk**

Three small weaknesses in how `MLFLOW_TRACKING_TOKEN` is handled that together create a real leak vector:

1. **Missing redaction** — `_REDACT_PATTERNS` in `tests/integration/utils.py` masks `API_KEY`, `POSTGRES_PASSWORD`, etc., but not `MLFLOW_TRACKING_TOKEN`. If the token appears in an `oc` error message, it goes into CI logs in plain text.
2. **Untested refresh code** — `refresh_mlflow_token()`, `mlflow_token_secret_name()`, `_apply_token_to_secret()`, and `HardcodedTokenError` have zero unit tests. If a bug in these functions puts the token into an error message, no test catches it.
3. **Manual first-run setup** — `MLFLOW_TRACKING_URI` must be set by hand (the CI SA can't discover routes). Manual steps increase the chance of hitting error paths — the very paths that aren't tested and won't redact.

Any one alone is manageable. Together: untested code hits an error path → token in the message → redaction doesn't catch it → token visible in CI logs.

**Suggested fix:** Add `MLFLOW_TRACKING_TOKEN` to `_REDACT_PATTERNS`, and add unit tests for the token refresh functions (especially error paths).

tests/behavioral/deterministic/deploy_agents.py
```diff
@@ -0,0 +1,1086 @@
...
+            "-n",
+            namespace,
+            "-f",
+            "-",
+        ],
+        stdin=json.dumps(secret),
+        timeout=60,
+    )
+
+
```
mpk-droid, line 859: **Missing test coverage for the token refresh subsystem**

`refresh_mlflow_token()`, `mlflow_token_secret_name()`, `_apply_token_to_secret()`, and `HardcodedTokenError` all lack unit tests — the rest of the module has excellent coverage (52 tests), so this is a gap.

The critical bit: `_apply_token_to_secret` uses `--server-side --field-manager=helm` to preserve Helm field ownership. If someone later refactors and drops that flag, no test catches it — the next `helm upgrade` fails with a conflict error in CI, not in a test run.

Suggested tests:
- `mlflow_token_secret_name` returns the correct secret name from deployment JSON
- `mlflow_token_secret_name` raises `HardcodedTokenError` when the token is a literal value
- `_apply_token_to_secret` passes `--field-manager=helm` to `_oc`
- `refresh_mlflow_token` deduplicates secret writes when two deployments share one secret
- `refresh_mlflow_token` skips deployments with no `secretKeyRef`

.github/workflows/eval-gating.yml
```diff
@@ -47,9 +47,66 @@ jobs:
...
+    # triggers on both push and pull_request to main, so two PRs in flight would
+    # otherwise deploy and undeploy identical deployment names concurrently.
+    # Not cancel-in-progress: a cancelled run would skip its teardown.
+    #
+    # Residual risk, accepted: QG4 (nightly, 03:00 UTC) deploys the same
+    # deployment names into ci-testing and undeploys in a `finally`. A QG4 run
+    # overlapping a QG7 run will tear QG7's agents down mid-suite. Fixing that
+    # needs a QG4 edit, tracked as RHAIENG-7101.
+    concurrency:
+      group: cluster-btests-ci-testing
```
mpk-droid, line 64: **Concurrency group only protects eval-gating runs**

This group prevents two `eval-gating.yml` runs from colliding in `ci-testing`, but it doesn't protect against the orchestrator pipeline (`quality-gates-pipeline.yml` from PR #328) running concurrently. A manual `workflow_dispatch` of the orchestrator overlapping with the nightly cron would deploy and undeploy in the same namespace simultaneously.

Low risk — cron fires once/day and manual dispatch is deliberate — but worth noting. Same class of accepted risk as RHAIENG-7101 (QG4 overlap). Consider whether the orchestrator should use the same concurrency group name.

mpk-droid (changes requested):

mpk-droid:
**Remaining from the token subsystem review: `MLFLOW_TRACKING_TOKEN` not in redaction patterns**

The new token refresh tests (a625815) close the untested-code leg of the compound risk — nice work. One piece remains: `_REDACT_PATTERNS` in `tests/integration/utils.py` masks `API_KEY`, `NVIDIA_API_KEY`, `POSTGRES_PASSWORD`, etc., but not `MLFLOW_TRACKING_TOKEN`.

The `_oc()` helper in `deploy_agents.py` passes error output through `_redact()`, so if an `oc` command fails and the token appears in stderr, it would be logged in plain text. Low probability (the token usually lives in a secret, not in command output), but easy to close:

```python
# in tests/integration/utils.py, add to _REDACT_PATTERNS:
re.compile(r"(MLFLOW_TRACKING_TOKEN=)\S+"),
```

mpk-droid (changes requested):
Overall looks good. Thanks for addressing the comments. Just one more comment left. if you could address, will approve the PR.

mpk-droid (approved):
lgtm

PR #330, opened 2026-08-26
fix(mcp_automl): pin mcp<2 to fix crash on fresh builds

andrewdonheiser:
## Summary

- Pins `mcp>=1.0.0,<2.0.0` in `agents/autogen/templates/mcp_agent/mcp_automl_template/pyproject.toml`
- Same root cause as #329 but for the MCP server package: mcp 2.x renamed `FastMCP` to `MCPServer`, so fresh builds crash-loop with `ModuleNotFoundError: No module named 'mcp.server.fastmcp'`
- Regenerated `uv.lock` to reflect the constraint
- Discovered while deploying mcp-automl to `ci-testing` for RHAIENG-7099
- extends fix in RHAIENG-7094

## Test plan

- [x] Fresh build in ci-testing with pinned dep installs `mcp==1.29.1` and server starts cleanly
- [x] Health endpoint returns `{"status":"healthy"}`
- [x] In-cluster reachability confirmed via `curl http://mcp-automl.ci-testing.svc:8080/health`
- [ ] CI passes


tarun-etikala (approved):
lgtm

tarun-etikala:
**Suggestion (non-blocking):** the `<2` pin is the right fix for this crash and LGTM to merge.

One follow-up worth tracking: this image still does not install from `uv.lock`. The Dockerfile copies `pyproject.toml` and runs `uv pip install --no-cache .`, so a fresh build re-resolves ranges from PyPI. That is how mcp 2.x got in; the same thing can happen on the next uncapped major (`langgraph>=1.0.0`, `pyyaml>=6.0`, etc.).

Longer-term:

1. **Image installs from the lockfile** — `COPY uv.lock` and `uv sync --frozen --no-dev` (repo-wide; every agent Dockerfile uses `uv pip install` today). That makes the committed lockfile actually apply to `make build`, which is what CONTRIBUTING already assumes.
2. **Keep `mcp<2` until the server is migrated** — `mcp_server.py` still imports `FastMCP` from `mcp.server.fastmcp`. Moving to `MCPServer` is how we can take mcp 2.x; the pin is the compatibility contract until then.

PR #332, opened 2026-08-31
ci(qg7): handle Langflow flow-import agent in pytest runner preflight

andrewdonheiser:
## Summary

Fix QG7 behavioral test runner preflight errors for non-standard agents like Langflow that don't have Kubernetes Deployments. The pytest runner now correctly:
- Detects flow-import agents via `agent.yaml` configuration
- Checks route + health endpoint instead of deployment for these agents
- Skips them when searching for MLflow configuration (they use Langfuse)

## Changes

- Add `is_flow_import()` helper to identify flow-import agents
- Preflight: handle flow-import agents separately with route + `/health_check` check
- MLflow detection: skip flow-import agents (they use Langfuse, not MLflow)
- Per-agent experiment discovery: skip deployment lookup for flow-import agents
- Preserve existing Langflow flow ID discovery logic unchanged

Standard agents remain unaffected by these changes.

## Test Results

- ✓ `is_flow_import()` detects Langflow as flow-import
- ✓ `is_flow_import()` skips standard agents (react_agent)
- ✓ Regex pattern ignores commented-out lines
- ✓ `is_flow_import` used in all required locations (definition + 3 call sites)

## Files Changed

- `tests/behavioral/deterministic/run-btests-pytest.sh` — add flow-import agent support to preflight checks

Fixes RHAIENG-7219.

**Depends on:** RHAIENG-7218 (empty agent selection fix). When the prereq merges, rebase this branch onto main.

---

### DIFF
--- a/tests/behavioral/deterministic/run-btests-pytest.sh
+++ b/tests/behavioral/deterministic/run-btests-pytest.sh
@@ -88,6 +88,12 @@ agent_env_var()   { echo "$1" | cut -d'|' -f2; }
 agent_deploy()    { echo "$1" | cut -d'|' -f3; }
 agent_ns()        { local ns; ns=$(echo "$1" | cut -d'|' -f4); echo "${ns:-${NAMESPACE}}"; }
 
+is_flow_import() {
+  local path="$1"
+  local agent_yaml="${REPO_ROOT}/agents/${path}/agent.yaml"
+  [[ -f "$agent_yaml" ]] && grep -qE '^\s*deploymentModel:\s*flow-import' "$agent_yaml"
+}
+
 validate_agent_url_map_sync() {
   local agent_tuples=("$@")
   if [[ ${#agent_tuples[@]} -eq 0 ]]; then
@@ -218,6 +224,28 @@ preflight() {
     local path
     path=$(agent_path "$agent_tuple")
 
+    # Flow-import agents have no Kubernetes Deployment; check route + health endpoint instead
+    if is_flow_import "$path"; then
+      local route_host
+      route_host=$(timeout 30 oc get route "${deploy}" -n "${ns}" \
+        -o jsonpath='{.spec.host}' 2>/dev/null || true)
+
+      if [[ -z "$route_host" ]]; then
+        fail "Route '${deploy}' not found for flow-import agent '${path}' (namespace: ${ns})"
+        all_healthy=false
+        continue
+      fi
+
+      # Curl the health endpoint
+      if curl -sk "${route_host}/health_check" >/dev/null 2>&1; then
+        ok "Route '${deploy}' (flow-import) — healthy"
+      else
+        warn "Route '${deploy}' (flow-import) — health check failed or endpoint unreachable"
+        all_healthy=false
+      fi
+      continue
+    fi
+
     if ! timeout 30 oc get deployment "${deploy}" -n "${ns}" >/dev/null 2>&1; then
       fail "Deployment '${deploy}' not found for agent '${path}' (namespace: ${ns})"
       all_healthy=false
@@ -255,9 +283,16 @@ detect_mlflow_config() {
   # Use the first available agent deployment to extract MLflow env vars
   local deploy_json=""
   for agent_tuple in "${AGENTS[@]}"; do
-    local deploy ns
+    local deploy ns path
     deploy=$(agent_deploy "$agent_tuple")
     ns=$(agent_ns "$agent_tuple")
+    path=$(agent_path "$agent_tuple")
+
+    # Skip flow-import agents (they use Langfuse, not MLflow)
+    if is_flow_import "$path"; then
+      continue
+    fi
+
     deploy_json=$(timeout 30 oc get deployment "${deploy}" -n "${ns}" -o json 2>/dev/null || true)
     if [[ -n "$deploy_json" ]]; then
       log "Using deployment '${deploy}' for MLflow config detection"
@@ -352,10 +387,12 @@ run_tests() {
 
     log "Launching: ${BOLD}${path}${RESET} -> ${logfile}"
 
-    # Detect per-agent experiment name from its deployment
+    # Detect per-agent experiment name from its deployment (skip for flow-import agents)
     local agent_experiment
-    agent_experiment=$(timeout 30 oc get deployment "${deploy}" -n "${ns}" \
-      -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="MLFLOW_EXPERIMENT_NAME")].value}' 2>/dev/null || true)
+    if ! is_flow_import "$path"; then
+      agent_experiment=$(timeout 30 oc get deployment "${deploy}" -n "${ns}" \
+        -o jsonpath='{.spec.template.spec.containers[0].env[?(@.name=="MLFLOW_EXPERIMENT_NAME")].value}' 2>/dev/null || true)
+    fi
     agent_experiment="${agent_experiment:-${MLFLOW_EXPERIMENT_NAME:-}}"
 
     # Langflow needs the flow ID discovered from its API
--- a/tests/behavioral/deterministic/run-btests-pytest.sh
+++ b/tests/behavioral/deterministic/run-btests-pytest.sh
@@ -237,7 +237,7 @@ preflight() {
       fi
 
       # Curl the health endpoint
-      if curl -sk "${route_host}/health_check" >/dev/null 2>&1; then
+      if curl -fsSk --connect-timeout 10 --max-time 30 "https://${route_host}/health_check" >/dev/null 2>&1; then
         ok "Route '${deploy}' (flow-import) — healthy"
       else
         warn "Route '${deploy}' (flow-import) — health check failed or endpoint unreachable"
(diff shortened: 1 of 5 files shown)


tests/behavioral/deterministic/run-btests-pytest.sh
```diff
@@ -218,6 +224,28 @@ preflight() {
...
+        -o jsonpath='{.spec.host}' 2>/dev/null || true)
+
+      if [[ -z "$route_host" ]]; then
+        fail "Route '${deploy}' not found for flow-import agent '${path}' (namespace: ${ns})"
+        all_healthy=false
+        continue
+      fi
+
+      # Curl the health endpoint
+      if curl -sk "${route_host}/health_check" >/dev/null 2>&1; then
```
mpk-droid: **\[critical\]** This curl has three compounding issues:

1. **Missing `https://`** — `route_host` is just a hostname (e.g. `langflow.apps.cluster.example.com`). Without a protocol prefix curl defaults to HTTP, which will fail on HTTPS-only OpenShift routes. Line 373 in this same script correctly uses `"https://${route_host}"`.

2. **Missing `-f` flag** — Without it, curl returns exit 0 even on 404/500 responses. A broken agent would pass the health check. The Python equivalent uses `resp.raise_for_status()`.

3. **No timeout** — An unreachable endpoint blocks CI for ~120s (OS TCP timeout). Other `oc` calls in this script all use `timeout 30`.

Suggested fix:
```bash
if curl -fsSk --connect-timeout 10 --max-time 30 "https://${route_host}/health_check" >/dev/null 2>&1; then
```

tests/behavioral/deterministic/run-btests-pytest.sh
```diff
@@ -88,6 +88,12 @@ agent_env_var()   { echo "$1" | cut -d'|' -f2; }
 agent_deploy()    { echo "$1" | cut -d'|' -f3; }
 agent_ns()        { local ns; ns=$(echo "$1" | cut -d'|' -f4); echo "${ns:-${NAMESPACE}}"; }
 
+is_flow_import() {
+  local path="$1"
+  local agent_yaml="${REPO_ROOT}/agents/${path}/agent.yaml"
+  [[ -f "$agent_yaml" ]] && grep -qE '^\s*deploymentModel:\s*flow-import' "$agent_yaml"
+}
```
mpk-droid, line 95: **[medium]** The bash `is_flow_import()` function has no automated test coverage. The PR description lists test results ("is_flow_import() detects Langflow as flow-import", "skips standard agents", "regex ignores commented-out lines"), but these were manual checks — no tests were committed.

The Python equivalent in `deploy_agents.py` has proper tests (`test_langflow_is_flow_import`, `test_standard_agent_has_no_deployment_model`). The bash version that actually runs in CI should have parity.

`test_run_btests_pytest.py` already tests other bash functions from this script via `BTEST_LIB_ONLY=1` sourcing — the same pattern works here.

tests/behavioral/deterministic/run-btests-pytest.sh
```diff
@@ -218,6 +224,28 @@ preflight() {
...
+      fi
+
+      # Curl the health endpoint
+      if curl -sk "${route_host}/health_check" >/dev/null 2>&1; then
+        ok "Route '${deploy}' (flow-import) — healthy"
+      else
+        warn "Route '${deploy}' (flow-import) — health check failed or endpoint unreachable"
+        all_healthy=false
+      fi
+      continue
```
mpk-droid, line 246: **[medium]** This entire flow-import preflight branch (route lookup → health check → report) is the core acceptance criterion from RHAIENG-7219 but has no automated test coverage. If someone refactors the preflight loop later, nothing in CI catches a regression here.

`test_run_btests_pytest.py` already tests other parts of this script — consider adding a case that verifies the preflight takes the route+health path (not `oc get deployment`) for a flow-import agent.

mpk-droid:
**[medium]** The `qg7_excluded` list in `quality-gates-pipeline.yml` (line 202) hardcodes the same agents as `EXCLUDED_AGENTS` in `deploy_agents.py` (lines 68-80). The comment says "mirrors deploy_agents.EXCLUDED_AGENTS" but there's no mechanism to keep them in sync — if someone updates one list and forgets the other, the QG4→QG7 handoff will pass agents that the deploy step skips (or vice versa), causing confusing CI failures.

Consider either:
- A `--print-excluded` flag on `deploy-btest-agents.sh` so the workflow reads the list dynamically
- A CI check that fails if the two lists diverge

_(This comment is about changes that were in the original PR but have since been moved to a separate commit. Flagging for awareness in case they return.)_

mpk-droid (changes requested):
added few comments that needs addressing.

tests/behavioral/deterministic/run-btests-pytest.sh
andrewdonheiser: Fixed in d3d5650. Updated to:
```bash
curl -fsSk --connect-timeout 10 --max-time 30 "https://${route_host}/health_check"
```

All three issues addressed:
1. Added `https://` prefix to match line 373
2. Added `-f` flag so curl fails on HTTP 4xx/5xx
3. Added `--connect-timeout 10 --max-time 30` to bound stalled requests

tests/behavioral/deterministic/run-btests-pytest.sh
andrewdonheiser: Fixed in d3d5650. Added two tests using the existing `BTEST_LIB_ONLY=1` sourcing pattern:
- `test_is_flow_import_detects_langflow_agent` — confirms langflow agent is detected as flow-import
- `test_is_flow_import_rejects_standard_agent` — confirms react_agent returns "no"

tests/behavioral/deterministic/run-btests-pytest.sh
andrewdonheiser: Fixed in d3d5650. Added two tests:
- `test_preflight_uses_route_health_check_for_flow_import_agent` — verifies preflight takes the route+health path (not `oc get deployment`) for flow-import agents
- `test_detect_mlflow_config_skips_when_all_agents_are_flow_import` — verifies flow-import-only selections skip MLflow detection instead of dying

Also added the guard in `detect_mlflow_config` so flow-import-only selections warn and return instead of hitting "No agent deployments found".

andrewdonheiser:
Re: exclusion list sync — fixed in 8a11bab.

Replaced the hardcoded `qg7_excluded` JSON in `quality-gates-pipeline.yml` with a sparse checkout + AST extraction that reads `EXCLUDED_AGENTS` from `deploy_agents.py` at runtime. The two lists can no longer drift.

Also added `--print-excluded` flag to `deploy-btest-agents.sh` for local use (`deploy-btest-agents.sh --print-excluded` outputs the JSON array).

mpk-droid (dismissed):
lgtm

mpk-droid (dismissed):
lgtm

mpk-droid (dismissed):
lgtm

kami619:
Thermo-nuclear maintainability review @andrewdonheiser, PTAL.

I would not merge this as-is. The flow-import handling is directionally right, but the current diff has two structural issues that make the CI path harder to trust.

1. Blocking: `collect-qg4` downloads `qg4-outcomes/` and then runs `actions/checkout`. Checkout defaults to cleaning the workspace, so it can delete the downloaded artifacts before `Build pass-list from QG4 outcomes` reads them. That turns a real QG4 result set into `pass_list=[]` and silently skips QG7. Move checkout before artifact download, or set `clean: false` if there is a reason the checkout must remain after download.

2. The `qg7_excluded` fix still feels too clever. This PR adds `deploy-btest-agents.sh --print-excluded`, then the workflow bypasses it and embeds an AST parser inside YAML. That removes one duplicate list but replaces it with a hidden second implementation of “how to read the Python constant.” The code-judo move is to make the workflow call a single supported interface. If `deploy-btest-agents.sh --print-excluded` cannot run there because it imports heavy deps, split the exclusion export into a tiny dependency-free helper or move the exclusion data into a small JSON/config file consumed by both paths. Do not keep both the unused flag and inline AST reflection in the workflow.

3. The current diff includes unrelated CI churn: action-version bumps, QG4 timeout/max-parallel changes, and repeated integration fixture timeout edits. Those are not part of “handle Langflow flow-import agent in pytest runner preflight,” and they make this PR much harder to reason about. Please rebase/drop already-merged or unrelated changes so this PR only contains the flow-import/QG7 selection work.

After those are addressed, the shell-side flow-import branch itself looks reasonable: route + `/health_check` for Langflow, skip deployment-derived MLflow detection for flow-import-only selections, and avoid Deployment lookups for per-agent MLflow experiment discovery.

kami619:
Current state: open, non-draft, 5 files changed, +137/-6, head branch ci/RHAIENG-7219-qg7-langflow-preflight.

  The latest diff is much cleaner than the earlier noisy version: the unrelated dependency bumps and repeated integration timeout edits are no longer in the active file set. The PR now focuses on:

  - .github/workflows/quality-gates-pipeline.yml: dynamic QG7 exclusion filtering
  - tests/behavioral/deterministic/run-btests-pytest.sh: flow-import route/health handling and MLflow skip logic
  - tests/behavioral/deterministic/deploy_btest_agents.py: new --print-excluded
  - tests/behavioral/deterministic/test_run_btests_pytest.py: shell-runner regression tests
  - tests/behavioral/deterministic/deploy-btest-agents.sh: usage docs

  My previous core concerns still stand:

  1. collect-qg4 still downloads qg4-outcomes/ before actions/checkout, so checkout can wipe the artifacts and silently produce pass_list=[].
  2. The workflow still uses inline Python AST parsing even though the PR adds deploy-btest-agents.sh --print-excluded; that is still unnecessary second-path logic.

kami619:
@andrewdonheiser thanks for addressing the comments, I will merge this in.

kami619 (approved):
lgtm

PR #341, opened 2026-09-02
feat(agentic-rag): migrate to MaaS models and integrate ai4rag

Wojciech-Rebisz:
## Description

In this PR:
  - Switch from ogx to MaaS for chat and embeddings
  - Add ai4rag-based document loading with Milvus 
  - Add OpenShell sandbox deployment scripts
  - Update .env.example with MaaS configuration

_Assisted with Claude Code_

## Testing

## Checklist

- [ ] I have read [CONTRIBUTING.md](CONTRIBUTING.md)
- [ ] No `.env` or secret files are included in this PR

## New Agent Proposal

- [ ] Fit check completed ([how to run](CONTRIBUTING.md#proposing-a-new-agent))
- Fit check discussion: 
- Fit score:


agents/langgraph/templates/agentic_rag/pyproject.toml
```diff
@@ -4,25 +4,24 @@ version = "0.2.0"
...
-requires-python = ">=3.12, <3.14"
+requires-python = ">=3.12, <3.15"
 dependencies = [
     "fastapi>=0.135.1",
     "uvicorn[standard]>=0.41.0",
     "langchain>=1.2.10",
     "langchain-openai>=1.1.10",
     "langchain-community>=0.4.1",
-    "langchain-milvus==0.3.3",
+    "langchain-milvus>=0.3.3",
```
Mateusz-Switala: Why do we need langchain-milvus? Don't we use ai4rag to connect with db?

agents/langgraph/templates/agentic_rag/src/agentic_rag/tools.py
```diff
@@ -12,55 +29,108 @@
...
+        embedding_model_id: Embedding model identifier
+        embedding_dimension: Embedding dimension
+        milvus_collection: Milvus collection name
 
     Returns:
-        Dict containing client and vector_store_id
+        ai4rag Retriever instance
     """
-    global _client_cache, _vector_store_id_cache
+    global _retriever_cache
```
Mateusz-Switala: Why do we need even to cache retriever?  Cannot we use closures to initialise retriver in the first call?

Wojciech-Rebisz: Should be better now, can you please check?

andrewdonheiser (changes requested):
## Thermo-Nuclear Code Quality Review — PR #341

**Perform a deep code quality audit of the current branch's changes.**

This PR (OGX→MaaS migration + ai4rag/Milvus retrieval + OpenShell sandbox) is functionally reasonable in places (the `agent.py` rewrite onto `create_agent` genuinely deletes a chunk of hand-rolled StateGraph plumbing — good), but it leaves several structural regressions and duplicated logic behind.

### 1. Security-relevant boundary regression: TLS verification silently downgraded (blocker)

`auth_wrapper.py`:
```python
-_tls_verify: str | bool = _K8S_CA_PATH if Path(_K8S_CA_PATH).is_file() else True
+_tls_verify: str | bool = _K8S_CA_PATH if Path(_K8S_CA_PATH).is_file() else False
```
Previously, when the CA file wasn't found, the client fell back to default trust-store verification (`True`). Now it falls back to `verify=False` — TLS verification off entirely for the K8s TokenReview call — with only a `log.warning`. This is a security-relevant boundary change dressed as a refactor. Can we fail closed (raise) instead of silently disabling TLS verification?

### 2. Structural regression: per-request HTTP client, cache removed with no justification

Same file — the module-level cached `httpx.AsyncClient` (`_get_http_client()`) was deleted in favor of creating a brand-new `AsyncClient` on every single token-review call:
```python
async with httpx.AsyncClient(verify=_tls_verify, timeout=10.0) as client:
```
This runs on every authenticated request when `AUTH_ENABLED=true`. This throws away connection reuse for no stated reason — can we restore a cached client (or explain the tradeoff if intentional)?

### 3. Copy-pasted sqlite3 shim, three divergent copies

The "fix sqlite3 for chromadb" patch appears in `main.py`, `src/agentic_rag/tools.py`, and `data/load_documents_wrapper.py` — and it's not even copied consistently:
- `main.py` / `tools.py`: `sys.modules["sqlite3"] = sys.modules.pop("pysqlite3")`
- `load_documents_wrapper.py`: `sys.modules["sqlite3"] = pysqlite3` (no pop)

This is a bespoke fix duplicated three times instead of one canonical helper. Suggest extracting a single `_patch_sqlite3()` (e.g. in a small `sqlite_shim.py`) and importing it everywhere — the two copies have already drifted, which is exactly the debt this pattern invites.

### 4. Copy-pasted MILVUS_SERVER_CERT normalization

The "is this a file path or PEM text, read + mutate `os.environ`" block is duplicated near-verbatim between `data/load_documents_ai4rag.py` (lines ~70–88) and `tools.py::_initialize_retriever`. Same canonical-utility-vs-bespoke-copy problem as #3 — pull this into a shared helper both the loader and the runtime tool call.

### 5. Feature-specific sandbox logic leaking into the canonical `main.py`

`main.py` grew from 558 → 697 lines almost entirely from OpenShell-sandbox-only concerns bolted directly onto the shared entrypoint: a custom `/docs` Swagger override with inline HTML/CSS, a `/playground` redirect, `/api/health`, and an `/api/chat` proxy that re-issues HTTP requests to `127.0.0.1:8080` with a raw `requests` call. None of this is core chat-completions logic — it's demo-sandbox plumbing. Suggest moving it into its own module/router (e.g. `playground_sandbox.py`, included via `app.include_router(...)` only when `_SANDBOX_MODE`) so `main.py` stays focused on the chat API.

### Lower-priority note (not blocking)

The `retriever_tool` factory (`create_retriever_tool()` + `nonlocal` cache) replacing the old module-level `global` cache is justified by the tests (`test_tools.py` calls `create_retriever_tool()` fresh per test to reset cache state) — not blocking, but it's a sign the caching design itself, not the wrapper, is the awkward part. Worth a second look later.

### Verdict

Requesting changes on items 1 and 2 (genuine regressions — security posture + connection reuse), not stylistic nits. Items 3–4 are mechanical extractions (low risk). Item 5 is a decomposition ask (moving routes to a separate module).

Wojciech-Rebisz:
Hi @andrewdonheiser and thanks for the feedback! I've updated the PR a little bit: TLS verification is now fail-closed by default (raising a RuntimeError unless `K8S_API_INSECURE=true` is explicitly set), and the `httpx.AsyncClient` cache has been restored via `_get_http_client()` for connection reuse across TokenReview calls. Additionally, I cleaned up code duplication for SQLite/Milvus logic and extracted the sandbox routes into a separate router.

andrewdonheiser (changes requested):
## `/api/chat` origin check in `playground_sandbox.py` doesn't actually prevent unauthorized use of `PLAYGROUND_TOKEN`

`playground_chat()` (`playground_sandbox.py:155`) proxies requests to the K8s-token-protected `/chat/completions` endpoint using the server-side `PLAYGROUND_TOKEN`, and the docstring/comments say the origin/referer check exists to prevent unauthorized use of that token (CWE-352).

The check has a bypass built into it: it explicitly allows the request through when both `Origin` and `Referer` are absent (`playground_sandbox.py:184`, the "no origin/referer = same-server call" branch). But that's the default behavior for any non-browser HTTP client — `curl`, `requests`, `httpx`, etc. — none of them send `Origin`/`Referer` unless explicitly told to. So:

```bash
curl -X POST http://<host>/api/chat -d '{"messages":[...]}'
```

sails straight through with no headers at all, using the pod's own `PLAYGROUND_TOKEN` to hit `/chat/completions` — completely bypassing the K8s ServiceAccount TokenReview auth that `auth_wrapper.py` otherwise enforces on that path.

This isn't a meaningful CSRF/origin protection since it can't distinguish "legitimate same-server call" from "attacker who just didn't set the headers." Suggest replacing it with something that can't be forged by an arbitrary client — e.g. a short-lived shared secret/cookie minted when the playground page is served, checked on `/api/chat`, rather than string-matching on client-supplied headers.

andrewdonheiser (changes requested):
.

andrewdonheiser (changes requested):
## Request changes

The new CSRF token check does not prevent unauthorized use of `PLAYGROUND_TOKEN`.

### Why this is still bypassable

At the current PR head:

- `/playground` issues `playground_csrf` to any caller (`playground_sandbox.py:83-95`).
- `/api/chat` accepts any matching cookie/header pair and then forwards the request with the server-side `PLAYGROUND_TOKEN` (`playground_sandbox.py:184-215`).

A direct client can therefore do:

```python
s = requests.Session()
s.get("https://host/playground")
csrf = s.cookies["playground_csrf"]
s.post(
    "https://host/api/chat",
    headers={"X-Playground-CSRF": csrf},
    json={"messages": [{"role": "user", "content": "..."}]},
)
```

`SameSite` cookies and a custom header mitigate cross-origin browser CSRF, but they do not authenticate arbitrary clients that can first fetch the public playground page.

### Required change

Please add real authentication or bind the token to an authenticated session, or remove the server-side-token proxy.

Also add an integration test covering `GET /playground` -> obtain the cookie -> `POST /api/chat`, asserting that an unauthenticated client is rejected. If the intended guarantee is only CSRF protection, update the security comments and documentation so they do not claim to prevent unauthorized `PLAYGROUND_TOKEN` use.

*Formatted follow-up to clarify my previous review; GitHub does not allow editing submitted review bodies.*

PR #328, opened 2026-08-25
ci: add quality gates orchestrator workflow (RHAIENG-6542)

mpk-droid:
## Description

Adds the quality gates orchestrator workflow (`quality-gates-pipeline.yml`) that chains QG4 and QG7 using the staged fan-out pattern with dynamic matrix, as designed in RFC Discussion #285.

Each gate handles its own setup and teardown independently. The orchestrator provides:

- **Staged fan-out**: QG4 runs all 11 agents in parallel (`fail-fast: false`), then a `collect-qg4` job gathers per-agent outcomes via uploaded artifacts and builds a JSON pass-list
- **Dynamic matrix**: QG7 uses `fromJson()` to run only for agents that passed QG4
- **Scoped secrets**: QG4 and QG7 each receive only the secrets they need. QG7 receives cluster credentials plus agent deployment credentials (`API_KEY`, `OPENAI_API_KEY`, `POSTGRES_PASSWORD`, etc.) because it deploys and tears down its own agents via the `run-qg7` action
- **Per-agent visibility**: Both QG4 and QG7 run as matrix jobs, giving clear per-agent pass/fail in the GitHub Actions UI

**Jobs:**

| Job | Purpose |
|---|---|
| `verify-cluster-connection` | Pre-flight cluster smoke test (extracted to composite action) |
| `qg4` | Matrix: deployment health checks (11 agents) |
| `collect-qg4` | Build JSON pass-list from QG4 outcome artifacts |
| `qg7` | Dynamic matrix: behavioral evals (QG4-passing agents only, with uv cache) |
| `notify-slack` | Standard Slack notification |

**Also in this PR:**
- `agent-deployment-test.yaml`: schedule trigger removed (orchestrator owns nightly), uses new `verify-cluster-connection` action
- `.github/actions/verify-cluster-connection/action.yml`: new composite action extracted from inline steps

Depends on: RHAIENG-6540 (#317), RHAIENG-6541 (#321)
Compatible with: #325 (QG7 deploy lifecycle — env vars synced)
Ref: RFC Discussion #285

## Jira Ticket

RHAIENG-6542

## Testing

- [ ] `make test` passes (run from the affected agent directory)
- [x] Manual testing performed (describe steps below)
- [ ] No testing required (documentation/config change only)

**Structural verification performed locally:**
- YAML parses successfully (pyyaml `safe_load`)
- Job dependency graph verified: `verify-cluster-connection` → `qg4` → `collect-qg4` → `qg7` → `notify-slack`
[...]

PR #342, opened 2026-09-03
fix(ci): track Quality Gates Pipeline on the CI health dashboard

mpk-droid:
## Summary
- The QG4 orchestrator was renamed to `Quality Gates Pipeline` in #328 and now runs nightly via cron; the old standalone `agent-deployment-test.yaml` lost its `schedule` trigger and became `workflow_dispatch`-only.
- The CI health dashboard (`ci-health-pages.yml` / `generate_ci_health_page.py`) only listened for the old standalone workflow's name, so the QG4 section silently stopped updating after the orchestrator was introduced.
- Now the dashboard tracks both: the nightly `Quality Gates Pipeline` run, and the manual-only `QG4: Agent Deployment Integration Tests` dispatch (kept around for one-off QG4 checks without triggering QG7).
- Added a regression test asserting `WORKFLOWS` in `generate_ci_health_page.py` stays in sync with the `workflow_run` trigger list in `ci-health-pages.yml`, to catch this exact class of silent-drift bug in the future.
- Updated `docs/ci-health-dashboard.md` to reflect all 7 tracked workflows and clarify that `agent-deployment-test.yaml`'s card showing "No completed runs" between manual dispatches is expected.

## Test plan
- [x] `pytest .github/scripts/tests/test_generate_ci_health_page.py` — 13 passed
- [x] Regenerated the dashboard from the fixture locally and confirmed both QG4 cards render distinctly
- [x] `ruff check` / `ruff format --check` on changed files

## Follow-up after merge
- Confirm a `Quality Gates Pipeline` or `QG4: Agent Deployment Integration Tests` run triggers a Pages rebuild in production.

Assisted by Claude Sonnet 5

### DIFF
--- a/.github/scripts/fixtures/ci-runs-sample.json
+++ b/.github/scripts/fixtures/ci-runs-sample.json
@@ -82,10 +82,10 @@
       "run_started_at": "2026-07-05T01:15:18Z"
     }
   ],
-  "agent-deployment-test.yaml": [
+  "quality-gates-pipeline.yml": [
     {
       "id": 10031,
-      "name": "QG4: Agent Deployment Integration Tests",
+      "name": "Quality Gates Pipeline",
       "event": "schedule",
       "head_branch": "main",
       "status": "completed",
@@ -97,7 +97,7 @@
     },
     {
       "id": 10030,
-      "name": "QG4: Agent Deployment Integration Tests",
+      "name": "Quality Gates Pipeline",
       "event": "schedule",
       "head_branch": "main",
       "status": "completed",
@@ -106,6 +106,20 @@
       "created_at": "2026-07-04T03:04:44Z",
       "updated_at": "2026-07-04T03:58:12Z",
       "run_started_at": "2026-07-04T03:04:52Z"
+    }
+  ],
+  "agent-deployment-test.yaml": [
+    {
+      "id": 10032,
+      "name": "QG4: Agent Deployment Integration Tests",
+      "event": "workflow_dispatch",
+      "head_branch": "main",
+      "status": "completed",
+      "conclusion": "success",
+      "html_url": "https://github.com/red-hat-data-services/agentic-starter-kits/actions/runs/10032",
+      "created_at": "2026-07-05T09:12:00Z",
+      "updated_at": "2026-07-05T10:01:33Z",
+      "run_started_at": "2026-07-05T09:12:08Z"
     }
   ]
 }
--- a/.github/scripts/generate_ci_health_page.py
+++ b/.github/scripts/generate_ci_health_page.py
@@ -53,9 +53,14 @@ WORKFLOWS = (
         "description": "Operator, DataScienceCluster, and KServe platform readiness checks.",
     },
     {
+        "file": "quality-gates-pipeline.yml",
+        "name": "Quality Gates Pipeline",
+        "description": "Nightly QG4 deploy/health checks fanning out into QG7 behavioral evals.",
+    },
+    {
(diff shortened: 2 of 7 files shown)


kami619 (dismissed):
Requesting changes. The catalog + `workflow_run` wiring is the right local fix, but this does not yet close the silent-staleness class of bug, and Slack/QG4 inventory was updated in the wrong docs layer.

**1. Regression test is necessary, not sufficient.** GitHub rebuilds by workflow `name:`; the dashboard API lists runs by **file**. Those are two contracts. `test_workflow_names_match_ci_health_pages_trigger_list` only checks catalog names vs `ci-health-pages.yml`. If `quality-gates-pipeline.yml` is renamed again and both lists are left alone (or both are updated to the wrong name), the test stays green and Pages goes stale the same way. Parse each `WORKFLOWS[].file` and require `YAML name: == WORKFLOWS[].name == trigger list`. Stronger: drop `name` from Python and read it from the workflow file. Also don’t rely only on `workflow_doc[True]` for `on:` (YAML 1.1 boolean trap).

**2. Alert inventory belongs in policy, not only the dashboard page.** This PR claims seven Slack senders including Quality Gates Pipeline, while still linking `ci-alert-policy.md` as source of truth. Policy/runbook still map canonical QG4 to `agent-deployment-test.yaml` with `schedule`. Update policy + runbook here, or drop the seven-workflow Slack claim.

**3. Two peer QG4 cards can mislead.** A stale-green manual dispatch next to a red nightly pipeline is worse than an empty card. If both stay, label the dispatch-only card as ad hoc, not nightly health. Align the “No completed runs” docs with the actual empty-state copy.

The one-dict + one-trigger-line change is the existing architecture and is fine for today’s strings. Ship when (1) and (2) are addressed.

---

[AIA PAI Nc Hin R Cursor Grok 4.6 v1.0](https://aiattribution.github.io/)

mpk-droid:
@kami619 thanks for the review — addressed all three in 476c090:

1. **Sync test insufficient**: `WORKFLOWS` no longer hardcodes display names — each one is now read directly from its own workflow file's `name:` field at import time, so there's a single source of truth instead of two hand-typed lists that could agree with each other while both being wrong. The test independently re-reads every workflow file and asserts it matches the catalog and the `ci-health-pages.yml` trigger list. Also dropped the `workflow_doc[True]` reliance — added a `GHWorkflowLoader` that disables implicit bool resolution so `on:` parses as the string `"on"`, not the YAML 1.1 boolean trap.
2. **Alert inventory**: added `Quality Gates Pipeline` as a canonical `QG4`/`QG7` alert source to `ci-alert-policy.md` and `ci-alert-runbook.md`, and corrected `agent-deployment-test.yaml`'s row there to `workflow_dispatch`-only (it lost its `schedule` trigger).
3. **Misleading peer cards**: labeled the retained `agent-deployment-test.yaml` card as "Ad hoc, manual-only QG4 deploy/health check — not the nightly QG4 signal" in its description, and updated the dashboard doc to quote the actual empty-state copy instead of a paraphrase.

Also caught along the way: the fix for (1) added a `pyyaml` import to the generator script, but the production Pages workflow never installs dependencies (only the test job does via `uv sync --extra test`). Added a `pip install pyyaml` step to `ci-health-pages.yml`, mirroring the existing pattern in `pre-release.yaml`, and verified the script runs end-to-end with only that installed.

128/128 tests pass, ruff and markdownlint clean.

Assisted by Claude Sonnet 5

kami619 (approved):
thanks for addressing all the changes

PR #346, opened 2026-09-04
fix(ci): correct Langflow namespace in QG7 btest runner

andrewdonheiser:
## Summary

- The QG7 btest runner had `langflow-agent` hardcoded as the namespace override for the Langflow agent tuple, but the Route lives in `ci-testing` (the cluster's shared test namespace)
- This caused `oc get route langflow -n langflow-agent` to fail with `RouteNotFoundError`, blocking the langflow behavioral tests entirely
- One-character fix: change namespace override from `langflow-agent` → `ci-testing` in `run-btests-pytest.sh`

## Test plan

- [ ] Verify QG7 run no longer errors with `Route 'langflow' not found for flow-import agent` in the langflow job
- [ ] Confirm `oc get route langflow -n ci-testing` resolves on the cluster

### DIFF
--- a/tests/behavioral/deterministic/run-btests-pytest.sh
+++ b/tests/behavioral/deterministic/run-btests-pytest.sh
@@ -49,7 +49,7 @@ AGENTS=(
   "vanilla_python/templates/openai_responses_agent|VANILLA_PYTHON_AGENT_URL|openai-responses-agent"
   "langgraph/templates/human_in_the_loop|HITL_AGENT_URL|langgraph-hitl-agent"
   "google/templates/adk|GOOGLE_ADK_AGENT_URL|google-adk-agent"
-  "langflow/templates/simple_tool_calling_agent|LANGFLOW_TOOL_CALLING_AGENT_URL|langflow|langflow-agent"
+  "langflow/templates/simple_tool_calling_agent|LANGFLOW_TOOL_CALLING_AGENT_URL|langflow|ci-testing"
   "a2a/templates/langgraph_crewai_agent|A2A_LANGGRAPH_CREWAI_AGENT_URL|a2a-langgraph-agent"
   "langgraph/examples/guardrailed_agent|GUARDRAILED_AGENT_URL|langgraph-guardrailed-agent"
 )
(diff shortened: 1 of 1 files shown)


mpk-droid:
Heads up — `tests/behavioral/deterministic/test_deploy_agents.py:163` (`test_langflow_uses_its_own_namespace`) still expects `namespace == "langflow-agent"` and `deployment_name == "langflow"`, but with this PR's tuple change those come out as `"ci-testing"` and `"langflow-agent"` now. Ran the suite and this is the only failing test (58 passed / 1 failed). Could we update the assertions there to match the new values so the suite stays green?

Assisted by Claude Sonnet 5

mpk-droid:
Just flagging so it doesn't slip through: the CI checks on this PR are green, but neither actually exercised this change — "Unit Tests" only runs for `agents/**` path changes (this PR only touches `tests/`, so it's a no-op here), and "Cluster Behavioral Tests" only does a real run on manual dispatch. So the failing test in the other comment wouldn't have been caught automatically. Might be worth a follow-up to get `tests/behavioral/deterministic/` covered by an automated gate at some point.

Assisted by Claude Sonnet 5

mpk-droid (changes requested):
added few comments

tests/behavioral/deterministic/run-btests-pytest.sh
```diff
@@ -49,7 +49,7 @@ AGENTS=(
   "vanilla_python/templates/openai_responses_agent|VANILLA_PYTHON_AGENT_URL|openai-responses-agent"
   "langgraph/templates/human_in_the_loop|HITL_AGENT_URL|langgraph-hitl-agent"
   "google/templates/adk|GOOGLE_ADK_AGENT_URL|google-adk-agent"
-  "langflow/templates/simple_tool_calling_agent|LANGFLOW_TOOL_CALLING_AGENT_URL|langflow|langflow-agent"
+  "langflow/templates/simple_tool_calling_agent|LANGFLOW_TOOL_CALLING_AGENT_URL|langflow-agent"
```
mpk-droid, line 52: Since this branch has landed on a new approach for the Langflow route/namespace, could we get a quick verification before merging — e.g. `oc get route -n ci-testing | grep langflow` output, or a link to a QG7 run that passes with this change? Since Langflow doesn't have an in-repo Helm chart, that's the only way to confirm the value is right.

Also, would be great to check off the test plan boxes once that's confirmed.

Assisted by Claude Sonnet 5

PR #356, opened 2026-09-08
Add Trustabl Agent Scanner to CI

trustabl-kathrina:
We came across your repo and we like that it provides production-ready starter kits for building and deploying AI agents on Red Hat OpenShift. We scanned the repo, and noticed agent runtime reliability findings that might be worth reviewing.

1. [MEDIUM] Agent has no safety_settings
   File: agents/google/templates/adk/src/adk_agent/agent.py
   What it means: Google ADK agents use Gemini models whose configurable content filters default to OFF when safety_settings is not provided.

Recommendations are based on our understanding of agent runtime reliability, some findings may be intentional. Please let us know if this was intentional or if our findings are helpful so we can improve the accuracy of the scanner.

Best,
Trustabl.ai
Open-source AI agent reliability scanner (runs locally, GitHub Action)

### DIFF
--- a/.github/workflows/trustabl.yml
+++ b/.github/workflows/trustabl.yml
@@ -1,0 +1,33 @@
+name: Trustabl Agent Scanner
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+  workflow_dispatch:
+
+# Minimal top-level permissions; write grants are scoped to the scan job only.
+permissions:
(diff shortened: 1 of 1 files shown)


kami619:
@mpk-droid and @sanafayyaz315 wdyt of this scanner's discoveries ?

PR #359, opened 2026-09-08
fix(a2a): correlate MLflow trace lookup to the right process (RHAIENG-7403)

andrewdonheiser:
## Summary

**The bug:** every tool call the eval harness observed came back as `web_search`, regardless of which tool the agent actually called. `test_pass_at_k_tool_usage` and `test_tool_selection_accuracy` failed 100% of the time as a result.

**Root cause:** the LangGraph orchestrator and CrewAI specialist are separate processes that both fall back to the same `MLFLOW_EXPERIMENT_NAME`. The eval harness's `get_latest_trace()` picks "most recent trace in the experiment" — since the CrewAI specialist's trace always starts *after* the orchestrator's (it's a downstream call triggered from inside the orchestrator's tool span), the harness always fetched the CrewAI trace instead of the orchestrator's. The CrewAI trace's only tool is `"Web Search"` → normalized to `web_search` by the scorer. The orchestrator's real `ask_crew_specialist` call was never observed at all.

**Before vs. after, verified live on the `agen-e2e-rhoai2` cluster (same agent, same deployment, same test suite):**

| | expected | actual (harness-reported) | result |
|---|---|---|---|
| Before fix | `ask_crew_specialist` | `web_search` | `test_pass_at_k_tool_usage` + 3x `test_tool_selection_accuracy` FAIL — pass@8 = 0/8 |
| After fix | `ask_crew_specialist` | `ask_crew_specialist` | 19/19 tests PASS — "No tool name mismatches" |

The model was calling the correct tool the whole time; the fix corrects what the harness *observes*, so `expected == actual` now.

## Changes
- `tests/behavioral/conftest.py` — prefer `MLFLOW_EXPERIMENT_NAME_LANGGRAPH` (falls back to `MLFLOW_EXPERIMENT_NAME`) so trace lookups target only the orchestrator's experiment.
- `tests/behavioral/deterministic/deploy_btest_agents.py` — for this agent, set distinct `MLFLOW_EXPERIMENT_NAME_LANGGRAPH` / `MLFLOW_EXPERIMENT_NAME_CREWAI` values (one per deployment) so the two processes stop sharing one experiment.
- `Makefile` — forward the two new env vars to both Helm deploy phases.
- `tests/behavioral/fixtures/golden_queries.yaml` — reverted `expected_tools` back to `ask_crew_specialist`. An earlier commit on this branch had instead changed the fixture to match the wrong observed value (`web_search`), masking the bug instead of fixing it.

Jira: RHAIENG-7403

## Test plan
- [x] Deployed `a2a/templates/langgraph_crewai_agent` to a live OpenShift cluster (`adonheis-testing` namespace) via the QG7 deploy tooling
- [x] Ran the full behavioral pytest suite against the live deployment before the fix — confirmed 4 failures, all showing `web_search` instead of `ask_crew_specialist`
- [x] Applied the fix, redeployed, re-ran the suite — 19/19 passed, `expected == actual` (`ask_crew_specialist`) on every case
- [x] Torn down the test deployment

### DIFF
--- a/agents/a2a/templates/langgraph_crewai_agent/Makefile
+++ b/agents/a2a/templates/langgraph_crewai_agent/Makefile
@@ -168,6 +168,8 @@ deploy: _check-env ## Two-phase Helm deploy (Routes first, then Deployments with
 	  --set-string env.MODEL_ID="$$MODEL_ID" \
 	  $${MLFLOW_TRACKING_URI:+--set-string env.MLFLOW_TRACKING_URI="$${MLFLOW_TRACKING_URI}"} \
 	  $${MLFLOW_EXPERIMENT_NAME:+--set-string env.MLFLOW_EXPERIMENT_NAME="$${MLFLOW_EXPERIMENT_NAME}"} \
+	  $${MLFLOW_EXPERIMENT_NAME_LANGGRAPH:+--set-string env.MLFLOW_EXPERIMENT_NAME_LANGGRAPH="$${MLFLOW_EXPERIMENT_NAME_LANGGRAPH}"} \
+	  $${MLFLOW_EXPERIMENT_NAME_CREWAI:+--set-string env.MLFLOW_EXPERIMENT_NAME_CREWAI="$${MLFLOW_EXPERIMENT_NAME_CREWAI}"} \
 	  $${MLFLOW_TRACKING_INSECURE_TLS:+--set-string env.MLFLOW_TRACKING_INSECURE_TLS="$${MLFLOW_TRACKING_INSECURE_TLS}"} \
 	  $${MLFLOW_WORKSPACE:+--set-string env.MLFLOW_WORKSPACE="$${MLFLOW_WORKSPACE}"} && \
 	echo "=== Waiting for OpenShift Route hostnames ===" && \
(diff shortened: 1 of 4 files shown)


kami619 (commented):
Thanks for picking up [RHAIENG-7403](https://redhat.atlassian.net/browse/RHAIENG-7403). I reviewed the change against the ticket and the agent architecture — sharing thoughts below with a soft pushback, not a hard block.

## Summary

PR #359 updates four `expected_tools` entries in `golden_queries.yaml` from `ask_crew_specialist` to `web_search`. That may unblock some tool-selection checks if MLflow is currently reporting the CrewAI specialist's `Web Search` span (normalized to `web_search`), but it looks like a **symptom-level fixture tweak** rather than a fix aligned with the ticket's original intent or the underlying trace mismatch.

## What RHAIENG-7403 originally intended

The Jira history (before the description was shortened) called for **different expectations by query type**, not a blanket rename:

| Query type | Expected tool |
|-----------|----------------|
| Easy specialist delegation | `ask_crew_specialist` |
| Medium web-search query | `web_search` |
| Hard / adversarial specialist queries | `ask_crew_specialist` |

The current PR changes **all four** specialist queries to `web_search`, which drifts from that design.

## Architecture context (why this matters)

This agent is two-layer A2A:

1. **LangGraph orchestrator** — exposes `ask_crew_specialist` (`langgraph_a2a_server.py`)
2. **CrewAI specialist** — exposes `DummyWebSearchTool` with display name **Web Search** (normalizes to `web_search` in the scorer)

So the PR description framing ("the agent's actual tool is `web_search`") is a bit misleading. The orchestrator tool is still `ask_crew_specialist`; `web_search` is the **inner CrewAI tool** when the specialist runs web search.

Behavioral evals get `tool_calls` from **MLflow trace enrichment**, not the HTTP response. The harness takes the latest trace and collects all `TOOL` spans. With both LangGraph and CrewAI tracing into MLflow, enrichment may be picking up the **CrewAI trace** (`Web Search` → `web_search`) instead of the LangGraph trace (`ask_crew_specialist`) — which would explain golden-query failures even when delegation worked.

Also, `score_tool_selection` requires an **exact set match** (`expected_set == actual_set`). If traces contain **both** tools, neither single-tool expectation passes.

The fixture header still documents the opposite behavior:

> MLflow enrichment picks up the LangGraph trace and reports `ask_crew_specialist` as the tool name.

PR #359 changes expectations but doesn't update that comment, so docs and tests now disagree.

## Gaps I noticed

- **`test_reliability.py`** still hardcodes `expected_tools = ["ask_crew_specialist"]` — pass@k can fail under the same trace behavior this PR is working around.
- **Stale docs** — golden fixture header, `test_tool_usage.py`, README still describe `ask_crew_specialist` as the traced tool.
- **Over-broad change** — easy query ("explain agent-to-agent protocols") may not invoke CrewAI Web Search at all; expecting `web_search` everywhere feels brittle.
- **Root cause untouched** — no change to trace selection (LangGraph vs CrewAI) or scorer handling of nested tool calls.

## Soft pushback / ask

Could we stick closer to the **original issue intent** and address the **root cause** rather than renaming one tool across all delegation queries?

Some paths that seem more durable:

1. **Fix trace enrichment** to prefer LangGraph orchestrator tool spans when scoring orchestrator delegation → keep `ask_crew_specialist` where appropriate.
2. **Per-query expectations** per the original Jira design (`ask_crew_specialist` for easy/hard/adversarial; `web_search` for the medium query).
3. If the team has decided to score the **inner CrewAI tool** instead, that's fine — but then please update **all** related tests/docs (`test_reliability.py`, fixture header, README) in the same change set, not just golden queries.

**Clarification request:** Is this PR intended as a quick unblock with follow-up fixes planned (trace selection, reliability test, docs)? Or is the one-file swap the complete scope for RHAIENG-7403? Happy to re-review quickly once that's clear.

---

[AIA PAI Ce Hin R Composer2.5 v1.0](https://aiattribution.github.io/statements/AIA-PAI-Ce-Hin-R-?model=Composer2.5)

kami619 (approved):
LGTM

PR #340, opened 2026-09-01
ci(orchestrator): integrate QG1 and QG2 into quality gates pipeline (RHAIENG-7210)

mpk-droid:
## Description

This PR integrates QG1 (cluster readiness) and QG2 (platform readiness) as upstream stages in the quality gates orchestrator pipeline, running before QG4 and QG7.

### Changes

- **Add QG1 job (Stage 0a)**: Cluster readiness checks using `.github/actions/run-qg1`
  - Assumes dedicated QG1 service account
  - Configurable via `QG1_CLUSTER_PROFILE`, `QG1_REQUIRE_GPU`, `QG1_REQUIRED_NAMESPACES`
  - Uploads results as artifacts

- **Add QG2 job (Stage 0b)**: Platform readiness checks using `.github/actions/run-qg2`
  - Runs sequentially after QG1 (`needs: qg1`) and only if QG1 succeeds
    (`if: needs.qg1.result == 'success'`)
  - Assumes dedicated QG2 service account
  - Configurable via `QG2_CLUSTER_PROFILE`, `QG2_CLUSTER_TYPE`, `QG2_REQUIRE_DSC_READY`, `QG2_REQUIRE_KSERVE`
  - Uploads results as artifacts

- **Update verify-cluster-connection**: Now depends on `[qg1, qg2]` and requires both to succeed
- **Update notify-slack**: Added `qg1` and `qg2` to needs array for status reporting
- **Update flow diagram**: Reflects new pipeline chain in header comments
- **Add orchestrator contract tests**: `.github/scripts/tests/test_orchestrator_contract.py`
  asserts the job dependency chain and failure-handling conditionals (QG2 gating on QG1
  success, verify-cluster-connection requiring both QG1 and QG2, notify-slack depending
  on all upstream gates)
- **Add orchestrator DAG simulation tests**: `.github/scripts/tests/test_orchestrator_dag_simulation.py`
  replays the actual `needs`/`if` graph against forced job outcomes (QG1 fails, QG2 fails,
  QG4 fails, no agents pass QG4) and asserts the resulting success/failure/skipped status
  of every job — verifying real failure propagation, not just condition-string presence

### Failure Handling

- **QG1 fails**: QG2 is skipped (does not run), all downstream gates blocked
- **QG2 fails**: QG4 and QG7 blocked, but QG1 results still reported via notify-slack
- **Both pass**: Full pipeline executes as before

QG1 and QG2 run strictly sequentially and QG1 must succeed for QG2 to run — see the
[...]

PR #360, opened 2026-09-09
feat(ci): show gate/agent status summary in QG pipeline Slack alert (RHAIENG-6543)

mpk-droid:
## Description

Redesigns the Quality Gates Pipeline's final Slack alert (`notify-slack` job) to show a QG1/QG2/QG4/QG7 gate-status summary plus a per-agent QG4/QG7 pass-fail matrix (✅/❌/⏭️/⏱️), instead of the generic "list of failed job names" pulled from the GitHub Jobs API. The single-message approach was a deliberate scope decision made with the ticket owner during implementation: rather than adding separate per-stage "immediate" alerts (as originally scoped), we consolidated on one comprehensive final message that always fires regardless of which gate a run reaches, to avoid duplicate/noisy alerts for the same failure.

### Changes

- `qg7` job now records a per-agent outcome artifact (`qg7-outcome-<agent>`), mirroring the existing `qg4` pattern, so the summary can show QG7 status per agent.
- New `.github/scripts/build_qg_summary.sh` builds the gate-status block and (only when QG4 actually ran) the per-agent matrix from the QG1/QG2/QG4/QG7 job results and the outcome artifacts. Validates each outcome artifact's JSON before use and degrades to a placeholder/warning instead of crashing on malformed data (see "Fixes from review" below).
- `notify-slack` job downloads both outcome artifact sets and calls the new script before deciding whether to notify. The existing `should_notify.sh` gating logic is unchanged — it already correctly aggregates matrix-level failures across QG1/QG2/QG4/QG7.
- Added an optional `QG_NOTIFY_ON_SUCCESS` repo variable: when `'true'`, an all-green run also posts a (green) summary — satisfying the "or configurable" part of the AC, default is silent-on-success like today. Folded directly into the "Send Slack notification" step's `if:` rather than a separate step.
- The QG7-exclusion list (agents that pass QG4 but skip QG7) is now a single `QG7_EXCLUDED_AGENTS_JSON` workflow-level `env:` value, read by both the `collect-qg4` job and `build_qg_summary.sh`, instead of two hardcoded copies kept in sync by comment.
- `.github/actions/notify-slack` gains an optional `summary_text` input: when supplied, it replaces the generic failed-jobs section (and skips that GitHub API call). Every other caller of this action (`agent-tests.yml`, `code-quality.yml`, `eval-gating.yml`, the standalone `qg1`/`qg2` workflows) doesn't pass it, so their behavior is unchanged.
- Fixed two now-stale lines in `docs/ci-alert-policy.md` / `docs/ci-alert-runbook.md` (severity fingerprint for `Quality Gates Pipeline` was `QG4`/`QG7`, now `QG1`/`QG2`/`QG4`/`QG7`; noted the QG-pipeline-specific message format).

### Note on branch history

This branch is built on top of #340 (RHAIENG-7210, not yet merged), which integrates QG1/QG2 into this same workflow — this ticket's gate-summary design assumes QG1/QG2 exist in the pipeline. #340's three commits are cherry-picked in here cleanly (no conflicts). **This PR should be reviewed/merged after #340**, or rebased to drop the now-duplicate commits once #340 lands.

### Fixes from review

A first review pass flagged several issues, addressed as follows:

- **Critical — a malformed/truncated outcome artifact could silence the entire alert:** `build_qg_summary.sh` now validates each outcome file's JSON before use, skips unreadable ones with a visible `⚠️` note instead of crashing, and the risky table-building logic is isolated so an unexpected error degrades to a placeholder rather than aborting the script. At the workflow level, the `Build QG summary` step got `continue-on-error: true` + `if: always()`, and `Decide notification status` got `if: always()`, so a summary-generation failure can no longer prevent the Slack notification step from running.
- **Redundant boolean re-derivation:** removed the separate "Apply success-notification override" step; `QG_NOTIFY_ON_SUCCESS` is now evaluated directly in the `if:` of "Send Slack notification".
- **Duplicated QG7-exclusion list:** moved to a single `QG7_EXCLUDED_AGENTS_JSON` workflow-level env var (see Changes above).
- **Cancelled/timed-out agents rendered identically to routine skips:** added distinct `⏱️` handling in both `gate_status_icon()` and `agent_status_icon()`.
- **Stale jq arg name in `render_payload.sh`:** renamed `failed_jobs_text` → `details_text` to match its broader purpose post-refactor.
- **Skipped, with reasoning:** (1) reconciling against a hardcoded "expected agent list" to flag artifact-download failures — would add a *third* place with the agent list to keep in sync, working against the exclusion-list dedup above, for a rare edge case already visible in job logs; (2) merging the QG4/QG7 outcome artifacts into one blob to avoid `collect-qg4`/`notify-slack` re-downloading the same per-agent artifacts — real but minor CI-minutes optimization, not correctness; (3) an end-to-end test of the composite action's `if: inputs.summary_text == ''` skip branch — noted as a known coverage gap, hard to unit test without an actions-runner harness.
- **Timing claim caveat:** the "per-stage failures already surface reasonably fast today" reasoning behind dropping the separate immediate-alert job is based on tracing the workflow's `needs`/`if` DAG (confirmed by `test_orchestrator_dag_simulation.py`'s skip-propagation model), not a live-run wall-clock measurement — treat it as "logically unblocked quickly," not a timed guarantee.

### Fixes from re-review

A follow-up review pass caught one issue my first fix didn't fully close, plus a coverage gap:

- **Row cap didn't actually bound Slack's 3000-char limit:** a fixed `MAX_AGENT_ROWS=40` doesn't account for agent name length — verified that 40 rows of realistic-length names (e.g. `vanilla-python-openai-responses-agent`, the longest in the repo today) already exceeds 3000 characters. Replaced the row-count cap with a character-budget-based truncation: the script tracks the actual rendered length against a `TOTAL_CHAR_BUDGET` (default 2900, leaving headroom below Slack's limit), reserving room for the gate summary, any warning lines, and the truncation trailer, and stops adding agent rows once the budget is spent. Added a test with 100 long-named synthetic agents asserting the output stays ≤3000 characters.
- **No test coverage for the `if:` condition on "Send Slack notification":** added direct tests in `test_orchestrator_contract.py` that parse the actual `if:` expression string from the workflow file and evaluate it against the should_notify/status/`QG_NOTIFY_ON_SUCCESS` combinations, so a future edit to that boolean logic fails a test instead of only being caught by review.

## Jira Ticket

RHAIENG-6543

## Testing

- [x] `make test` passes (run from the affected agent directory) — N/A, no agent code changed; ran the relevant script test suite instead

Manual/automated verification performed:
- `actionlint` clean across the entire repo (`.github/workflows/*.yml`, all composite actions)
- `shellcheck` clean on all touched/new bash scripts
- `uv run pytest .github/scripts/tests/` — 161/161 passing, including 8 tests in `test_build_qg_summary.py` (malformed-JSON resilience, char-budget truncation at scale with long names, cancelled-status rendering, plus the original 3) and 5 new `if:`-expression tests in `test_orchestrator_contract.py`
- Manually ran `build_qg_summary.sh` against fixture scenarios: all-pass, QG1-fails-nothing-downstream-runs, partial QG4+QG7 failure, a truncated/malformed outcome file (confirmed graceful degradation, exit 0), and 100 synthetic agents with realistic long names (confirmed output stays under 3000 characters — verified via bash's `${#var}`/Python's `len()`, both count Unicode codepoints; `wc -c` counts UTF-8 bytes and isn't the right comparison here)
- `ruff check` / `ruff format --check` clean on touched Python test files

## Checklist

Scope note: the ticket's "per-stage notify-slack job fires immediately when any agent fails a gate" AC was intentionally dropped in favor of the single comprehensive final alert described above — this was discussed and agreed with the ticket owner during implementation to avoid duplicate alerts for the same underlying failure.

## Review Guidance

**Key areas for review:**
- `build_qg_summary.sh` — the jq logic mapping each agent's QG4 status + QG7-exclusion + QG7 outcome to a single display status (`blocked_fail` / `blocked_skip` / `excluded` / `not_run` / pass-through), and the malformed-JSON / row-cap resilience logic
- The `notify-slack` job's new step ordering in `quality-gates-pipeline.yml` (artifact downloads → build summary → decide notification → send), and the `continue-on-error`/`always()` resilience chain
- `QG_NOTIFY_ON_SUCCESS` — confirm the default (silent on success) matches expectations

**Safe to skim:**
- The three `RHAIENG-7210` commits (QG1/QG2 integration) — already reviewed on #340
- Doc fixes in `ci-alert-policy.md` / `ci-alert-runbook.md`

## Related PRs

Depends on / includes: #340 (RHAIENG-7210)

---

### DIFF
--- a/.github/actions/notify-slack/action.yml
+++ b/.github/actions/notify-slack/action.yml
@@ -29,6 +29,10 @@ inputs:
   status:
     description: "Notification status label, for example failure or cancelled"
     required: true
+  summary_text:
+    description: "Pre-rendered mrkdwn summary to show instead of the generic failed-jobs list"
+    required: false
+    default: ""
   preview:
     description: "If true, print the payload instead of sending it"
     required: false
@@ -39,6 +43,7 @@ runs:
   steps:
     - name: Collect failed jobs
       id: collect-failed-jobs
+      if: inputs.summary_text == ''
       shell: bash
       env:
         GITHUB_TOKEN: ${{ inputs.github_token }}
@@ -83,6 +88,7 @@ runs:
         DASHBOARD_URL: ${{ inputs.dashboard_url }}
         REPOSITORY: ${{ inputs.repository }}
         FAILED_JOBS_JSON: ${{ steps.collect-failed-jobs.outputs.failed_jobs_json }}
+        SUMMARY_TEXT: ${{ inputs.summary_text }}
       run: |
         set -euo pipefail
 
--- a/.github/actions/notify-slack/render_payload.sh
+++ b/.github/actions/notify-slack/render_payload.sh
@@ -78,7 +78,7 @@ jq -n \
   --arg ref_name "${REF_NAME}" \
   --arg repository "${REPOSITORY}" \
   --arg timestamp "${TIMESTAMP}" \
-  --arg failed_jobs_text "${DETAILS_TEXT}" \
+  --arg details_text "${DETAILS_TEXT}" \
   --arg links_text "${LINKS_TEXT}" \
   '{
     text: $fallback_text,
@@ -128,7 +128,7 @@ jq -n \
             type: "section",
             text: {
               type: "mrkdwn",
-              text: $failed_jobs_text
+              text: $details_text
             }
           },
           {
--- a/.github/actions/notify-slack/render_payload.sh
+++ b/.github/actions/notify-slack/render_payload.sh
@@ -17,12 +17,18 @@ REPOSITORY="${REPOSITORY:?REPOSITORY is required}"
 STATUS="${STATUS:-failure}"
 TIMESTAMP="${TIMESTAMP:-$(date -u +%Y-%m-%dT%H:%M:%SZ)}"
 FAILED_JOBS_JSON="${FAILED_JOBS_JSON:-[]}"
+SUMMARY_TEXT="${SUMMARY_TEXT:-}"
 
 if ! jq -e 'type == "array"' <<<"${FAILED_JOBS_JSON}" >/dev/null 2>&1; then
   FAILED_JOBS_JSON='[]'
 fi
 
 case "${STATUS}" in
+  success)
+    COLOR="#2eb67d"
+    TITLE="CI Success: ${WORKFLOW_NAME}"
+    STATUS_LABEL="Passed"
+    ;;
   failure | failed | timed_out)
     COLOR="#d00000"
     TITLE="CI Failure: ${WORKFLOW_NAME}"
@@ -40,19 +46,25 @@ case "${STATUS}" in
     ;;
 esac
 
-FAILED_JOBS_TEXT="$(
-  jq -nr --argjson jobs "${FAILED_JOBS_JSON}" '
-    if ($jobs | length) == 0 then
-      "*Failed jobs*\n- Job details unavailable"
-    elif ($jobs | length) <= 10 then
-      "*Failed jobs*\n" + ($jobs | map("- `" + . + "`") | join("\n"))
-    else
-      "*Failed jobs*\n"
-      + ($jobs[:10] | map("- `" + . + "`") | join("\n"))
-      + "\n- ... and \((($jobs | length) - 10)) more"
-    end
-  '
-)"
+if [[ -n "${SUMMARY_TEXT}" ]]; then
+  # Caller supplied its own rendered summary (e.g. a gate/agent status
+  # breakdown) — use it verbatim instead of the generic failed-jobs list.
+  DETAILS_TEXT="${SUMMARY_TEXT}"
+else
+  DETAILS_TEXT="$(
+    jq -nr --argjson jobs "${FAILED_JOBS_JSON}" '
+      if ($jobs | length) == 0 then
+        "*Failed jobs*\n- Job details unavailable"
+      elif ($jobs | length) <= 10 then
+        "*Failed jobs*\n" + ($jobs | map("- `" + . + "`") | join("\n"))
+      else
+        "*Failed jobs*\n"
+        + ($jobs[:10] | map("- `" + . + "`") | join("\n"))
+        + "\n- ... and \((($jobs | length) - 10)) more"
+      end
+    '
+  )"
+fi
 
 LINKS_TEXT="<${RUN_URL}|View workflow run> | <${DASHBOARD_URL}|View CI dashboard>"
 FALLBACK_TEXT="CI ${STATUS_LABEL}: ${WORKFLOW_NAME} on ${REF_NAME}"
@@ -66,7 +78,7 @@ jq -n \
   --arg ref_name "${REF_NAME}" \
   --arg repository "${REPOSITORY}" \
   --arg timestamp "${TIMESTAMP}" \
-  --arg failed_jobs_text "${FAILED_JOBS_TEXT}" \
+  --arg failed_jobs_text "${DETAILS_TEXT}" \
   --arg links_text "${LINKS_TEXT}" \
   '{
     text: $fallback_text,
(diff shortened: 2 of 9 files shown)


tarun-etikala:
Thanks for the thorough iteration on this — the RHAIENG-6543 work (gate/agent Slack summary) looks good. CI is green and I ran the script tests locally (161/161).

### Merge path

This branch is built on an **older snapshot of #340** (first 3 commits only). #340 is approved and should merge first. Please **rebase onto latest `RHAIENG-7210/integrate-qg1-qg2-into-orchestrator`** (`41f68fd` or later) before merging #360. That brings in:

- `qg1-gate` / `qg2-gate` composite actions (orchestrator should use these, not inline QG1/QG2 steps)
- Standalone QG1/QG2 `schedule` removal (avoids duplicate nightly Slack alerts)
- Explicit DAG simulator (replaces the regex+eval version on this branch)

Expect conflicts in `quality-gates-pipeline.yml`, `test_orchestrator_contract.py`, and `test_orchestrator_dag_simulation.py`. The notification `if:` tests from this PR should port cleanly onto #340's contract test file.

### What looks good (RHAIENG-6543)

- `build_qg_summary.sh` — gate block + per-agent matrix, malformed-artifact resilience, character-budget truncation (not row count)
- Workflow resilience chain (`continue-on-error` + `if: always()` on summary build) — summary failure degrades to generic failed-jobs fallback instead of silencing the alert
- `notify-slack` `summary_text` input is backward-compatible for all other callers
- `QG7_EXCLUDED_AGENTS_JSON` as single workflow-level source of truth
- Test coverage is solid (`test_build_qg_summary.py`, notification `if:` tests, `test_notify_slack.py`)

### Small fixes worth doing on rebase

1. **QG7 status map — filter bad `.name` before aggregation** (`build_qg_summary.sh` ~L233–236): valid JSON with a missing/non-string `.name` makes the whole `jq -s` fail → `qg7_json='{}'` → every agent shows "QG7 did not run". Filter with `select((.name | type) == "string")` so one bad artifact doesn't wipe the rest.

2. **Docs — document `QG_NOTIFY_ON_SUCCESS`**: `docs/ci-alert-policy.md` / `docs/ci-alert-runbook.md` still say successful runs never notify. Add that success notifications are opt-in via `QG_NOTIFY_ON_SUCCESS == 'true'` (default: silent on success).

### Acknowledged / fine to defer

- Artifact download failure with partial gate summary (no matrix, no jobs-API fallback) — rare, visible in logs
- Merging QG4/QG7 outcome artifacts to avoid re-download — minor CI-minutes optimization

### Verdict

**Approve the RHAIENG-6543 feature after rebase onto merged #340** + the two small fixes above. Happy to re-review once rebased.

tarun-etikala (approved):
LGTM after rebase onto merged #340. Clean 3-commit delta, composite gates wired correctly, CI green, and I ran the script tests locally (178/178).

The consolidated single-alert design (vs per-stage alerts) is fine — documented and intentional.

**Suggestions (non-blocking):**
1. `build_qg_summary.sh` ~L235 — filter QG7 records with `select((.name | type) == "string")` before building the status map so one malformed artifact doesn't wipe all QG7 statuses.
2. Docs — note that `QG_NOTIFY_ON_SUCCESS == 'true'` opts into green success notifications (default remains silent on success).
3. PR description still references cherry-picked #340 commits — worth a quick update now that #340 is merged.

tarun-etikala:
Follow-up after rebase check: looks good. The branch is now a clean RHAIENG-6543-only delta on top of merged #340 — `qg1-gate`/`qg2-gate` composites, no duplicate commits, no merge conflicts. Rebase concern from my earlier comment is resolved.

**Suggestions (non-blocking, can be follow-up):**
- Filter QG7 outcome records by `.name` type in `build_qg_summary.sh` before aggregation (one bad file currently drops all QG7 statuses)
- Document `QG_NOTIFY_ON_SUCCESS` opt-in in `docs/ci-alert-policy.md` / `docs/ci-alert-runbook.md`
- Refresh PR description (still mentions cherry-picked #340 commits)

PR #362, opened 2026-09-09
test(langgraph): xfail non-streaming tool_calls parity check (RHAIENG-7429)

andrewdonheiser:
## Summary
- `test_streaming_parity_tool_calls` for `react_with_database_memory` has failed on every nightly QG7 scheduled run since at least 2026-09-02 (7+ consecutive days, 100% reproduction — not intermittent flakiness): `Tool sets differ: sync=set(), stream={'search'}`.
- Root cause is under investigation in [RHAIENG-7429](https://redhat.atlassian.net/browse/RHAIENG-7429) — likely a vLLM tool-call parser or LangChain non-streaming `ainvoke` gap, not the same issue fixed by #359 (RHAIENG-7403), which was scoped to a2a's MLflow trace correlation and doesn't apply to this test (it bypasses trace enrichment entirely).
- As a workaround, marks the test `xfail(strict=False)` so QG7 CI passes while the backend/parser issue is investigated. Remove the marker once RHAIENG-7429 is resolved.

## Test plan
- [x] `pre-commit` hooks pass
- [ ] Confirm QG7 nightly run for `langgraph-db-memory-agent` goes green with this change

[RHAIENG-7429]: https://redhat.atlassian.net/browse/RHAIENG-7429?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tarun-etikala (dismissed):
Makes sense as a temporary unblock for QG7 — the failure is deterministic (`sync=set(), stream={'search'}`), not the same nondeterminism we addressed in #363, and `test_streaming_parity_content` still covers streaming transport + content parity.

One suggestion: use `strict=True` on the xfail marker so CI fails with XPASS once RHAIENG-7429 is fixed, forcing removal of the marker instead of letting it rot silently.

tarun-etikala (approved):
Looks good — `strict=True` addresses the nit. Ready to merge.

PR #363, opened 2026-09-09
test(RHAIENG-7431): relax LlamaIndex streaming parity

andrewdonheiser:
## Summary

- Refactor the LlamaIndex websearch streaming test into a response-mode smoke test.
- Keep success and non-empty response checks for both streaming and non-streaming requests.
- Remove tool-call equality because the two modes issue independent LLM requests and do not guarantee identical tool selection.
- Tool-selection behavior remains covered by the dedicated behavioral tests.

## Evidence

The previous assertion failed with `non-streaming=set(), streaming={"dummy_web_search"}` in repeated QG7 runs. The agent creates separate workflow requests for the two modes, so this mismatch is not evidence of an SSE transport failure.

## Verification

- `make test` — 11 passed
- Behavioral test collection — passed
- `ruff check tests/behavioral/test_streaming_parity.py` — passed
- Pre-commit hooks — passed


tarun-etikala (changes requested):
Requesting changes for three issues I think should be addressed before merge:

1. The new QG summary treats every `QG7_RESULT == "skipped"` as "no agents eligible", but QG7 is also skipped when `collect-qg4` fails. That would misreport a real upstream failure as a benign skip in the Slack alert.

2. If the QG4/QG7 outcome artifact downloads fail transiently, the summary can omit the entire `*Agent Results*` section because those downloads are `continue-on-error`. That makes the alert look cleaner than reality instead of surfacing that agent-level data was unavailable.

3. The LlamaIndex `test_streaming_parity` no longer checks parity between streaming and non-streaming behavior; it now only asserts that both modes return non-empty text. That weakens coverage for regressions where streaming omits tool calls or diverges materially from the non-streaming path.

I did not find concrete security issues in the diff.

andrewdonheiser:
@tarun-etikala I addressed items 1 and 2 in commit 59e7ce9:

- QG7 skipped status now distinguishes **no agents eligible** from **QG4 result collection failure** and **QG4 not running**.
- QG4/QG7 artifact download outcomes are passed into the Slack summary. Download failures now keep the **Agent Results** section visible and explicitly report unavailable data.

Added regression coverage; all 181 CI-script tests and pre-commit checks pass.

For item 3, I’m keeping **test_streaming_parity** as-is. Streaming and non-streaming make independent LLM requests, so tool-call parity is nondeterministic and the previous assertion produced false positives. Tool-selection behavior remains covered by the dedicated behavioral tests; this test still verifies that both response modes succeed and return non-empty content.

tarun-etikala (approved):
Items 1 and 2 look good — the new COLLECT_QG4_RESULT branching and download-failure handling match what I was asking for, and the regression tests cover the main paths.

On item 3, I agree cross-request tool parity was flaky for this agent. Accepting the smoke-test scope.

